PR #23, opened 2020-08-26
Add specifications for functions for searching arrays

kgryte:
This PR

-   adds specifications for functions for searching arrays.
-   is derived from comparing API [signatures](https://github.com/data-apis/array-api-comparison/tree/314129762c09d54e299dd662491904ac594ed9ab/signatures/searching) across array libraries.

## Notes

-   NumPy includes additional search functions (e.g., `nanargmax`, `nanargmin`, `argwhere`, `flatnonzero`, `extract`, `searchsorted`); however, these functions are not widely implemented by other analyzed array libraries (less than half; see [here](https://github.com/data-apis/array-api-comparison/blob/7260fe1442bd325fe6f821ed8953eabdd52582cb/data/join.csv)) or are out-of-scope (`nan` variants) and, thus, were not included in this initial specification. Should additional search functions be necessary, they can be proposed in follow-up proposals.

## Questions (and notes)

-   **argmax**

    -   Some discrepancy with [`max`](https://github.com/data-apis/array-api/pull/17) in that multiple axes cannot be specified (CuPy supports, but no one else). Commonly, when computing the maximum value, array libraries allow users to reduce over multiple axes when performing the statistical "reduction". However, apart from CuPy, most of the analyzed array libraries do not seem to allow the same when searching for the indices of the maximum values. Should this be reconciled?
    -   Torch and CuPy both support `keepdims` which we support here in order to ensure broadcast compatibility as we do with statistical functions.
    -   Similar to `max` et al, should we also support an `out` keyword argument? This is supported by NumPy et al, but not Torch and TensorFlow.

-   **argmin**

    -   Same concerns/notes as for `argmax`.

-   **nonzero**

    -   How should zero-dimensional arrays be handled? NumPy requires at least `1d`, while Torch supports handling of zero-dimensional arrays as if one-dimensional.
    -   Torch has `as_tuple` keyword argument to either return a tuple of arrays or a single multi-dimensional array. Not clear on why most other array libraries insist on a tuple and not an ndarray (maybe because of NumPy)?

-   **where**

    -   The current spec requires "truthiness" rather than a strict array of booleans. This follows NumPy's behavior, where non-boolean conditions are permitted and act as expected (e.g., `0` is falsy while non-zero values are truthy). Should we be stricter here?
    -   Should we support an `out` keyword argument? Would seem consistent with other (element-wise) functions.
    -   The current spec proposes a ternary function (requires 3 args), omitting the trivial case of just providing a `condition` argument, as array library docs consistently point to alternative approaches (e.g., `np.asarray(condition).nonzero()`).


rgommers:
A little more archeology: http://numpy-discussion.10968.n7.nabble.com/just-curious-why-does-numpy-where-return-tuples-td14182.html#a14183 (2008):


_I assume that you are talking about where()'s single-argument form
which is equivalent to nonzero() (which I recommend using instead).
The reason that it returns a tuple is that the result is intended to
be usable as a "fancy" index. E.g._

```python
In [4]: x = random.randint(0, 2, [3, 10])

In [5]: x
Out[5]:
array([[1, 0, 0, 0, 0, 0, 1, 1, 0, 0],
       [1, 1, 0, 0, 1, 1, 1, 0, 0, 1],
       [1, 0, 0, 1, 0, 1, 0, 1, 0, 1]])

In [6]: nonzero(x)
Out[6]:
(array([0, 0, 0, 1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2]),
 array([0, 6, 7, 0, 1, 4, 5, 6, 9, 0, 3, 5, 7, 9]))

In [7]: x[nonzero(x)]
Out[7]: array([1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1])
```

_In Python, `x[i,j]` is implicitly `x[(i,j)]`. In order to support both
`x[i,j]` and `x[some_array]` (i.e., `some_array` is an array indexing on the
first axis only), we differentiate the inputs by type._

```python
In [8]: x[array(nonzero(x))]
IndexError
```

kgryte:
@rgommers Thanks for digging this up. :)

jakirkham:
Did we add `argwhere`? Didn't see it when searching through the PR, but could have easily missed it

kgryte:
@jakirkham `argwhere` is not currently in the [spec](https://github.com/data-apis/array-api/blob/7038c9c2b4f1f6685a6ea5f727bd66be7fb999ac/spec/API_specification/searching_functions.md).

jakirkham:
Ah ok. It would be useful to have `argwhere`

Since these are functions that can have unknown array lengths in their results (in Dask or other lazy computation libraries we don't know how many non-zeros there are until computed), it is helpful to have all of the indices in a single array like in `argwhere`. Having a single array lets us more easily work with that unknown length (as it is shared by all indices). Whereas something like `nonzero` has unknown lengths for each index. So it may not be obvious they can be used together (say in addition for example)

rgommers:
Argh, why do we have both `nonzero` and `argwhere`! They're almost identical. I actually like `argwhere` better, but because of the name it's used far less than `nonzero` I believe.

PyTorch has an `unbind` method that allows to efficiently go from 2-D tensor to tuple of 1-D tensors. Which, when applied to the output of `argwhere`, would be an O(1) method that makes `nonzero` unnecessary. And it would resolve the pain point PyTorch is having with incompatible `nonzero` definition.

This may be worth considering. The name is a little annoying though. Especially because `argwhere` isn't a counterpart to `where`, but the same thing as the discouraged one-element form of `where`.
```python
>>> x = np.array([0, 4, 7, 0])
>>> np.nonzero(x)
(array([1, 2]),)
>>> np.where(x)
(array([1, 2]),)
>>> np.argwhere(x)
array([[1],
       [2]])
>>> x[np.nonzero(x)]
array([4, 7])
>>> x[np.where(x)]
array([4, 7])
>>> x[np.argwhere(x)]  # this needs a squeeze() to roundtrip for 1-D input
array([[4],
       [7]])
```

PR #76, opened 2020-11-10
Add cross-reference targets to all the function, method, constant, and attribute headers

asmeurer:
This will make them easier to parse for the test suite. See https://github.com/data-apis/array-api/pull/72/files#r520185462.

Also, it turns out I was wrong. Sphinx make the anchor name based on the header text, not the cross-reference name. So the anchor URLs are still the same after this change. Actually it would be better if they were based on the cross-reference name, but I don't know how to do that.

### DIFF
--- a/spec/API_specification/array_object.md
+++ b/spec/API_specification/array_object.md
@@ -179,6 +179,7 @@ For the expected numerical behaviour, see their left-hand equivalents.
 
 <!-- NOTE: please keep the attributes in alphabetical order -->
 
+(attribute-dtype)=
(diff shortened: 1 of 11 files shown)


kgryte:
Ensuring readable URLs was the reason I included HTML markup in the first place.

kgryte:
And the use of `#` has precedent elsewhere in docs to refer to the readable URL.

asmeurer:
My guess it it's more of a Sphinx question than a MyST question.

PR #81, opened 2020-11-10
add an action to deploy to github pages when pushed to master.

### LINKED ISSUE #75: Add CI that auto-deploys Sphinx docs after merges to master
It should do something like (now doing locally)
```
# Install packages in requirements.txt
git clean -xdf
pushd spec
make html
popd
(git checkout --orphan tmp && git branch -D gh-pages || true)
git checkout --orphan gh-pages
git reset --hard
cp -R spec/_build/html/ latest
touch .nojekyll
git add .nojekyll
git add latest
git commit -m "Commit API Standard doc build";
git push --set-upstream origin gh-pages --force
```

Put the above in a file `deploy-ghpages.sh` at `..` from the repo root, then run (from `master`, in the repo root):
```
source ../deploy-ghpages.sh
```

tonyfast:
fixes #75

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -19,16 +19,16 @@ jobs:
         # add dependencies based on the conf.py
         pip install -r requirements.txt
     - name: Build docs
-      if: ${{ github.event_name == 'pull_request'}}
       run: |
         sphinx-build -b html -WT --keep-going spec build/latest
         touch build/.nojekyll
     - uses: actions/upload-artifact@v2
+      if: ${{ github.event_name == 'pull_request'}}
       with:
         name: html
         path: build/
         if-no-files-found: error
-    - name: Build and Deploy
+    - name: Deploy
       if: ${{ github.ref == 'refs/heads/master'}}
       uses: JamesIves/github-pages-deploy-action@master
       env:
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -20,7 +20,8 @@ jobs:
         pip install -r requirements.txt
     - name: Build docs
       run: |
-        sphinx-build -b html -WT --keep-going spec build/latest
+        # Turn warnings into errors; ensure .doctrees doesn't get deployed
+        sphinx-build -b html -WT --keep-going spec build/latest -d doctrees
         touch build/.nojekyll
     - uses: actions/upload-artifact@v2
       if: ${{ github.event_name == 'pull_request'}}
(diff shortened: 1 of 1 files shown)


.github/workflows/pages.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-18.04
+    steps:
+    # - uses: docker://pandoc/latex:2.9
+    - uses: actions/checkout@master
+    - uses: actions/setup-python@v2
+      with:
+        python-version: '3.8' # Version range or exact version of a Python version to use, using semvers version range syntax.
+        architecture: 'x64' # (x64 or x86)
+    - run: |
+        sudo apt install pandoc
```
rgommers, line 16: Just a question, did the build not work without `pandoc`? I have it installed but it wasn't in my conda env, so if it's needed it was a very hidden dependency.

.github/workflows/pages.yml
```diff
@@ -0,0 +1,27 @@
...
+        # add dependencies based on the conf.py
+        pip install sphinx myst-parser sphinx-copybutton sphinx-markdown-tables
```
rgommers, line 18: `pip install -r `requirements.txt`? shorter and probably more future-proof.

.github/workflows/pages.yml
```diff
@@ -0,0 +1,27 @@
...
+        sphinx-build spec build
```
rgommers, line 19: Should be `build/latest` I believe, to deploy to https://data-apis.github.io/array-api/latest/

rgommers (commented):
Thanks @tonyfast! Looks great, just a few comments. The last one is the most important, `latest/` needs to be in the URLs to not break incoming links.

.github/workflows/pages.yml
tonyfast: i removed this line. we shouldnt need. i think it is necessary with latex.

.github/workflows/pages.yml
tonyfast: this makes sense. where would you like the requirements to go?

tonyfast:
i resolved your comments. i set up the actions to create an artifact of the documents so they can be previewed. the github pages deployment only updates on changes to master.

.github/workflows/pages.yml
rgommers: They're already present in the repo root. With version pinning for Sphinx (safer), and `sphinx-material` which is missing here. I'll push a fix.

rgommers (approved):
Tested on my fork and fixed all remaining issues, let's give this a go! Thanks @tonyfast!

PR #87, opened 2020-11-13
Use maxdepth=3 in the API index

asmeurer:
This makes the API index page list every function in the spec. On the NumPy list, @mattip requested a page that lists every API function. This is an easy way to achieve this.


rgommers:
I like this idea. My first instinct was that it was very verbose and lost the overview, however on second thought the overview with top-level section is still in the left navbar:

image

And getting the overview of all objects with full signatures is nice. Let's go with this, we can iterate if needed.

PR #92, opened 2020-11-21
Clarify that the multiply() sign rules don't apply if the result is NaN

asmeurer:
We should actually define "mathematical sign" somewhere (especially since it
*really* means "sign bit"), but I'm not sure where this definition should go.

### DIFF
--- a/spec/API_specification/elementwise_functions.md
+++ b/spec/API_specification/elementwise_functions.md
@@ -977,8 +977,8 @@ Calculates the product for each element `x1_i` of the input array `x1` with the 
 #### Special Cases
 
 -   If either `x1_i` or `x2_i` is `NaN`, the result is `NaN`.
--   If `x1_i` and `x2_i` have the same mathematical sign, the result has a positive mathematical sign.
--   If `x1_i` and `x2_i` have different mathematical signs, the result has a negative mathematical sign.
+-   If `x1_i` and `x2_i` have the same mathematical sign, the result has a positive mathematical sign, except where it is `NaN` as in the rules below.
+-   If `x1_i` and `x2_i` have different mathematical signs, the result has a negative mathematical sign, except where it is `NaN` as in the rules below.
 -   If `x1_i` is either `+infinity` or `-infinity` and `x2_i` is either `+0` or `-0`, the result is `NaN`.
 -   If `x1_i` is either `+0` or `-0` and `x2_i` is either `+infinity` or `-infinity`, the result is `NaN`.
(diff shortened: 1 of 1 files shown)


kgryte:
Re: mathematical sign. No, the intent is not the sign bit; that is an implementation detail. We mean by mathematical sign the [sign attribute](https://en.wikipedia.org/wiki/Sign_%28mathematics%29) of a number. One way to check the sign of a floating-point number is by examining the binary representation of an IEEE 754 float, but this is not the only way. Comparison with zero and, for signed zero, division of one will also inform us as to the sign attribute.

Re: this PR. I am not in favor of including these changes as the phrasing is not correct. This PR indicates that a `NaN` result is the exception, or must obey alternative rules, but this is not required. As discussed elsewhere, an implementation could choose to returned a "signed" `NaN` for purposes of encoding the signed-ness of the operands.

The current spec is not contradictory in this regard, as both rules can be true at the same time. If there is ambiguity regarding how a `NaN` should be encoded, that is purposeful, as the signed-ness of `NaN` is implementation defined. While I am not in favor of the current changes, I am open to including a note or some such explanatory phrase indicating that how `NaN` values are encoded is implementation defined.

asmeurer:
Well to me "mathematical sign" means the mathematical definition of sign, which is akin to what the sign() function returns (mathematically 0 and -0 are the same thing, and 0 is neither positive nor negative). So while perhaps "sign bit" suggests an implementation detail, "mathematical sign" isn't much better than what was here before. I would suggest finding a better term, and more importantly, actually defining it.

> Re: this PR. I am not in favor of including these changes as the phrasing is not correct.

The current phrasing is also not correct if nan is allowed to not have a "mathematical sign", as it says that the result does have a sign (which is positive/negative). So if nan is allowed to not have a sign, it must be excluded from the rules that talk about mathematical sign. This doesn't preclude nan having a sign if the implementation choses that (although this is currently impossible to test in-spec as we don't have any functions that return the sign of a number).

kgryte:
I've pushed a [PR](https://github.com/data-apis/array-api/pull/94) for RFC 2119 compliance which also includes language which should address the concerns in this PR.

It is not that the rules for mathematical sign don't (or more precisely can't) apply to a `NaN` result, it is that we shouldn't require specification conforming implementations to follow any particular convention. In short, whether a `NaN` result is signed or not should be implementation-defined.

rgommers:
I agree that `NaN` shouldn't be special-cased here, and I think gh-94 addressed the concern here. So I'll close this PR. Thanks @asmeurer and @kgryte

PR #93, opened 2020-11-23
Specify rounding mode as half to even

kgryte:
This PR

-   specifies the rounding mode as half to even, rather than half away from zero. This follows NumPy and Python `3.x`, rather than C and Python `2.7` round half away from zero behavior.

### DIFF
--- a/spec/API_specification/elementwise_functions.md
+++ b/spec/API_specification/elementwise_functions.md
@@ -1136,11 +1136,6 @@ Returns the remainder of division for each element `x1_i` of the input array `x1
 
 Rounds each element `x_i` of the input array `x` to the nearest integer-valued number.
 
-```{note}
-
-The default rounding mode must be round half to even (i.e., rounding ties to the nearest even integer).
-```
-
 #### Special Cases
 
 -   If `x_i` is already integer-valued, the result is `x_i`.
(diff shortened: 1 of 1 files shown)


rgommers:
Thanks for the fix @kgryte.

> includes a TODO to consider whether round should support alternative rounding modes

My preference would be not to bother, since this is mostly a niche detail and current libraries don't implement multiple options.

I did a quick scan of NumPy issues, and while this question (e.g., why does `round(0.5)` give `0`?) is asked multiple times, no one argues an alternative matters to them.

For reference, here is the rationale for the rule: https://github.com/numpy/numpy/issues/3382#issuecomment-18692720, _Round-0.5-to-the-nearest-even, the rule numpy uses, is usually considered the best option for serious numerical code because it rounds up just as often as it rounds down, thus avoiding a source of statistical bias._

kgryte:
Removed TODO.

spec/API_specification/elementwise_functions.md
```diff
@@ -1136,10 +1136,15 @@ Returns the remainder of division for each element `x1_i` of the input array `x1
 
 Rounds each element `x_i` of the input array `x` to the nearest integer-valued number.
 
+```{note}
+
+The default rounding mode must be round half to even (i.e., rounding ties to the nearest even integer).
```
rgommers, line 1141: Does it make sense to say "default rounding mode" if there's just one mode? Maybe this note can simply be left out, because the same info is already present in the line you changed under Special Cases?

kgryte: Removed.

rgommers:
Thanks, merged!

PR #108, opened 2021-01-12
Fix an incorrect special case for pow

asmeurer:
I didn't catch this before because NumPy doesn't have `pow` (it uses `power`), so I hadn't actually run the tests for these prior to now.


kgryte (approved):
@asmeurer Thanks for catching this!

PR #110, opened 2021-01-14
Add Cholesky function specification

kgryte:
This PR

-   specifies the interface for performing Cholesky decomposition.
-   is derived from comparing signatures across [array libraries](https://github.com/data-apis/array-api-comparison/blob/0459e5dd51fa38df8bf24363f4fa5895ac5c2929/signatures/linalg/cholesky.md).

## Notes

-   NumPy (along with CuPy, JAX, MXNet, TF) does not allow returning either the lower- or upper-triangular Cholesky factor. However, SciPy, Torch, and Dask do support returning either. The ability to return either factor is common outside of the PyData ecosystem (MATLAB, LAPACK). Accordingly, the decision was made to include an `upper` keyword to support returning the upper-triangular Cholesky factor.

-   Following Torch, MXNet, TF, NumPy, and JAX, this proposal allows for providing a stack of square matrices.


leofang:
It is straightforward for CuPy to support the `upper` keyword, as internally we delegate to cuSOLVER which supports both modes: https://docs.nvidia.com/cuda/cusolver/index.html#cuSolverDN-lt-t-gt-potrf

spec/API_specification/linear_algebra_functions.md
```diff
@@ -16,9 +16,27 @@ A conforming implementation of the array API standard must provide and support t
...
+
+    -   input array having shape `(..., M, M)` and whose innermost two dimensions form square matrices. Must have a data type of either `float32` or `float64`.
+
+-   **upper**: _bool_
+
+    -   If `True`, the result must be the upper-triangular Cholesky factor. If `False`, the result must be the lower-triangular Cholesky factor. Default: `False`.
+
+#### Returns
+
+-   **out**: _Tuple\[ &lt;array&gt;, ... ]_
```
leofang, line 37: Same question here as I asked in the `solve()` PR: Do we really need to return a tuple here?

kgryte: I've updated the proposal to return an array, rather than a tuple.

leofang (approved):

kgryte:
Thanks, @leofang, for the review. This is ready for merge...

lucascolley:
What was the motivation for choosing the `upper=False` default here?

kgryte:
@lucascolley See the API comparison linked to in the OP: https://github.com/data-apis/array-api-comparison/blob/0459e5dd51fa38df8bf24363f4fa5895ac5c2929/signatures/linalg/cholesky.md

TL;DR: the kwarg was not universally supported. Among array libraries, PyTorch had `upper=False`, so we followed PyTorch.

lucascolley:
Thanks @kgryte . cc @mdhaber for awareness

PR #111, opened 2021-01-14
Add slogdet specification

kgryte:
This PR

-   specifies an interface for computing the sign and natural logarithm of the absolute value of the determinant.
-   is derived from comparing signatures across [array libraries](https://github.com/data-apis/array-api-comparison/blob/a5abdf0798704eaebd9034ec5ac0b3d5bf0d6fb8/signatures/linalg/slogdet.md).

## Notes

-   The `slogdet` interface is spec'd to return a namedtuple as more than one output array is returned.

-   Following Torch, MXNet, TF, NumPy, CuPy, and JAX and maintaining consistency with `det`, this proposal allows for providing a stack of square matrices.


spec/API_specification/linear_algebra_functions.md
```diff
@@ -265,9 +265,30 @@ TODO
...
+```{note}
+
+The purpose of this function is to calculate the determinant more accurately when the determinant is either very small or very large, as calling `det` may overflow or underflow.
+```
+
+#### Parameters
+
+-   **x**: _&lt;array&gt;_
+
+    -   input array having shape `(..., M, M)` and whose innermost two dimensions form square matrices. Must have a data type of either `float32` or `float64`.
```
rgommers, line 281: Why the limitation on input dtype? I checked numpy and cupy just now, they're perfectly happy with integer input. Output is always floating point.

kgryte: This currently follows [`det`](https://github.com/data-apis/array-api/blob/56fb2fbdec1f8d31fdb0ac92cf8540aa6c347222/spec/API_specification/linear_algebra_functions.md#detx-). In general, we've limited linalg specs to floating-point data types for consistency (i.e., not having a patchwork of linalg APIs supporting different data types). Can relax here (and for `det`), if desired.

leofang: If this is a concern (disclaimer: I don't really care), perhaps we could say the input would be promoted to floating-point types following the type promotion rule?

spec/API_specification/linear_algebra_functions.md
```diff
@@ -265,9 +265,30 @@ TODO
 TODO
 
 (function-slogdet)=
-### slogdet()
+### slogdet(x, /)
 
-TODO
+Returns the sign and the natural logarithm of the absolute value of the determinant of a square matrix (or stack of square matrices) `x`.
```
leofang, line 270: Nitpick:
```suggestion
Returns the sign and the natural logarithm of the absolute value of the determinant of a square matrix (or a stack of square matrices) `x`.
```

spec/API_specification/linear_algebra_functions.md
```diff
@@ -265,9 +265,30 @@ TODO
...
+
+-   **x**: _&lt;array&gt;_
+
+    -   input array having shape `(..., M, M)` and whose innermost two dimensions form square matrices. Must have a data type of either `float32` or `float64`.
+
+#### Returns
+
+-   **out**: _Tuple\[ &lt;array&gt;, ... ]_
+
+    -   if `x` is a two-dimensional array, a namedtuple (`sign`, `logabsdet`) whose first element must be zero-dimensional array containing a number representing the sign of the determinant and whose second element is a zero-dimensional array containing the determinant. Otherwise, a namedtuple (`sign`, `logabsdet`) whose first element must be a non-zero dimensional array containing the sign for each square matrix and whose second element must be a non-zero dimensional array containing the determinant for each square matrix.
```
leofang, line 287: Can we simplify the description here a bit to avoid repeating? Maybe we can just say `sign` and `logabsdet` are arrays of shape `x.shape[:-2]`?

spec/API_specification/linear_algebra_functions.md
kgryte: As the current type promotion rules do not address mixed-kind promotion, supporting integer input would be ill-defined. The only behavior the spec can presently define is when provided floating-point input.

spec/API_specification/linear_algebra_functions.md
kgryte: Resolved.

spec/API_specification/linear_algebra_functions.md
kgryte: Addressed.

spec/API_specification/linear_algebra_functions.md
kgryte: Loosened the dtype requirements to "should have a floating-point data type". This recommendation stems from needing to cast to a floating-point data type and mixed-kind promotions being undefined in the spec.

leofang (approved):

kgryte:
Thanks, @leofang, for the review. This is ready for merge...

PR #112, opened 2021-01-14
Add matrix_power specification

kgryte:
This PR

-   specifies an interface for raising a square matrix to an integer power.
-   is derived from comparing signatures across [array libraries](https://github.com/data-apis/array-api-comparison/blob/adc46493757c0b150504471c220c5956b259ceb8/signatures/linalg/matrix_power.md).

## Notes

-   The output array dtype for NumPy depends on both the input array and whether `n` is negative. If negative, then the result is floating-point. If non-negative, the result has the same dtype as the input. For CuPy, only `float32` and `float64` is allowed. In general, we recommend restricting linalg functions to floating-point dtypes. That practice is followed here.

-   Following Torch and maintaining consistency with other linalg interfaces operating in square matrices, a stack of square matrices is accepted.


spec/API_specification/linear_algebra_functions.md
```diff
@@ -145,9 +145,25 @@ TODO
...
+
+-   **n**: _int_
+
+    -   integer exponent.
+
+#### Returns
+
+-   **out**: _&lt;array&gt;_
+
+    -   if `n` is equal to zero, an array containing the identity matrix for each square matrix. If `n` is less than zero, an array containing the inverse of each square matrix raised to the absolute value of `n`, provided that each square matrix is invertible. If `n` is greater than zero, an array containing the result of raising each square matrix to the power `n`. The returned array must have the same data type and shape as `x`.
```
rgommers, line 166: > provided that each square matrix is invertible.

I assume all implementations will raise some kind of exception (which kind will vary) - okay to not specify, so just commenting on this to indicate I agree.

rgommers (changes requested):
> For CuPy, only `float32` and `float64` is allowed. In general, we restrict linalg functions to floating-point dtypes. That practice is followed here.

Only specifying float32/64 sounds right, however the phrasing is a bit off. I'd rather add a note saying that the behaviour is only specified for these dtypes, and that for other dtypes implementation may differ - they will either cast or raise for some or all values of `n`.

kgryte:
Re: dtypes. We'd need to add that note more generally at the top of the linalg spec, as this is generally true for most linalg specs, given their current support being "limited" to floating-point dtypes. The restriction to floating-point dtypes was discussed a couple times during meetings and found its support there.

spec/API_specification/linear_algebra_functions.md
```diff
@@ -145,9 +145,25 @@ TODO
...
+
+    -   input array having shape `(..., M, M)` and whose innermost two dimensions form square matrices. Must have a data type of either `float32` or `float64`.
+
+-   **n**: _int_
+
+    -   integer exponent.
+
+#### Returns
+
+-   **out**: _&lt;array&gt;_
```
leofang, line 164: Allow me to ask a question opposite to what I asked elsewhere: Do we wanna return a namedtuple here?

spec/API_specification/linear_algebra_functions.md
```diff
@@ -145,9 +145,25 @@ TODO
 TODO
 
 (function-matrix_power)=
-### matrix_power()
+### matrix_power(x, n, /)
 
-TODO
+Raises a square matrix (or stack of square matrices) `x` to an integer power `n`.
```
leofang, line 150: Nitpick:
```suggestion
Raises a square matrix (or a stack of square matrices) `x` to an integer power `n`.
```

spec/API_specification/linear_algebra_functions.md
kgryte: I've updated linalg spec proposals to only return a tuple when more than one array output is returned. In this case, as only one array is returned, seems okay to leave as is.

spec/API_specification/linear_algebra_functions.md
kgryte: Done.

spec/API_specification/linear_algebra_functions.md
kgryte: Added a note regarding possible exceptions.

kgryte:
@rgommers re: dtypes. We could change the language from `must` to `should`. As in, we could recommend floating-point input, but allow implementations to accept, e.g., integer dtypes.

spec/API_specification/linear_algebra_functions.md
```diff
@@ -145,9 +145,29 @@ TODO
...
+
+#### Returns
+
+-   **out**: _&lt;array&gt;_
+
+    -   if `n` is equal to zero, an array containing the identity matrix for each square matrix. If `n` is less than zero, an array containing the inverse of each square matrix raised to the absolute value of `n`, provided that each square matrix is invertible. If `n` is greater than zero, an array containing the result of raising each square matrix to the power `n`. The returned array must have the same data type and shape as `x`.
+
+#### Raises
+
+-   if `n` is less than zero and a square matrix is not invertible.
```
asmeurer, line 170: Also if the matrix isn't square?

kgryte: Yeah, we should be explicit. Currently, it is implied based on the shape requirement.

kgryte: Done.

kgryte:
This PR has been open for some time without further comment and has been discussed/approved during meetings. Will merge, and we can submit follow-up PRs to resolve any issues/concerns which may arise.

kgryte:
For reference, inclusion of this API in the standard was discussed [here](https://github.com/data-apis/array-api/issues/147#issuecomment-806664629).

Issue #95, opened 2020-11-30
API for variable number of returns in linalg


rgommers:
Feedback from @mruberry: _it may be nice to support always returning the same type from functions like Cholesky, with tensors accessed by names. That way frameworks could return additional tensors and there wouldn't be BC-compatibility issues if more tensor returns are added later. This suggestion, however, comes from a place of "it'd be nice if people wanted to use this API", as opposed to the perspective of, "this API is a lowest common denominator only intended for library writers."_

`cholesky` always return a single array, so I thought it'd be fine - but the comment may be related to https://github.com/pytorch/pytorch/issues/47608 (desire to return an error code rather than raise a `RuntimeError` for failures). For `qr`, `svd` and perhaps some more functions there's the related issue of different returns based on keywords. Using a class to stuff all return values in is a common design pattern for, e.g., `scipy.optimize`. I'm not convinced it's a good idea for the linalg functions here, but worth considering at least.

asmeurer:
I don't understand the technical details of that pytorch issue. Does it mean that any API that raises an exception is potentially problematic?

mruberry:
> I don't understand the technical details of that pytorch issue. Does it mean that any API that raises an exception is potentially problematic?

Yes, because checking for the exception will cause a cross-device synchronization, which PyTorch tries very hard to avoid. In PyTorch also have cases where error-checking occurs exclusively on the CPU to avoid this issue.

We intend to keep np.linalg-like error-checking behavior (with a warning that it causes cross-device synchronization) but also offer `_ex` variants that avoid it. For example, `torch.linalg.cholesky` and `torch.linalg.cholesky_ex`, the latter of which performs no error checking but returns the `info` required for error-checking as an array (tensor) on the device. This lets the user perform their own error-checking (or not).

rgommers:
Now that there are a number of PRs up for linear algebra functions, it looks to me like the issues with variable number of returned values, order of returns, and transposes within the array even for matches shapes:

- `svd`: NumPy and PyTorch order the returned tuple `(u, s, v)`, while TF orders as `(s, u, v)` (see gh-114).
- `svd`:TensorFlow returns both the left and right singular vectors in columns. NumPy returns the left singular vectors in rows.
- `lstsq`: what gets returned is wildly different, gh-119 lists six variations.
- `pinv`: mostly okay for array/tensor libraries, but SciPy returns rank in addition to the pseudo-inverse (see gh-118)

[...]

PR #114, opened 2021-01-14
Add SVD function specification

kgryte:
This PR

-   specifies the interface for performing singular value decomposition (SVD).
-   is derived from comparing signatures across [array libraries](https://github.com/data-apis/array-api-comparison/blob/a82d5fc3cda4d535936f827af89b028231c64222/signatures/linalg/svd.md).

## Notes

-   NumPy and Torch order the returned tuple `(u, s, v)`, while TF orders as `(s, u, v)`. This proposal follows NumPy.

-   NumPy et al provide a `compute_uv` keyword to indicate whether to return the left and right singular vectors along with the singular values. This proposal omits such a keyword, in favor of a separate `svdvals` specification (see [gh-160](https://github.com/data-apis/array-api/pull/160)). This is to promote consistency with `eig` and `eigvals` and minimize polymorphism within the set of linagl APIs.

-   By default, NumPy computes full matrixes, while TF and Torch do not. This proposal follows NumPy and sets the `full_matrices` keyword argument default as `True`.

-   MXNet does not currently support any keyword arguments.

-   Dask currently supports a `coerce_signs` keyword argument to indicate whether or not to apply sign coercion to singular vectors in order to maintain deterministic results. It is alone in this regard.

-   This proposal follows NumPy, Torch, MXNet, JAX, and TF in supporting stacks of matrices. CuPy and Dask do not currently support providing stacks.

-   NumPy currently supports a `hermitian` keyword argument for speeding up computation; however, it is alone in doing this.

-   TF (and LAPACK) returns both the left and right singular vectors in columns. NumPy returns the left singular vectors in rows (i.e., NumPy returns the adjoint). This proposal follows NumPy.


leofang:
> CuPy and Dask do not currently support providing stacks.

I checked that we have this capability, so it's straightforward for us to support it:
https://github.com/cupy/cupy/blob/a6c75b901caa3be19ce8c4f717f2f780e457559d/cupy/cusolver.py#L71
It's just that for some reason we ended up staying with the QR-based SVD (`gesvd`, without the suffix `j`) from cuSOLVER, which does not have a batched version.

rgommers:
> This proposal follows TF (and LAPACK) in requiring that both the left and right singular vectors be returned in columns. NumPy returns the left singular vectors in rows.

`tensorflow.experimental.numpy` doesn't have an `svd` function yet, PyTorch matches NumPy, not sure about the rest (probably matches NumPy as well) - so this choice doesn't seem quite right. It's also hard to interpret the specification here. How about adding a note on this, mentioning how to reconstruct the input:
```
s, u, v = svd(x)
y = dot(u, dot(s, v))   # or does this need a transpose somewhere??
assert_allclose(x, y)
```

rgommers:
There's an incompatibility between NumPy and PyTorch missed in number of values returned for `compute_uv = False` - see https://github.com/data-apis/array-api/issues/95#issuecomment-770016642

leofang:
> > CuPy and Dask do not currently support providing stacks.
> 
> I checked that we have this capability (added in [cupy/cupy#3247](https://github.com/cupy/cupy/pull/3247)), so it's straightforward for us to support it:
> https://github.com/cupy/cupy/blob/a6c75b901caa3be19ce8c4f717f2f780e457559d/cupy/cusolver.py#L71
> It's just that for some reason we ended up staying with the QR-based SVD (`gesvd`, without the suffix `j`) from cuSOLVER, which does not have a batched version.

So it turns out not as straightforward, but is still possible on both CUDA and HIP, see cupy/cupy#4628. 

btw, we could also note the support for complex numbers will land in the next version (like what we discussed in the eigensolver PR #113). The nice point of SVD is that `u` and `v` are real if the matrix is real, so unlike eigensolvers here it is not blocked by the lack of complex support.

kgryte:
Updated this PR based on the above feedback and meeting discussions. Reordered the output values to follow NumPy `(u,s,v)` and specified that `v` be the adjoint (i.e., right singular vectors returned in rows).

kgryte:
Updated this PR to return an array, rather than a tuple, when `compute_uv` is `False`.

spec/API_specification/linear_algebra_functions.md
```diff
@@ -275,9 +275,35 @@ TODO
 TODO
 
 (function-svd)=
-### svd()
+### svd(x, /, *, compute_uv=True, full_matrices=True)
```
leofang, line 278: Are we purposely deviating from NumPy's signature, in which `full_matrices` goes before `compute_uv`? Logically the choice here makes sense as `compute_uv` could be ignored, but I'd like to make sure this is a delibrate choice.

spec/API_specification/linear_algebra_functions.md
```diff
@@ -275,9 +275,35 @@ TODO
...
+
+    -   input array having shape `(..., M, N)` and whose innermost two dimensions form matrices on which to perform singular value decomposition. Must have a data type of either `float32` or `float64`.
+
+-   **compute_uv**: _bool_
+
+    -   If `True`, compute the left and right singular vectors and return as `u` and `v`, respectively. Default: `True`.
+
+-   **full_matrices**: _bool_
+
+    -   If `True`, compute full-sized `u` and `v`, such that `u` has shape `(..., M, M)` and `v` has shape `(..., N, N)`. If `False`, compute on the leading `K` singular vectors, such that `u` has shape `(..., M, K)` and `v` has shape `(..., N, K)` and where `K = min(M, N)`. This option must be ignored if `compute_uv` is `False`. Default: `True`.
```
leofang, line 294: In NumPy, `v.shape == (..., K, N)`, not `(..., N, K)`. I suppose we wanna follow NumPy here?

spec/API_specification/linear_algebra_functions.md
```diff
@@ -275,9 +275,35 @@ TODO
...
+
+#### Returns
+
+-   **out**: _Union\[ &lt;array&gt;, Tuple\[ &lt;array&gt;, ... ] ]_
+
+    -   if `compute_uv` is `False`, an array having shape `(..., K)` and containing the vector(s) containing the singular values. For each vector, the values must be sorted in descending order according to magnitude, such that `s[..., 0]` is the largest value, `s[..., 1]` is the second largest value, et cetera. The first `x.ndim-2` dimensions must have the same size as those of the input `x`.
```
leofang, line 300: Probably this is the style used throughout the spec, but I find the first sentence is a bit terse. How about this?
```suggestion
    -   if `compute_uv` is `False`, an array with shape `(..., K)` that contains the vector(s) of singular values of length `K`. For each vector, the singular values must be sorted in descending order by magnitude, such that `s[..., 0]` is the largest value, `s[..., 1]` is the second largest value, et cetera. The first `x.ndim-2` dimensions must have the same shape as those of the input `x`.
```

spec/API_specification/linear_algebra_functions.md
```diff
@@ -275,9 +275,35 @@ TODO
...
+
+    -   if `compute_uv` is `True`, a namedtuple `(u, s, v)` whose
+    
+        -   first element must be an array whose shape depends on the value of `full_matrices` and contain unitary array(s) (i.e., the left singular vectors). The left singular vectors must be stored as columns. If `full_matrices` is `True`, the array must have shape `(..., M, M)`. If `full_matrices` is `False`, the array must have shape `(..., M, K)`, where `K = min(M, N)`. The first `x.ndim-2` dimensions must have the same size as those of the input `x`.
```
leofang, line 304:
```suggestion
        -   first element must be an array whose shape depends on the value of `full_matrices` and contain unitary array(s) (i.e., the left singular vectors). The left singular vectors must be stored as columns. If `full_matrices` is `True`, the array must have shape `(..., M, M)`. If `full_matrices` is `False`, the array must have shape `(..., M, K)`, where `K = min(M, N)`. The first `x.ndim-2` dimensions must have the same shape as those of the input `x`.
```

spec/API_specification/linear_algebra_functions.md
```diff
@@ -275,9 +275,35 @@ TODO
...
+        -   second element must be an array having shape `(..., K)` and contain the vector(s) containing the singular values. For each vector, the values must be sorted in descending order according to magnitude, such that `s[..., 0]` is the largest value, `s[..., 1]` is the second largest value, et cetera. The first `x.ndim-2` dimensions must have the same size as those of the input `x`.
```
leofang, line 305:
```suggestion
        -   second element must be an array with shape `(..., K)` that contains the vector(s) of singular values of length `K`. For each vector, the singular values must be sorted in descending order by magnitude, such that `s[..., 0]` is the largest value, `s[..., 1]` is the second largest value, et cetera. The first `x.ndim-2` dimensions must have the same shape as those of the input `x`.
```

spec/API_specification/linear_algebra_functions.md
```diff
@@ -275,9 +275,35 @@ TODO
...
+        -   third element must be an array whose shape depends on the value of `full_matrices` and contain unitary array(s) (i.e., the right singular vectors). The right singular vectors must be stored as rows (i.e., the array is the adjoint). If `full_matrices` is `True`, the array must have shape `(..., N, N)`. If `full_matrices` is `False`, the array must have shape `(..., K, N)` where `K = min(M, N)`. The first `x.ndim-2` dimensions must have the same size as those of the input `x`.
```
leofang, line 306:
```suggestion
        -   third element must be an array whose shape depends on the value of `full_matrices` and contain unitary array(s) (i.e., the right singular vectors). The right singular vectors must be stored as rows (i.e., the array is the adjoint). If `full_matrices` is `True`, the array must have shape `(..., N, N)`. If `full_matrices` is `False`, the array must have shape `(..., K, N)` where `K = min(M, N)`. The first `x.ndim-2` dimensions must have the same shape as those of the input `x`. 
```

leofang: Note that the shape is correctly given as `(..., K, N)` here.

leofang (changes requested):

spec/API_specification/linear_algebra_functions.md
kgryte: These are keyword-only, and not positional, arguments, so the order in which they appear should not matter. Here, they are listed in alphabetical order.

spec/API_specification/linear_algebra_functions.md
kgryte: @leofang Thanks for the suggestion!

spec/API_specification/linear_algebra_functions.md
kgryte: Good catch!

spec/API_specification/linear_algebra_functions.md
leofang: Ah yes, thanks @kgryte! My builtin parser automatically discarded / and * ...

spec/API_specification/linear_algebra_functions.md
kgryte: Indeed. I updated below, but not here. PR should be fixed.

leofang (approved):

IvanYashchuk:
Why does `compute_uv = True/False` exists in NumPy and should NumPy be followed in this case? There is no `compute_v = True/False` for `numpy.linalg.eig` and `numpy.linalg.eigh`, there are separate functions for that: `numpy.linalg.eigvals` and `numpy.linalg.eigvalsh`.

Should this flag be dropped for `svd` and functionality replaced by a separate function `svdvals`, similarly to [SciPy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.svdvals.html) and [Julia](https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#LinearAlgebra.svdvals)?

rgommers:
Good question @IvanYashchuk. No/fewer boolean keywords would make the design better, that is probably how we'd design it if starting from scratch. I do see that `scipy.linalg.svd` has a `compute_uv` keyword. That's a bit unnecessary because it has `svdvals` too, it was probably added for compatibility with numpy.

I like the idea of dropping `compute_uv`, the comparison with `eig`/`eigvals` is nice.

rgommers:
The signature `svd(x, /, *, full_matrices=True)` would be fully compatible with what existing libraries already have, so this may be a case where we can make the design more consistent without introducing extra issues/work for libraries that want to support this API standard in their current/main namespace.

leofang:
> Why does `compute_uv = True/False` exists in NumPy and should NumPy be followed in this case?

I think in reality it depends on how things are implemented at the low level. Taking CUDA as an example, this function [`cusolverDn<t>gesvd()`](https://docs.nvidia.com/cuda/cusolver/index.html#cuSolverDN-lt-t-gt-gesvd) is one of the routines implementing SVD, and if you take a closer look it provides both `compute_uv` and `full_matrices` options, which CuPy (and at least JAX I think) takes advantage of. 

In fact, I think this interface dates way back to Lapack, and I assume this was why NumPy had this API design in the first place. So regardless how we design/split the API, under the hood they all call the same routine. So, it might lead to some code duplication if we split, though it's an implementation detail and I am not sure we should care here. 

But, if we take into account decades of familiarity in numerical routines I don't think changing the API is a good idea, though I don't have strong opinion.

spec/API_specification/linear_algebra_functions.md
```diff
@@ -275,9 +275,35 @@ TODO
...
+
+-   **full_matrices**: _bool_
+
+    -   If `True`, compute full-sized `u` and `v`, such that `u` has shape `(..., M, M)` and `v` has shape `(..., N, N)`. If `False`, compute on the leading `K` singular vectors, such that `u` has shape `(..., M, K)` and `v` has shape `(..., K, N)` and where `K = min(M, N)`. This option must be ignored if `compute_uv` is `False`. Default: `True`.
+
+#### Returns
+
+-   **out**: _Union\[ &lt;array&gt;, Tuple\[ &lt;array&gt;, ... ] ]_
+
+    -   if `compute_uv` is `False`, an array with shape `(..., K)` that contains the vector(s) of singular values of length `K`. For each vector, the singular values must be sorted in descending order by magnitude, such that `s[..., 0]` is the largest value, `s[..., 1]` is the second largest value, et cetera. The first `x.ndim-2` dimensions must have the same shape as those of the input `x`.
```
leofang, line 300: Perhaps to be consistent across the `linalg` spectrum (ref: #95), this should return a namedtuple `(s,)` too?

rgommers: I think those are all mistakes. A tuple with a single array in it is kind of weird and unpythonic. Doesn't matter if it's a namedtuple or not.

The design consistency is quite hard to see across all these linalg PRs. We need the overview PR with design principles for all linalg functions first I think.

leofang: Sounds good. Yeah I left a bunch of questions yesterday about whether it should be a namedtuple or just an array when there is only one object to be returned.

kgryte:
I'm also in favor of @IvanYashchuk's proposal to have a separate API which only returns singular values.

Regarding whether an implementation reuses the same routine for both `svd` and `svdvals` I don't think should be a strong factor in how we specify the APIs. In general, APIs which always return the same object shapes (array or tuple, but not either) should be desired given that code is (a) easier to optimize (non-polymorphism) and (b) easier to reason about.

leofang (approved):

kgryte:
I've updated this proposal (and OP) to no longer include the `compute_uv` keyword in favor a separate `svdvals` to support only returning the singular values (see [gh-160](https://github.com/data-apis/array-api/pull/160)).

spec/API_specification/linear_algebra_functions.md
kgryte: This should now be resolved due to splitting of `svd` into `svd` and `svdvals`.

kgryte:

Issue #98, opened 2020-12-05
Issues with "Mixing arrays and Python scalars" section


asmeurer:
A few issues with the [mixing arrays and Python scalars](https://github.com/data-apis/array-api/blob/main/spec/API_specification/type_promotion.md#mixing-arrays-with-python-scalars) section that came up from testing:

- The matmul operator (`@`) does not (and should not) support scalar types, so it should be excluded.
- Should boolean arrays be required to work with arithmetic operators. NumPy gives errors in some cases, for example:

```py
>>> import numpy as np
>>> np.array(False) - False
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: numpy boolean subtract, the `-` operator, is not supported, use the bitwise_xor, the `^` operator, or the logical_xor function instead. 
```

- How should casting work? For example, `int8 array + large Python int`. Should it cast the int, or give an error. Or should this be unspecified.
- The above question also applies for Python float -> float32 casting, which is implicitly supported by the text that is currently there, but is unsafe in general (is there a "safe" variant of float64 -> float32, or is it always unsafe? Either way, this would be value based casting, which I thought we were trying to avoid).

asmeurer:
Actually, I think `-` is the only operator that NumPy doesn't allow for booleans. But there's also the reverse which is that bitwise operators are not supported for floats. So I think there should be some text like "for operations that are supported by the given dtypes" and the operators text should specify that (I guess the bitwise ones already do, but there's nothing in `__sub__` about boolean support or non-support).

asmeurer:
IMHO, the spec should split the operators into four classes:

- Arithmetic operators (+, -, *, /, etc. (including unary + and -))
- Boolean operators (~, ^, |, etc.)
- Comparison operators (==, <=, ...)
- Array operators (@ is the only one)

Arithmetic operators should be defined for integer and floating point dtypes, but not required for the boolean dtype. Boolean operators should be defined for integer (bitwise) and boolean dtypes, but are not required for floating point dtypes. Comparison operators are defined for all arrays regardless of dtype. The array operator (matmul) is defined only for integer and floating point dtypes, and also has additional shape constraints (the input arrays must have at least 1 (2?) dimensions, which align in a certain way).

But I may also be missing or forgetting prior discussions that happened with this.

asmeurer:
A similar split could be made for the 2-argument array functions.

asmeurer:
[...]

PR #116, opened 2021-01-20
Fix a typo in a type annotation

asmeurer:
This also removes the dtype from the `where()` type annotation. We should document input and output dtypes (https://github.com/data-apis/array-api/issues/98), but type annotations aren't the right place to do this.

### DIFF
--- a/spec/API_specification/creation_functions.md
+++ b/spec/API_specification/creation_functions.md
@@ -143,7 +143,7 @@ Returns a new array having a specified `shape` and filled with `fill_value`.
 
     -   output array shape.
 
--   **fill_value**: _Union\[ int, float ] ]_
+-   **fill_value**: _Union\[ int, float ]_
 
     -   fill value.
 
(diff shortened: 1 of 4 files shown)


asmeurer:
I changed the annotation for the `axis` argument to `stack` from `Optional[int]` to just `int`, since the default is `0` rather than `None`. However, I'm not sure if this should be done, so I can revert it if we think the `Optional` is better. It's worth noting that `np.stack(axis=None)` gives an error.

asmeurer:
I also changed the `endpoint` argument of linspace to be `bool` instead of `Optional[bool]`. Perhaps there was just some confusion about the difference between `Optional[x]`, which is a shorthand for `Union[x, None]` (see https://docs.python.org/3/library/typing.html#typing.Optional), and keyword-only arguments with default values, which may be `None` or something else.

kgryte (approved):
Thanks for catching this!

kgryte:
> Perhaps there was just some confusion about the difference between Optional[x].

No. Just a copy and paste error.

PR #117, opened 2021-01-21
Fix an incorrect statement about indexing

asmeurer:
The equivalent slice for a negative integer index should be n+i:n+i+1.


kgryte (approved):
Thanks for catching this!

PR #121, opened 2021-01-30
Explicitly specify which categories of dtypes the elementwise functions are defined for

asmeurer:
Please double check that I didn't make any mistakes.

Some comments:

- I made all transcendental functions require floating-point inputs (i.e., no integers). This is because they say the output must use type promotion, but integer dtypes do not type promote to floating-point dtypes according to the type promotion rules. 
- For `pow`, `pow(a, b)` does not make sense if `a` and `b` are integers and `b` is negative. Should we restrict it to only being defined for nonnegative `b`? I think I might have brought this up before somewhere, but I don't remember what was decided.
  - (as an aside, shouldn't `pow(0, 0) == 1` be listed as a special case for `pow`?)
- The same question for `divide`. Should it be defined for integer inputs, and if so, does that preclude the definition when the second argument is 0?

I will also update the array methods as per https://github.com/data-apis/array-api/issues/98#issuecomment-739088199 in another PR once this is merged.

### DIFF
--- a/spec/API_specification/data_types.md
+++ b/spec/API_specification/data_types.md
@@ -82,10 +82,10 @@ A conforming implementation of the array API standard may provide and support ad
 (data-type-categories)=
 ## Data Type Categories
 
-For the purposes of this spec, the following data type categories are defined.
+For the purposes of this specification, the following data type categories are defined.
 Libraries do not need to organize dtypes according to these categories. These
-are only for organizing the functions in this spec itself. Future versions of
-the spec will include additional categories for complex data types.
+are only for organizing the functions in this specification itself. Future versions of
+the specification will include additional categories for complex data types.
 
 ### Numeric Data Types
 
--- a/spec/API_specification/data_types.md
+++ b/spec/API_specification/data_types.md
@@ -78,3 +78,29 @@ Implementations may provide other ways to specify data types (e.g.,
 
 A conforming implementation of the array API standard may provide and support additional data types beyond those described in this specification.
 ```
+
+(data-type-categories)=
+## Data Type Categories
+
+For the purposes of this spec, the following data type categories are defined.
+Libraries do not need to organize dtypes according to these categories. These
+are only for organizing the functions in this spec itself. Future versions of
+the spec will include additional categories for complex data types.
+
+### Numeric Data Types
+
+``int8`, `int16`, `int32`, `int64`, `uint8`, `uint16`, `uint32`,
+`uint64`, `float32`, and `float64` (i.e., all dtypes except for `bool`).
+
+### Integer Data Types
+
+``int8`, `int16`, `int32`, `int64`, `uint8`, `uint16`, `uint32`, and
+`uint64`.
+
+### Floating-point Data Types
+
+`float32` and `float64`.
+
+### Boolean Data Types
+
+`bool`.
--- a/spec/API_specification/elementwise_functions.md
+++ b/spec/API_specification/elementwise_functions.md
@@ -10,6 +10,7 @@ A conforming implementation of the array API standard must provide and support t
 -   Optional parameters must be [keyword-only](https://www.python.org/dev/peps/pep-3102/) arguments.
 -   Broadcasting semantics must follow the semantics defined in {ref}`broadcasting`.
 -   Unless stated otherwise, functions must support the data types defined in {ref}`data-types`.
+-   Functions may only be required for a subset of input data type. Libraries may chose to implement functions for additional data types, but that behavior is not required by the spec. See {ref}`data-type-categories`.
 -   Unless stated otherwise, functions must adhere to the type promotion rules defined in {ref}`type-promotion`.
 -   Unless stated otherwise, floating-point operations must adhere to IEEE 754-2019.
 -   Unless stated otherwise, element-wise mathematical functions must satisfy the minimum accuracy requirements defined in {ref}`accuracy`.
@@ -35,7 +36,7 @@ For floating-point operands,
 
 -   **x**: _&lt;array&gt;_
 
-    -   input array.
+    -   input array. Should have a numeric data type.
 
 #### Returns
 
@@ -61,7 +62,7 @@ For floating-point operands,
 
 -   **x**: _&lt;array&gt;_
 
-    -   input array.
+    -   input array. Should have a floating-point data type.
 
 #### Returns
 
@@ -87,7 +88,7 @@ For floating-point operands,
 
 -   **x**: _&lt;array&gt;_
 
-    -   input array whose elements each represent the area of a hyperbolic sector.
+    -   input array whose elements each represent the area of a hyperbolic sector. Should have a floating-point data type.
 
 #### Returns
 
@@ -131,11 +132,11 @@ Floating-point addition is a commutative operation, but not always associative.
 
 -   **x1**: _&lt;array&gt;_
 
-    -   first input array.
+    -   first input array. Should have a numeric data type.
 
 -   **x2**: _&lt;array&gt;_
 
-    -   second input array. Must be compatible with `x1` (see {ref}`broadcasting`).
+    -   second input array. Must be compatible with `x1` (see {ref}`broadcasting`). Should have a numeric data type.
 
 #### Returns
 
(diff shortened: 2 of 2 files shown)


rgommers:
Thanks Aaron. This categorization makes sense, and I agree with "must" -> "should".

rgommers:
> I made all transcendental functions require floating-point inputs (i.e., no integers). This is because they say the output must use type promotion, but integer dtypes do not type promote to floating-point dtypes according to the type promotion rules.

This seems correct to me. Most libraries will happily accept integer dtypes and upcast to float, but IIRC at least TensorFlow raises an exception instead.

+1 for the `pow` proposal and listing the special case.

`divide` should not be limited I believe. It's different from transcendental functions, because it's more important that it matches `/` (which in Python 3 does `int / int -> float`) than that there's no casting rules for int to float.

spec/API_specification/data_types.md
```diff
@@ -78,3 +78,29 @@ Implementations may provide other ways to specify data types (e.g.,
...
+## Data Type Categories
+
+For the purposes of this spec, the following data type categories are defined.
+Libraries do not need to organize dtypes according to these categories. These
+are only for organizing the functions in this spec itself. Future versions of
+the spec will include additional categories for complex data types.
+
+### Numeric Data Types
+
+``int8`, `int16`, `int32`, `int64`, `uint8`, `uint16`, `uint32`,
```
leofang, line 92:
```suggestion
`int8`, `int16`, `int32`, `int64`, `uint8`, `uint16`, `uint32`,
```

spec/API_specification/data_types.md
```diff
@@ -78,3 +78,29 @@ Implementations may provide other ways to specify data types (e.g.,
...
+`uint64`, `float32`, and `float64` (i.e., all dtypes except for `bool`).
+
+### Integer Data Types
+
+``int8`, `int16`, `int32`, `int64`, `uint8`, `uint16`, `uint32`, and
```
leofang, line 97:
```suggestion
`int8`, `int16`, `int32`, `int64`, `uint8`, `uint16`, `uint32`, and
```

leofang (commented):
Caught two typos while digesting the changes 🙂

spec/API_specification/elementwise_functions.md
```diff
@@ -1035,11 +1036,11 @@ Floating-point multiplication is not always associative due to finite precision.
 
 -   **x1**: _&lt;array&gt;_
 
-    -   first input array.
+    -   first input array. Should have a numeric data type.
 
 -   **x2**: _&lt;array&gt;_
 
-    -   second input array. Must be compatible with `x1` (see {ref}`broadcasting`).
+    -   second input array. Must be compatible with `x1` (see {ref}`broadcasting`). Should have a numeri
```
kgryte, line 1043: Missing rest of sentence.

kgryte:
The special case of `pow(0,0)` is already addressed in the spec by:

> If `x2_i` is `+0`, the result is `1`, even if `x1_i` is `NaN`.
> If `x2_i` is `-0`, the result is `1`, even if `x1_i` is `NaN`.

kgryte:
Re: `divide`. If integer dtypes are to be explicitly permitted in the spec, we should state explicitly something to the effect of

> If `x1` has an integer dtype, the array must be cast to the default floating-point data type prior to computing results.

Ditto for `x2`.

If the goal is to match Python 3 `/` behavior exactly, then we'd need to recommend casting to `float64`, but this does not seem desirable.

asmeurer:
I made `pow` and `divide` only defined for floating-point inputs.

spec/API_specification/elementwise_functions.md
```diff
@@ -270,11 +271,11 @@ For floating-point operands,
 
 -   **x1**: _&lt;array&gt;_
 
-    -   input array corresponding to the y-coordinates.
+    -   input array corresponding to the y-coordinates. Should have a floating-point data type
```
kgryte, line 274:
```suggestion
    -   input array corresponding to the y-coordinates. Should have a floating-point data type.
```

spec/API_specification/data_types.md
```diff
@@ -78,3 +78,29 @@ Implementations may provide other ways to specify data types (e.g.,
 
 A conforming implementation of the array API standard may provide and support additional data types beyond those described in this specification.
 ```
+
+(data-type-categories)=
+## Data Type Categories
+
+For the purposes of this spec, the following data type categories are defined.
```
kgryte, line 85:
```suggestion
For the purposes of this specification, the following data type categories are defined.
```

To ensure consistent language across the specification, we've typically not relied on abbreviations.

spec/API_specification/data_types.md
```diff
@@ -78,3 +78,29 @@ Implementations may provide other ways to specify data types (e.g.,
...
+Libraries do not need to organize dtypes according to these categories. These
+are only for organizing the functions in this spec itself. Future versions of
+the spec will include additional categories for complex data types.
```
kgryte, line 88:
```suggestion
the specification will include additional categories for complex data types.
```

spec/API_specification/data_types.md
```diff
@@ -78,3 +78,29 @@ Implementations may provide other ways to specify data types (e.g.,
 
 A conforming implementation of the array API standard may provide and support additional data types beyond those described in this specification.
 ```
+
+(data-type-categories)=
+## Data Type Categories
+
+For the purposes of this spec, the following data type categories are defined.
+Libraries do not need to organize dtypes according to these categories. These
+are only for organizing the functions in this spec itself. Future versions of
```
kgryte, line 87:
```suggestion
are only for organizing the functions in this specification itself. Future versions of
```

spec/API_specification/data_types.md
```diff
@@ -78,3 +78,29 @@ Implementations may provide other ways to specify data types (e.g.,
...
+the spec will include additional categories for complex data types.
+
+### Numeric Data Types
+
+`int8`, `int16`, `int32`, `int64`, `uint8`, `uint16`, `uint32`,
```
kgryte, line 92: Should we make this a list (i.e., demarcated by bullets)?

spec/API_specification/elementwise_functions.md
```diff
@@ -10,6 +10,7 @@ A conforming implementation of the array API standard must provide and support t
 -   Optional parameters must be [keyword-only](https://www.python.org/dev/peps/pep-3102/) arguments.
 -   Broadcasting semantics must follow the semantics defined in {ref}`broadcasting`.
 -   Unless stated otherwise, functions must support the data types defined in {ref}`data-types`.
+-   Functions may only be required for a subset of input data type. Libraries may chose to implement functions for additional data types, but that behavior is not required by the spec. See {ref}`data-type-categories`.
```
kgryte, line 13:
```suggestion
-   Functions may only be required for a subset of input data type. Libraries may choose to implement functions for additional data types, but that behavior is not required by the specification. See {ref}`data-type-categories`.
```

Not entirely clear to me what is meant by the following sentence:

> Functions may only be required for a subset of input data type

spec/API_specification/elementwise_functions.md
```diff
@@ -763,13 +764,13 @@ Tests each element `x_i` of the input array `x` to determine whether the element
...
 
-    -   input array.
+    -   input array. Should have a numeric data type.
 
 #### Returns
 
 -   **out**: _&lt;array&gt;_
 
-    -   an array containing test results. An element `out_i` is `True` if `x_i` is `NaN` and `False` otherwise. The returned array must have a data type of `bool`.
+    -   an array containing test results. An element `out_i` is `True` if `x_i` is `NaN` and `False` otherwise. The returned array Should have a data type of `bool`.
```
kgryte, line 773:
```suggestion
    -   an array containing test results. An element `out_i` is `True` if `x_i` is `NaN` and `False` otherwise. The returned array should have a data type of `bool`.
```

spec/API_specification/data_types.md
rgommers: I think it's fine as is, clear enough and bullet lists will make this quite long.

rgommers (approved):
LGTM now. Thanks @asmeurer, and thanks @kgryte and @leofang for reviewing.

rgommers:
> I will also update the array methods as per [#98 (comment)](https://github.com/data-apis/array-api/issues/98#issuecomment-739088199) in another PR once this is merged.

Can you do this follow-up @asmeurer?

Issue #123, opened 2021-01-30
Type of return value from indexing


shoyer:
The current spec is silent on what the return value from an indexing operation should be: https://data-apis.github.io/array-api/latest/API_specification/indexing.html

I propose that the return value should always be another array object, i.e., so `__getitem__` could be type annotated as `__getitem__(self: Array, key: IndexKey) -> Array`. However, this is inconsistent with NumPy, which returns NumPy scalars in some cases when indexing with integers.

rgommers:
Thanks @shoyer. I agree with that proposal. This is what most libraries do I think - PyTorch, JAX, TensorFlow, Dask and MXNet (from memory, will verify later), with only CuPy agreeing (?) with NumPy here.

leofang:
CuPy always returns array too.

rgommers:
Thanks Leo. I didn't have a GPU machine at hand and couldn't find it in the docs. Turns out it's documented after all: https://docs.cupy.dev/en/stable/reference/difference.html#zero-dimensional-array

rgommers:
This was fixed in gh-125, closing.

PR #125, opened 2021-02-15
Address return value type for single-element access

kgryte:
This PR seeks to address the issue discussed in [gh-123](https://github.com/data-apis/array-api/issues/123) concerning single-element access by always returning an array, including when accessing a single element. This aligns with accelerator array libraries which always return arrays.

**Resolved**: the PR includes a TODO for the type hint for `__getitem__`, where `key` may be one of either `int`, `<array>`, slice, or selection tuple.

### DIFF
--- a/spec/API_specification/array_object.md
+++ b/spec/API_specification/array_object.md
@@ -458,7 +458,23 @@ Element-wise results must equal the results returned by the equivalent element-w
 (method-__getitem__)=
 ### \_\_getitem\_\_(x, key, /)
 
-_TODO: dependent on the indexing specification._
+Returns `x[key]`.
+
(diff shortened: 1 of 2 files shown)


rgommers:
> The PR includes a TODO for the type hint for `__getitem__`, where `key` may be one of either `int`, `<array>`, slice, or selection tuple. Not clear to me how we want to communicate the type hint in this case.

How about simply `Union[int, Tuple, slice, <array>]`?

kgryte:
Updated this PR to include the type annotation for the index key.

rgommers (approved):
LGTM, thanks @kgryte

PR #128, opened 2021-02-16
Add specification for computing the number of non-zero singular values of a matrix (linalg: matrix_rank)

kgryte:
This PR

-   specifies the interface for computing the rank (i.e., number of non-zero singular values) of a matrix.
-   is derived from comparing signatures across [array libraries](https://github.com/data-apis/array-api-comparison/blob/1d19815a4f3d1206d834d70770bbae137983344f/signatures/linalg/matrix_rank.md).

## Notes

-   CuPy and Torch (`torch.matrix_rank`) do not support providing stacks of matrices. Torch latest `master` does. This proposal follows NumPy and other linalg APIs in supporting stacks.

-   Dask does not have a `matrix_rank` API.

-   NumPy, MXNet, Torch (latest `master`) support a `hermitian` keyword argument to indicate a potential optimization. This keyword argument is not widely supported and is **not** included in this proposal.

-   Torch (latest `master`), CuPy, and possibly TF only support specifying `tol` as a `float`. This PR follows NumPy and other API proposals (e.g., `pinv` and `lstsq`) in allowing `tol` to be specified as either a `float` or an `array` of floats. Providing an `array` allows specifying one or more tolerances. This PR specifies tolerance as `rtol` to more explicitly indicate relative tolerance and to distinguish from `atol` (absolute tolerance).

-   TF supports a `validate_args` keyword argument. TF is alone in doing so, and such a keyword argument is **not** included in this proposal.


kgryte:

kgryte:
This PR has been open for some time without further comment and has been discussed/approved during meetings. Will merge, and we can submit follow-up PRs should any issues/concerns arise.

kgryte:
For reference, inclusion of this API was discussed [here](https://github.com/data-apis/array-api/issues/147#issuecomment-806664629).

PR #133, opened 2021-02-23
Small formatting fixes to make the spec parsable by the test suite scripts

asmeurer:


asmeurer:
I also pushed a fix to the in-place and reflected operators. 

I noticed we don't have `__divmod__` (https://docs.python.org/3/reference/datamodel.html#emulating-numeric-types). I don't know if it is important (though it can be easily be naively implemented as `__div__` + `__mod__`).

Also `__pow__` is missing the optional third argument `[modulo]`. I don't know if we want to consider adding that as well. Certainly it should not be disallowed.

spec/API_specification/array_object.md
```diff
@@ -396,12 +396,16 @@ Converts a zero-dimensional boolean array to a Python `bool` object.
 
 
 (method-__dlpack__)=
-### \_\_dlpack\_\_(/, *, stream=None)
+### \_\_dlpack\_\_(x, /, *, stream=None)
```
rgommers, line 399: This change is incorrect. `__dlpack__` is a method which returns a capsule with the data of the array inside. It does not take any positional parameters.

Same comment for `__dlpack_device__`.

asmeurer: The methods in this file include the `self` argument (except they are called `x` or `x1` instead of `self`, which I personally don't like, but that's the way it is). Unless this is supposed to be a static method, which isn't mentioned (if that's the case, why isn't it just a function?). But since it says "exports **the** array", I'm assuming that's not the case.

kgryte: This change is consistent with the rest of the array object specification.

Whether to document `self` was discussed previously. Originally, the spec approach was based on [NumPy](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.__abs__.html#numpy.ndarray.__abs__)'s and [TF](https://www.tensorflow.org/api_docs/python/tf/Tensor#methods_2)'s docs, which both documented `self`. Both were and still are inconsistent; NumPy less so, with now fewer dunder methods having documentation including `self` as a documented argument. Python [docs](https://docs.python.org/3/reference/datamodel.html#object.__lt__) do document self.

The use of `x` and `x#` over `self` and, e.g., `other`, was done following discussed spec guidance that recommended array argument names use consistent conventions across the spec. Functional APIs use `x` and `x#`, and that practice was continued for the array object methods.

However, not using `self` for the first argument of instance methods goes against [PEP 8](https://www.python.org/dev/peps/pep-0008/#function-and-method-arguments). Accordingly, I think we should adopt `self` and, where necessary, `other` for method names.

As this was my original doing, I can submit a PR that updates the array object method documentation.

rgommers: Agreed, we should either leave it out or use `self`, anything else will be very confusing.

asmeurer:
Since this PR is still open, I also pushed a fix to an issue I noticed with the indexing spec. It was ambiguous whether empty slices are required or optional. I removed the text that said they were required and left in the note that said they are optional. 

Assuming this is indeed what we want, should the NumPy array API namespace indexing support empty slices?

spec/API_specification/indexing.md
```diff
@@ -105,11 +105,6 @@ Slice syntax must have the following defaults. Let `n` be the axis (dimension) s
 -   If `k` is greater than `0` and `j` is not provided (e.g., `0::2`), `j` must equal `n`.
 -   If `k` is less than `0` and `i` is not provided (e.g., `:10:-2`), `i` must equal `n-1`.
 -   If `k` is less than `0` and `j` is not provided (e.g., `0::-2`), `j` must equal `-n-1`.
-
-Using a slice to index a single array axis must adhere to the following rules. Let `n` be the axis (dimension) size.
```
kgryte, line 109: This line should not be removed. Otherwise, the item defined by L113 will look like it belongs to the list ending at L107.

kgryte:
To split hairs a bit, the rule you removed was referring to the case where `i == j`; while the potential conflict arises in the note regarding out-of-bounds indexing (e.g., where an index exceeds dimensions). It is not clear to me that the two cases are the same thing.

asmeurer:
Oh I think you are right. I will revert the commit.

spec/API_specification/indexing.md
asmeurer: Well that was also confusing to me. The two lists both seem like they are about the same thing (the behavior of a slice on an axis of size `n`).

asmeurer:
I guess the reason it confused me is that this isn't the only way to get an empty slice. As long as the element indexed by stop is before than the start (or after for negative step), the slice is empty. For example, `a[3:1]`. 

I'm unclear which cases we should allow or disallow in the NumPy array API. Should we disallow any slice that results in an empty selection except for the case where start == stop?

spec/API_specification/array_object.md
rgommers: The `self` change was made, resolving.

rgommers: Almost merged, in gh-135

rgommers (approved):
LGTM now, thanks @asmeurer and @kgryte

PR #134, opened 2021-03-01
Add specification for computing the matrix product (linalg: matmul)

kgryte:
This PR

-   specifies the interface for computing the matrix product.
-   is derived from comparing signatures across [array libraries](https://github.com/data-apis/array-api-comparison/blob/b6659bc3655abb1aa784d6edffe3d0b4ae260496/signatures/linalg/matmul.md).

## Notes

-   NumPy supports `gufunc` keyword arguments: `out`, `casting`, `dtype`, `subok`, `order`, etc.

-   NumPy, CuPy, Torch, MXNet support providing an output array.

-   JAX supports specifying a `precision` keyword argument.

-   TF supports various keywords for transposing array arguments (`transpose_a`, `transpose_b`), conjugating array arguments (`adjoint_a`, `adjoint_b`), and indicating whether an array argument is sparse (`a_is_sparse`, `b_is_sparse`). TF is alone in supporting these arguments.

-   This PR follows [PEP 465](https://www.python.org/dev/peps/pep-0465) and the semantics of the built-in `@` operator.

### DIFF
--- a/spec/API_specification/array_object.md
+++ b/spec/API_specification/array_object.md
@@ -725,7 +725,7 @@ Element-wise results must equal the results returned by the equivalent element-w
 ```
 
 (method-__matmul__)=
-### \_\_matmul\_\_(x1, x2, /)
+### \_\_matmul\_\_(self, other, /)
 
 Computes the matrix product.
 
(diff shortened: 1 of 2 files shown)


asmeurer:
Should we require that 0-dimensional arrays give an error? Note that the PEP says they should.

spec/API_specification/array_object.md
```diff
@@ -727,23 +727,36 @@ Element-wise results must equal the results returned by the equivalent element-w
...
+
+The `matmul` function must implement the same semantics as the built-in `@` operator (see [PEP 465](https://www.python.org/dev/peps/pep-0465)).
+```
 
 #### Parameters
 
 -   **x1**: _&lt;array&gt;_
 
-    -   array instance.
+    -   array instance. Should have a numeric data type. If `x1` is one-dimensional having shape `(M)` and `x2` has more than one dimension, `x1` must be promoted to a two-dimensional array by prepending `1` to its dimensions (i.e., must have shape `(1, M)`). After matrix multiplication, the prepended dimensions in the returned array must be removed. If `x1` has more than one dimension (including after vector-to-matrix promotion), `x1` must be compatible with `x2` (see {ref}`broadcasting`). If `x1` has shape `(..., M, K)`, the innermost two dimensions form matrices on which to perform matrix multiplication.
```
rgommers, line 741: Can we just refer to the `matmul` documentation, rather than duplicating all this content?

kgryte: We can; however, to be consistent, we'd need to do that for all other dunder methods. Personally, I am fine with duplication, as it requires readers to take fewer hops to understand the semantics for dunder APIs.

kgryte:
@asmeurer Thanks for the suggestion. I have documented the exceptions in accordance with the PEP (0d) and NumPy (nd).

kgryte:
Latest merge resolves build failures. CI is now passing.

kgryte:
This PR has been open for some time without comment and has been discussed/approved during meetings. Will merge, and we can submit follow-up PRs should any issues/concerns arise.

PR #135, opened 2021-03-08
Update specification to use `self` in array object APIs

kgryte:
This PR

-   Ensures conformance with [PEP 8](https://www.python.org/dev/peps/pep-0008/#function-and-method-arguments) in always using `self` in method argument signatures. The inclusion of `self` in the dunder signatures follows official Python [documentation](https://docs.python.org/3/reference/datamodel.html#object.__lt__).


rgommers (approved):
All green now, merging. Thanks @kgryte

PR #136, opened 2021-03-08
Add specification for computing a tensor contraction (linalg: tensordot)

kgryte:
This PR

-   specifies the interface for performing a tensor contraction.
-   is derived from comparing signatures across [array libraries](https://github.com/data-apis/array-api-comparison/blob/e59db40311ad629471a4f482094ef2156af9f063/signatures/linalg/tensordot.md).

## Notes

-   JAX supports providing a `precision` keyword argument to control matrix-multiplication precision. It is alone in doing so among compared array libraries.

-   TF allows providing a `(2,K)` shaped `array` for the `axes` keyword argument. NumPy et al support array-like values or lists when specifying multiple axes. This proposal follows NumPy et al in only allowing `int` and `Sequence`.

-   TF returns an array whose data type is that of `x1`. This proposal requires that the result follow type promotion rules.

### DIFF
--- a/spec/API_specification/linear_algebra_functions.md
+++ b/spec/API_specification/linear_algebra_functions.md
@@ -284,6 +284,8 @@ TODO
 
 Returns a tensor contraction of `x1` and `x2` over specific axes.
 
+#### Parameters
+
 -   **x1**: _&lt;array&gt;_
 
     -   first input array. Should have a numeric data type.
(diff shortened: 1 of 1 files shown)


kgryte:
This PR has been open for some time without comment and discussed/approved during meetings. Will merge and we can submit follow-up PRs should any issues/concerns arise.

PR #140, opened 2021-03-09
Add __array_namespace__ as a required method on the array object

asmeurer:


spec/purpose_and_scope.md
```diff
@@ -362,11 +363,10 @@ standard can become final. We prefer to delay this decision, to see how
 NEP 37 adoption will work out.
 ```
 
-The `xp` namespace must contain the array object and all functionality
```
asmeurer, line 365: I removed "array object" here, because the array class is not itself required to be a name in the namespace.

spec/API_specification/array_object.md
```diff
@@ -377,6 +377,29 @@ Evaluates `x1_i & x2_i` for each element `x1_i` of an array instance `x1` with t
...
+
+-   **api_version**: _&lt;Optional\[str\]&gt;_
+
+    -   string representing the version of the array API specification to be returned, in `'YYYY.MM'` form, for example, `'2020.10'`. If it is `None`, it should return the namespace corresponding to latest version of the array API specification.  If the given version is invalid or not implemented for the given module, an error should be raised. Default: `None`.
+
+#### Returns
+
+-   **out**: _&lt;object&gt;_
+
+    -   an object representing the array API namespace. It should have every top-level function defined in the specification as an attribute. It may contain other public names as well, but it recommended to only include those names that are part of the specification.
```
leofang, line 400:
```suggestion
    -   an object representing the array API namespace. It should have every top-level function defined in the specification as an attribute. It may contain other public names as well, but it is recommended to only include those names that are part of the specification.
```

leofang (changes requested):

PR #148, opened 2021-03-24
Update dtype requirements for existing linear algebra APIs

kgryte:
This PR

-   updates and/or adds dtype requirements for existing linear algebra APIs. Following element-wise APIs, the use of `must` is replaced with `should` in order to allow some flexibility in accepted dtypes when warranted.


rgommers (approved):
Looks fine to me, thanks @kgryte

leofang (approved):

PR #149, opened 2021-03-25
Add linear algebra design principles

kgryte:
This PR

-   adds a section on design principles for linear algebra APIs.

### DIFF
--- a/spec/API_specification/linear_algebra_functions.md
+++ b/spec/API_specification/linear_algebra_functions.md
@@ -11,6 +11,46 @@ A conforming implementation of the array API standard must provide and support t
 -   Unless stated otherwise, functions must adhere to the type promotion rules defined in {ref}`type-promotion`.
 -   Unless stated otherwise, floating-point operations must adhere to IEEE 754-2019.
 
+## Design Principles
+
+A principle goal of this specification is to standardize commonly implemented interfaces among array libraries. While this specification endeavors to avoid straying too far from common practice, this specification does, with due restraint, seek to address design decisions arising more from historical accident than first principles. This is especially true for linear algebra APIs, which have arisen and evolved organically over time and have often been tied to particular underlying implementations (e.g., to BLAS and LAPACK).
+
+Accordingly, the standardization process affords the opportunity to reduce interface complexity among linear algebra APIs by inferring and subsequently codifying common design themes, thus allowing more consistent APIs. What follows is the set of design principles governing the APIs which follow:
+
+1.  **Batching**: if an operation is explicitly defined in terms of matrices (i.e., two-dimensional arrays), then the associated interface should support "batching" (i.e., the ability to perform the operation over a "stack" of matrices). Example operations include:
+
+    -   `inv`: computing the multiplicative inverse of a square matrix.
+    -   `cholesky`: performing Cholesky decomposition.
+    -   `matmul`: performing matrix multiplication.
+
+2.  **Data types**: if an operation requires decimal operations and {ref}`type-promotion` semantics are undefined (e.g., as is the case for mixed-kind promotions), then the associated interface should be specified as being restricted to floating-point data types. While the specification uses the term "SHOULD" rather than "MUST", a conforming implementation of the array API standard should only ignore the restriction provided overly compelling reasons for doing so. Example operations which should be limited to floating-point data types include:
+
+    -   `inv`: computing the multiplicative inverse.
+    -   `slogdet`: computing the natural logarithm of the absolute value of the determinant.
+    -   `norm`: computing the matrix or vector norm.
+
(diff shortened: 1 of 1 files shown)


spec/API_specification/linear_algebra_functions.md
```diff
@@ -11,6 +11,46 @@ A conforming implementation of the array API standard must provide and support t
...
+
+Accordingly, the standardization process affords the opportunity to reduce interface complexity among linear algebra APIs by inferring and subsequently codifying common design themes, thus allowing more consistent APIs. What follows is the set of design principles governing the APIs which follow:
+
+1.  **Batching**: if an operation is explicitly defined in terms of matrices (i.e., two-dimensional arrays), then the associated interface should support "batching" (i.e., the ability to perform the operation over a "stack" of matrices). Example operations include:
+
+    -   `inv`: computing the multiplicative inverse of a square matrix.
+    -   `cholesky`: performing Cholesky decomposition.
+    -   `matmul`: performing matrix multiplication.
```
rgommers, line 24: The unfortunate absence of a matrix transpose again comes to mind. May be noted explicitly here perhaps, so that's a gap that may be filled in libraries and then added in the future.

rgommers (commented):
Looks good, thanks @kgryte. Two other things come to mind:

1. Namespaces. NumPy has the weird split between in linalg functions between its main namespace and `numpy.linalg`. Other libraries tend to follow that. There's no optimal solution; we should say something about it here though.
2. You may want to include a part of gh-137. A principle may be "orthogonality", which is a reason for including only certain "dot" functions.

spec/API_specification/linear_algebra_functions.md
kgryte: I could be missing something, but batch matrix transpose can be achieved now via the current `transpose` API by providing an axes permutation in which only the last two dimensions are swapped. As in TF, a `matrix_transpose` method supporting batching could be defined as such and simply be shorthand.

kgryte:
@rgommers I've added an orthogonality principle and a note regarding namespaces.

spec/API_specification/linear_algebra_functions.md
```diff
@@ -11,8 +11,74 @@ A conforming implementation of the array API standard must provide and support t
...
+    -   `matmul`: matrix multiplication (dot product for two vectors).
+    -   `einsum`: Einstein summation convention.
+
+    where `dot` is overloaded based on input array dimensionality and `vdot` and `inner` exhibit a high degree of overlap with other interfaces. By consolidating interfaces and more clearly delineating behavior, this specification aims to ensure that each interface has a unique purpose and defined use case.
+
 ## Objects in API
 
+```{note}
+
+Historically, linear algebra functionality has been split across interfaces belonging to different namespaces. For example, NumPy exposes linear algebra interfaces in both its top-level namespace and its `linalg` sub-namespace. The determination of which linear algebra interface belongs to which namespace is often unintuitive and a common source of confusion.
```
shoyer, line 77: What exists in `numpy.linalg` can be understood as based on upon what requires a matrix factorization and what doesn't, i.e., LAPACK vs BLAS.

This is arguably a pretty important distinction for both users and array libraries as well. The matrix factorization stuff is much harder to implement and has more niche use-cases for scientific computing. In contrast, the stuff that doesn't (even including matmul) is easier to implement and useful even for more generic data analysis use-cases, e.g., computing a weighted average. Libraries like xarray and dask would use/implement only the non-factorization stuff.

`matmul` is a bit of a gray area (a naive loop particularly inefficient, compared to an optimized implementations), but I think there's a strong case for including it in the main namespace for a few reasons:
1. It's a really fundamental building block. Every array library implements it and you need it both for almost all scientific computing use-cases and also deep learning. Even data analysis libraries like xarray use it (though perhaps they don't really _need_ it).
2. The algorithms for efficient matmul (at least in practice) tend to be relatively simple in contrast to matrix factorization -- the main question is how to implement the method efficiently, not what method can be used. The efficiency trade-off is more similar to the use of SIMD instructions for vectorized arithmetic than a matrix factorization.

rgommers: > This is arguably a pretty important distinction for both users and array libraries as well.

I would say for libraries it is, for users it is not. Users don't care about how easy something is to implement. And saying, e.g., that `tensordot` or `einsum` is more important or fundamental than `svd` or `qr` is (imho) wrong. Only physicists/mathematicians for example tend to know Einstein notation - and they'd certainly also know and use `svd` and `qr`.

I think that's the main issue I have with the current split in NumPy. API design driven from implementation details up tends to be suboptimal.

rgommers: One idea could be to have the `linalg` namespace contain both the factorization set and the linalg functions in the main namespace. That'd be much clearer to users. For all regular array libraries that anyway ship with BLAS/LAPACK: only use the `linalg` namespace, everything is there.

kgryte:
I've removed the note, and will submit a follow-up PR regarding the mirroring of top-level linear algebra APIs in the `linalg` submodule as discussed during recent meetings.

PR #160, opened 2021-04-12
Add specification for computing singular values using singular value decomposition (linalg: svdvals)

kgryte:
This PR

-   adds a specification for computing singular values using singular value decomposition.

## Notes

-   This API deviates from current array libraries which overload the `svd` function modeled after NumPy via a `compute_uv` keyword. In that API, when `compute_uv` is `False`, only the singular values are returned. As discussed in a thread on [gh-114](https://github.com/data-apis/array-api/pull/114#issuecomment-801680942) and in a prior weekly meeting, in order to ensure API consistency with `eig` and `eigvals` and to limit polymorphism, splitting `svd` into `svd` and `svdvals` was deemed desirable. The former API is equivalent to `compute_uv=True`, while the latter is equivalent to `compute_uv=False`.

### DIFF
--- a/spec/API_specification/linear_algebra_functions.md
+++ b/spec/API_specification/linear_algebra_functions.md
@@ -279,6 +279,23 @@ TODO
 
 TODO
 
+(function-svdvals)=
+### svdvals(x, /)
+
+Computes the singular values of a matrix (or stack of matrices) `x`.
+
+#### Parameters
+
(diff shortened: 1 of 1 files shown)


kgryte:
This PR has been discussed and approved during meetings. Will merge, and we can submit follow-up PRs should any issues/concerns arise.

PR #162, opened 2021-04-12
Add specification for computing the eigenvalues of a symmetric matrix (linalg: eigvalsh)

kgryte:
This PR

-   adds a specification for computing the eigenvalues of a real symmetric matrix.

## Notes

-   In contrast to NumPy's `UPLO` keyword pattern (following LAPACK), this PR follows [Torch v1.7.1 symeig](https://pytorch.org/docs/1.7.1/generated/torch.symeig.html) and uses an `upper` keyword to have better consistency with other linalg spec'd functions.

### DIFF
--- a/spec/API_specification/linear_algebra_functions.md
+++ b/spec/API_specification/linear_algebra_functions.md
@@ -112,8 +112,8 @@ TODO
 
 TODO
 
-(function-eigvalsh)=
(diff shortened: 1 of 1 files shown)


kgryte:
This PR has been discussed and approved during meetings. Will merge, and we can submit follow-up PRs should any issues/concerns arise.

Issue #84, opened 2020-11-11
Boolean array indexing is not compatible with static memory allocation


shoyer:
JAX is built on top of XLA which (currently) requires static memory allocation for operations. This means it is not possible to express operations like `x[y]` where `y` is a boolean, because the resulting array has value dependent size: https://data-apis.github.io/array-api/latest/API_specification/indexing.html#boolean-array-indexing

It's definitely possible that the static memory allocation requirement could be relaxed in the future, but dynamic memory allocation is always going to be harder to implement in a performant way. For example, I would guess Numba also struggles with this sort of operation. I don't think we should require it for array libraries implementing our standard, since it isn't needed for the majority of array operations.

TomAugspurger:
Dask’s lazy evaluation struggles with this a bit too. The output shape of a Boolean mask will have NaN for the shape in that dimension, and .size will be NaN.

rgommers:
> For example, I would guess Numba also struggles with this sort of operation

Numba does support boolean indexing, but only for 1-D arrays - and you cannot mix it with other advanced indexing.

rgommers:
Related JAX issues with discussion:

- https://github.com/google/jax/issues/166
- https://github.com/google/jax/issues/2765

It does seem that there's very limited support for boolean indexing (only static boolean arrays, which are not all that useful in practice), and none for jitted functions.

This led me to wonder about other functions where output shape depends on values in the input, such as `unique`. That is implemented, but also not jittable: https://github.com/google/jax/pull/2760

TensorFlow is in a similar boat, but provides a [boolean_mask function](https://www.tensorflow.org/api_docs/python/tf/boolean_mask) instead. And there's an [open PR for JAX](https://github.com/google/jax/pull/2608) to add `index_mask` which looks similar.

rgommers:
`nonzero` and `where` are also problematic here.

rgommers:
I _think_ that's the complete list of functionality where shapes depend on values - so maybe that's a good set to separate out somehow. They are fairly heavily used functions, so would be good to dig up some examples - say in SciPy or scikit-learn - and figure out if those are easy to rewrite.

rgommers:
Of course this can happen with basic indexing as well:
```
x = arange(5)
idx_max = random.randint(0, 5)
x2 = x[:idx_max]
```
So there needs to be something said about the principle of shapes depending on values.

rgommers:
The summary of a discussion we had a few days ago:
[...]

PR #164, opened 2021-04-14
Clarify that boolean array indexing requirements

asmeurer:
We only require that boolean array indexing be supported when the array is the
sole index. The previous text was ambiguous, and could be read as supporting
boolean arrays plus other non-array indices as part of a larger
multidimensional index. However, this is complicated in general, as it can
lead to some corner cases. See the discussion in #84.

With that being said, it is still an open discussion whether we should allow
some subset of boolean array indices along with other non-array indices in the
same index. I believe regardless, what we do not want to allow is:

- multiple boolean arrays (this requires broadcasting semantics, which are
  confusing for boolean arrays)
- boolean arrays and integers separated by a non-integer index (for example
  a[b, :, 0] where b is a boolean array). This is the corner case currently
  implemented by NumPy which we want to avoid for the standard.

### DIFF
--- a/spec/API_specification/indexing.md
+++ b/spec/API_specification/indexing.md
@@ -160,7 +160,7 @@ _Rationale: this is consistent with bounds-checking for single-axis indexing. An
 
 ## Boolean Array Indexing
 
-An array must support indexing via a **single** `M`-dimensional boolean array `B` with shape `S1 = (s1, ..., sM)` according to the following rules. Let `A` be an `N`-dimensional array with shape `S2 = (s1, ..., sM, ..., sN)`.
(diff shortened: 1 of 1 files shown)


kgryte (approved):
LGTM. Thanks, @asmeurer!

rgommers (approved):
This looks like a useful textual clarification for what the current version of the text aims to express, so let's merge it.

> With that being said, it is still an open discussion whether we should allow
some subset of boolean array indices along with other non-array indices in the
same index. I believe regardless, what we do not want to allow is:

Agreed. Let's continue that conversation on gh-84.

rgommers:
Thanks @asmeurer

PR #166, opened 2021-04-15
Note that in-place operators should not change the dtype of the array

asmeurer:
Also note that the Python scalar rules do apply to in-place operators in
addition to normal operators.

The broadcasting rule here is duplicated in the broadcasting section. I'm
unsure if we should leave it duplicated or remove it so it is only mentioned
once.


kgryte (approved):
LGTM

leofang (approved):

rgommers (approved):
LGTM too, thanks @asmeurer

Issue #154, opened 2021-03-27
Consider removing array constants


shoyer:
I stumbled across [this page](https://data-apis.org/array-api/latest/API_specification/constants.html), which lists four constants in the spec: `e`, `inf`, `nan` and `pi`.

I'm not strongly opposed, but it isn't really clear to me why we need these in the spec when they already exist in Python's builtin `math` built. Given that these should be builtin Python float objects, it isn't clear to me what we gain from including these, except saving users another import?

rgommers:
Mostly convenience indeed. Every library has them, users expect them, and they're used a lot. Writing
```
y = xp.sin(x + xp.pi/4)
```
seems more natural than `y = xp.sin(x + math.pi/4)`. 

These would otherwise the only things we'd rely on that are not Python builtins I believe.

shoyer:
I'm not sure what you mean by "every library has them." As far as I can tell, only libraries that directly try to copy NumPy's API do. E.g., TensorFlow does not.

rgommers:
Argh, you're right - PyTorch also doesn't. Although I'm not sure why not, the test suite is full of `nan = float('nan')` which is pretty clumsy.

rgommers:
I could live with leaving these out, but personally lean towards leaving them in for convenience. And because I don't like the `math` module much, would rather not rely on anything in the stdlib.

rgommers:
Would be nice to hear what maintainers of other libraries think. From a NumPy and its downstream libraries perspective these are expected, but if there's strong feelings especially from the libraries that don't have them then it could go the other way. I checked with PyTorch, they'd be fine with an alias. @edloper for TensorFlow, any problem with adding these constants?

edloper:
I don't have a strong opinion on whether these constants be included in the spec or not.

If they are included in the spec, then are they required to be python float values?  Or array values?  Or is it up to the library?  If array values, then are there any requirements on what dtype they have?  Note that many libraries don't do automatic casting between dtypes, so if a library defines pi to be a float32 scalar, then users can't combine it with a float64 value (without casting), and vice versa.

shoyer:
I think they should be required to be Python `float` values (which effectively means `float64`).

asmeurer:
> Note that many libraries don't do automatic casting between dtypes, so if a library defines pi to be a float32 scalar, then users can't combine it with a float64 value (without casting), and vice versa.

[...]

PR #169, opened 2021-04-26
Clarify that constants are Python scalars

kgryte:
This PR

-   clarifies that constants are Python scalars.

Currently, the spec does not indicate how the constants should be exposed (e.g., Python scalars, zero-dimensional arrays, etc). This PR aims to remedy this.

The choice of Python scalar, rather than as a zero-dimensional array with a dtype described in the specification, stems from greater flexibility in how these constants participate in operations.

For example, according to the current specification, `pi` would be truncated when participating in an operation with an array having an integer dtype, which may not be the case if `pi` was a zero-dimensional `float64` array and participating in mixed-type promotion.

Further, as a Python scalar, `pi` would be converted to a `float32` array when participating in an operation with an array having a `float32` dtype, which would not be the case if `pi` was a zero-dimensional `float64` array.

### DIFF
--- a/spec/API_specification/constants.md
+++ b/spec/API_specification/constants.md
@@ -2,7 +2,9 @@
 
 > Array API specification for constants.
 
-A conforming implementation of the array API standard must provide and support the following constants.
+A conforming implementation of the array API standard must provide and support the following constants adhering to the following conventions.
+
+-   Each constant must have a Python numeric data type (i.e., `int`, `float`, or `complex`) and be provided as a Python scalar value.
 
 <!-- NOTE: please keep the constants in alphabetical order -->
 
@@ -11,7 +13,7 @@ A conforming implementation of the array API standard must provide and support t
 (constant-e)=
 ### e
 
-Euler's constant.
+IEEE 754 floating-point representation of Euler's constant.
 
 ```text
 e = 2.71828182845904523536028747135266249775724709369995...
@@ -30,7 +32,7 @@ IEEE 754 floating-point representation of Not a Number (`NaN`).
 (constant-pi)=
 ### pi
 
-The mathematical constant `π`.
+IEEE 754 floating-point representation of the mathematical constant `π`.
 
 ```text
 pi = 3.1415926535897932384626433...
--- a/spec/API_specification/constants.md
+++ b/spec/API_specification/constants.md
@@ -4,7 +4,7 @@
 
 A conforming implementation of the array API standard must provide and support the following constants adhering to the following conventions.
 
--   Each constant must have a Python numeric data type (i.e., `int`, `float`, or `complex`) and be provided as a Python scalar value.
+-   Each constant must have a Python floating-data type (i.e., `float`) and be provided as a Python scalar value.
 
 <!-- NOTE: please keep the constants in alphabetical order -->
 
--- a/spec/API_specification/constants.md
+++ b/spec/API_specification/constants.md
@@ -4,7 +4,7 @@
 
 A conforming implementation of the array API standard must provide and support the following constants adhering to the following conventions.
 
--   Each constant must have a Python floating-data type (i.e., `float`) and be provided as a Python scalar value.
+-   Each constant must have a Python floating-point data type (i.e., `float`) and be provided as a Python scalar value.
 
 <!-- NOTE: please keep the constants in alphabetical order -->
 


rgommers (approved):
LGTM, thanks @kgryte

leofang (approved):

leofang:
By merging this PR, I assume we can safely ignore #154?

rgommers:
> By merging this PR, I assume we can safely ignore #154?

Don't know, could still be relevant. Would be nice to see a few more opinions on it. Let me comment on it.

spec/API_specification/constants.md
```diff
@@ -2,7 +2,9 @@
 
 > Array API specification for constants.
 
-A conforming implementation of the array API standard must provide and support the following constants.
+A conforming implementation of the array API standard must provide and support the following constants adhering to the following conventions.
+
+-   Each constant must have a Python numeric data type (i.e., `int`, `float`, or `complex`) and be provided as a Python scalar value.
```
asmeurer, line 7: Shouldn't they all specifically be `float`?

asmeurer: Also, any reason to not just require exact values for `pi` and `e`, like

```py
pi = 3.141592653589793
e = 2.718281828459045
nan = float('nan')
inf = float('inf')
```

kgryte: The descriptions for each constant state the required Python numeric data type. In general, as a matter of convention, a constant must be a Python scalar. This section is intended to cover both current and potential use cases where constants may be either `float` or `int`.

asmeurer: You mean if a library includes constants that are outside of the spec? Or do you mean for potential future versions of the spec? Including `int` and `complex` here just seems confusing to me since all four constants are floats.

kgryte: Future versions.

I am not sure the source of the confusion. The general principle states that a constant must have a numeric data type and then, for each constant, we specify the exact data type for that constant. The spec, as currently written, does not say that constants must be provided as an `int`, `float`, AND `complex`. Nor is the option available to choose which type is desired, given that the descriptions for each of the current constants state that they must be IEEE 754 floats.

kgryte: Regarding exact values for `pi` and `e`, to do so, we'd need to specify the exact bit sequence. From C11 6.4.4.2/3

> When `FLT_RADIX` is not a power of `2`, the result is either the nearest representable value, or the larger or smaller representable value immediately adjacent to the nearest representable value, chosen in an implementation-defined manner.

kgryte: How the constants are shown in the spec follows NumPy's [docs](https://numpy.org/doc/stable/reference/constants.html#numpy.e).

rgommers: I agree with @asmeurer here, if we have 4 constants that are all `float`, it makes little sense to include `int` and `complex`. It's also not future-proof; the only non-`float` constant in numpy is an alias for `None`, so not numerical.

asmeurer: That's the C spec, but in Python floats exactly round-trip from their string representation (though I'm actually having a hard time finding documentation of this fact). However, we can also specify a hex representation using (see `float.hex` and `float.fromhex`).

kgryte: @asmeurer Round tripping from a string representation is not guaranteed due to floating-point...

```
In [1]: float("3.141592653589793")                                                                                                          
Out[1]: 3.141592653589793

In [2]: float("3.141592653589792")                                                                                                          
Out[2]: 3.141592653589792

In [3]: float("3.1415926535897925")                                                                                                         
Out[3]: 3.1415926535897927

In [4]: float("3.1415926535897932")                                                                                                         
Out[4]: 3.141592653589793
```

However, going the other way should round trip (i.e., map to the same bit sequence), but this is tricky when the value is not already rounded as shown below...

```
In [5]: str(float("3.1415926535897925"))                                                                                                    
Out[5]: '3.1415926535897927'

In [6]: float(str(3.1415926535897927))                                                                                                      
Out[6]: 3.1415926535897927

In [7]: str(3.1415926535897925)                                                                                                             
Out[7]: '3.1415926535897927'

In [8]: float(str(3.1415926535897925))                                                                                                      
Out[8]: 3.1415926535897927
```

The point regarding the C spec is that the last digit can vary depending on the platform (i.e., is implementation-defined). I don't think Python is exceptional here.

The reason for saying **an** IEEE 754 floating-point representation of `3.1415926535897932384626433...` is to acknowledge that the last bit may not be the same across all platforms.

kgryte: Updated language in latest commits.

PR #172, opened 2021-04-26
Add specification for `__setitem__`

kgryte:
This PR

-   adds a specification for `__setitem__`.

### DIFF
--- a/spec/API_specification/array_object.md
+++ b/spec/API_specification/array_object.md
@@ -1046,7 +1046,7 @@ Sets `self[key]` to `value`.
 
 -   **value**: _Union\[ int, float, bool, &lt;array&gt; ]_
 
-    -   value(s) to set.
+    -   value(s) to set. Must be compatible with `self[key]` (see {ref}`broadcasting`).
 
 ```{note}
 
--- a/spec/API_specification/array_object.md
+++ b/spec/API_specification/array_object.md
@@ -1032,7 +1032,30 @@ Element-wise results must equal the results returned by the equivalent element-w
 (method-__setitem__)=
 ### \_\_setitem\_\_(self, key, value, /)
 
-_TODO: dependent on the indexing specification._
+Sets `self[key]` to `value`.
+
+#### Parameters
+
+-   **self**: _&lt;array;&gt;_
+
+    -   array instance.
+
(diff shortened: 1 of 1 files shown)


spec/API_specification/array_object.md
```diff
@@ -1032,7 +1032,30 @@ Element-wise results must equal the results returned by the equivalent element-w
...
+
+    -   value(s) to set.
+
+```{note}
+
+Setting array values must not affect the data type of `self`.
+
+When `value` is a Python scalar (i.e., `int`, `float`, `bool`), behavior must follow specification guidance on mixing arrays with Python scalars (see {ref}`type-promotion`).
+
+When `value` is an `array` of a different data type than `self`, how values are cast to the data type of `self` is implementation defined.
```
asmeurer, line 1057: Why is this? I would have expected the value to just cast to the dtype of self (or error if that isn't allowed).

asmeurer:
Should this mention broadcasting?

spec/API_specification/array_object.md
kgryte: I had in mind how `int <-> uint` conversions would work for overflow/underflow. Do values wrap, error when exceeding bounds, or something else? When converting from a float to an int, are values truncated, trigger an exception, etc, given that how arrays can handle mixed kind ops is implementation defined.

kgryte:
@asmeurer Yes, I will add a note that `value` must be broadcast compatible with `self`. Thanks!

asmeurer:
It should be broadcast compatible with `self[key]`.

spec/API_specification/array_object.md
rgommers: I agree that this is hard to make uniform. Example:
```
In [5]: x = np.linspace(0, 1)

In [6]: x[:2] = np.array([1, 1+1.j])
ComplexWarning: Casting complex values to real discards the imaginary part

In [7]: x[:4] = 1+1.j
ComplexWarning: Casting complex values to real discards the imaginary part
```
I'd hope that that errors in other implementations.

kgryte: Looks like NumPy is following C [semantics](https://en.cppreference.com/w/c/language/conversion) in the above example. But certainly, other implementations should be free to disallow and raise an exception.

rgommers (approved):
LGTM. This looks relatively straightforward and was a clear gap, so merging. Thanks @kgryte and @asmeurer

Issue #143, opened 2021-03-11
Some issues with type annotations


asmeurer:
Some issues with type annotations that were brought up by @Bvb93 at https://github.com/numpy/numpy/pull/18585

- [x] For `asarray`, it isn't clear how to define `NestedSequence` (`typing` annotations don't support recursive types).
- [x] `asarray` needs to also include `array` as a valid type for `obj` (`array` should not be considred a subtype of `NestedSequence`)
- [x] The pipe operator for types isn't valid yet (https://www.python.org/dev/peps/pep-0604/) (it's used in the `NestedSequence` type in `asarray`)
- [x] For `stack` and `concat`, the type for `arrays` should be `Tuple[array, ...]` instead of `Tuple[array]`
- [x] For `stack` and `concat` (and possibly others), `Tuple` is too restrictive vs. `Sequence`.
- [x] `unique` can be implemented as an `@overload` (see https://github.com/numpy/numpy/pull/18585#discussion_r591757472). I don't know if this needs to be changed in the spec.
- [x] The return type for `finfo` as `finfo` doesn't make sense if `finfo` is a function (see https://github.com/numpy/numpy/pull/18585#discussion_r591763368).
- [x] `__getitem__` and `__setitem__` should include the `ellipsis` type (https://github.com/numpy/numpy/pull/18585#discussion_r592256384)
- [x] ~~`__len__` should return `int`~~ (**update**: `__len__` is slated for removal from the spec; see https://github.com/data-apis/array-api/pull/289)
- [x] The return type for `shape` is currently `Union[ Tuple[ int, …], <shape> ]`, which needs to be addressed (there is a TODO in the spec for this). (**update**: see https://github.com/data-apis/array-api/pull/289)
- [x] ~~Some of the other types for properties also use `<...>` types. We should make all the type annotations be valid Python/mypy that can be used in the annotations in the signature.~~
- [x] ~~Should the array object subclass from `typing.Protocol` (https://github.com/numpy/numpy/pull/18585#discussion_r592261549)? (I don't know if this is relevant for the spec).~~
- [x] The dunder methods (`__add__`, and so on) should include `int`, `float`, (and `bool` as appropriate) as allowed types for `other`.

rgommers:
Thanks for the great feedback @BvB93, and for collecting it here @asmeurer. I'll try and fix the straightforward ones now.

[...]

PR #173, opened 2021-04-26
Update typing for array object dunder methods

kgryte:
This PR

-   updates type hints for various array object dunder methods, as mentioned in [gh-143](https://github.com/data-apis/array-api/issues/143) ([comment](https://github.com/data-apis/array-api/issues/143#issuecomment-811487190)).
-   updates parameter descriptions to match element-wise counterparts (e.g., `MUST` vs `SHOULD`, etc).

### DIFF
--- a/spec/API_specification/array_object.md
+++ b/spec/API_specification/array_object.md
@@ -288,7 +288,7 @@ For floating-point operands, let `self` equal `x`.
 
 -   **self**: _&lt;array&gt;_
 
-    -   array instance.
+    -   array instance. Should have a numeric data type.
 
 #### Returns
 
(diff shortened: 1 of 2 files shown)


rgommers (approved):
Thanks @kgryte. LGTM.

Technically `int` is a subtype of `float` so not necessary to include (and there's a mypy guideline saying to not do `Union[int, float]` when you can simply do `float`, however for clarity I think it's preferred to do so here anyway.

asmeurer:
> Technically int is a subtype of float so not necessary to include (and there's a mypy guideline saying to not do Union[int, float] when you can simply do float, however for clarity I think it's preferred to do so here anyway.

`__pow__` and `__divide__` in the spec currently only require floating-point, meaning for the NumPy API, they error on `int` inputs. Is that actually correct behavior based on the spec? If so, this is a breakage from the mypy typing. Does mypy have a way to specify float but not int?

rgommers:
> `__pow__` and `__divide__` in the spec currently only require floating-point, meaning for the NumPy API, they error on `int` inputs. 

That does sound right, `x**2` should work right? Only negative integers will error. Which is not something easy to express in type annotation (which is fine).

> Does mypy have a way to specify float but not int?

Not that I know of.

asmeurer:
Actually, I didn't notice that the spec says "numeric data type" for `__pow__` and `__truediv__` but "floating-point data type" for `pow` and `divide`. So I presently had them not working for integer inputs, but I should change this. Or the spec should be fixed (it doesn't mention anything about negative integers for `__pow__`).

rgommers:
Ah that should be made consistent. And negative integers should be disallowed I believe.

PR #174, opened 2021-04-26
Some small fixes to function signatures to make them valid Python

asmeurer:


spec/API_specification/creation_functions.md
```diff
@@ -290,7 +290,7 @@ Returns evenly spaced numbers over a specified interval.
     -   a one-dimensional array containing evenly spaced values.
 
 (function-meshgrid)=
-### meshgrid(*arrays, /, *, indexing='xy')
+### meshgrid(*arrays, indexing='xy')
```
rgommers, line 293: I tried, that's indeed a syntax error otherwise - mildly surprised.

rgommers:
```python
>>> x = y = np.arange(2)
>>> np.meshgrid(x, y, 'ij')
[array([[[0],
        [1]],

       [[0],
        [1]]]), array([[[0],
        [0]],

       [[1],
        [1]]]), array([[['ij'],
        ['ij']],

       [['ij'],
        ['ij']]], dtype='<U2')]
```

`:facepalm:`

rgommers (approved):
Thanks Aaron!

spec/API_specification/creation_functions.md
asmeurer: The way I remember is that `*args` and `*` are both basically the same thing. They both separate keyword only arguments. The `*` is just a version of `*args` that doesn't actually allow any additional positional arguments.

PR #179, opened 2021-05-12
Fix `matmul` formatting

leofang:
Fixing this return list: 
截圖 2021-05-12 上午1 38 53

PR #181, opened 2021-05-12
Fix `norm` formatting

leofang:
See https://github.com/data-apis/array-api/pull/179#issuecomment-839462530. The fix can be visually verified by inspecting the GH Action's artifact.


kgryte (approved):
Thanks, @leofang! Bit odd that the empty lines were the issue.

leofang:
Yeah looks like after a few markdown levels it stops interpreting empty lines correctly...

Issue #165, opened 2021-04-15
Linear algebra namespace overview


kgryte:
In NumPy (and other array libraries), linear algebra APIs can be found in both the top-level namespace and a separate linear algebra (`linalg`) sub-namespace.

The purpose of this issue is to provide an overview of how array libraries (notably NumPy and TensorFlow) currently delineate linear algebra APIs and to provide a place for discussing whether the specification should adopt a similar division and what that division should be.

As discussed in [gh-147](https://github.com/data-apis/array-api/issues/147) and in meetings, whether the specification should adopt a similar division is a matter of some debate.

Notably, some array libraries, such as Dask, may have difficulty implementing some linear algebra APIs, and, thus, the question is raised as to whether we should include "essential", relatively straightforward linear algebra APIs in the top-level namespace (i.e., those APIs whose implementations do not require factorization (of the BLAS variety, rather than of the LAPACK variety)) and place the remaining APIs in a sub-namespace which could be an optional extension.

Some individuals have proposed that we include all linear algebra APIs (mirroring NumPy) in a sub-namespace `linalg`, but also mirror a select subset of APIs in the top-level namespace (e.g., `matmul` and `linalg.matmul`). The benefit of this is that (1) linear algebra APIs in the top-level namespace would be guaranteed and provide a minimal subset of linear algebra APIs in all standard-conforming array libraries and (2) for consistency, a user could simply request the `linalg` sub-namespace and get all linear algebra APIs (something which is not currently possible in, e.g., NumPy today).

What follows is an overview of NumPy's linear algebra API division which is adopted by several other array libraries (e.g., Dask, CuPy, JAX, etc).

### Top-Level Namespace

-   cross
-   diagonal
-   einsum
-   matmul
-   outer
-   tensordot
-   trace
-   transpose
-   vecdot

**Note**: essentially sum-products and basic array manipulation (transpose, diagonal).

### Linalg Namespace

-   cholesky
-   det
-   eig
-   eigh
-   eigvals
-   eigvalsh
-   inv
-   lstsq
-   matrix_power
-   matrix_rank
-   norm
-   pinv
-   qr
-   slogdet
-   solve
-   svd
-   svdvals

## TensorFlow

TensorFlow adopts a similar division, but places the following APIs in a `linalg` sub-namespace, rather than than the top-level namespace...

### Top-Level

-   einsum
-   tensordot
-   transpose

### Linalg Namespace

-   ...
-   cross
-   diag (diagonal)
-   matmul
[...]

PR #182, opened 2021-05-16
Move linear algebra APIs to an extension

kgryte:
This PR

-   builds on [gh-165](https://github.com/data-apis/array-api/issues/165) whereby only a small subset of linear algebra APIs are present in the top-level namespace and the rest are specified in a dedicated `linalg` sub-namespace as an extension to the array API standard.

## Notes

-   How extensions should be authored and consumed remains to be decided. This PR should lay the groundwork for further specification.


asmeurer:
Did we decide to mirror the functions in the top-level namespace?

rgommers:
> Did we decide to mirror the functions in the top-level namespace?

I believe that was indeed the preferred option. Discussion at https://github.com/data-apis/array-api/pull/149#pullrequestreview-621441318 followed by a discussion in the May 6th call where people seemed happy with @kgryte's proposal as implemented in this PR.

For others, it may help to capture how the PR in its current state is rendered. The API specification for the main namespace has these functions:

image

Extensions is a separate section:

image

And that section has all linear algebra functions in it (including the 5 in the main namespace, which are aliased).

rgommers (approved):
LGTM, thanks @kgryte

PR #186, opened 2021-05-25
Add a missing requirement for the producer of DLPack capsules

leofang:
See https://github.com/dmlc/dlpack/issues/34#issuecomment-847669399. Manipulations of a  PyCapsule rely on the precise knowledge of the capsule name (`PyCapsule_GetPointer`, `PyCapsule_GetName`, `PyCapsule_IsValid`, etc), so we cannot allow any ambiguity.


rgommers (approved):
Thanks Leo!

PR #201, opened 2021-06-14
Update output dtypes for bitwise shift operations

kgryte:
This PR

-   updates the output dtypes for the bitwise shift operations to abide by type promotion. Current guidance requires that the output array have the same dtype as the first array argument. This PR changes that guidance to requiring type promotion and thus ensures consistency with other binary (i.e., 2-ary) elementwise functions.


rgommers (approved):
For context, see https://github.com/pytorch/pytorch/issues/59867. This was simply an inconsistency that was not noticed in the initial PR that added the `bitwise_*` functions. So I'll merge this as a bug fix. Thanks @kgryte

PR #203, opened 2021-06-21
Update guidance on tuple field names

kgryte:
This PR

-   updates the guidance on tuple field names for named tuples returned from linear algebra functions by making explicit field name requirements.

cc @asmeurer


asmeurer (approved):

rgommers (approved):
LGTM too, thanks @kgryte and @asmeurer

PR #204, opened 2021-06-21
Clarify `eig` and `eigvals` todos

kgryte:
This PR

-   updates the TODOs to indicate that the function `eig` and `eigvals` are dependent on complex number support being added to the specification.


oleksandr-pavlyk (approved):

rgommers (approved):
Thanks @kgryte and @oleksandr-pavlyk

PR #205, opened 2021-06-21
Require unique counts array to be default integer dtype

kgryte:
This PR

-   requires that the array containing unique counts be the default integer dtype. This change is consistent with guidance included elsewhere in the specification, such as for index arrays.


rgommers (approved):
This looks like an obvious fix, thanks @kgryte

PR #206, opened 2021-06-21
Update formatting and organization of dtype guidance

kgryte:
This PR

-   updates the copy for data type guidance, reorganizing content and fixing grammar and punctuation. Guidance and specification requirements remains the same.


rgommers (approved):
Looks like an improvement, thanks @kgryte

Issue #209, opened 2021-06-29
[Consistency] Definition of the properties of the output


Lezcano:
Some functions such as [`linalg.qr`](https://data-apis.org/array-api/latest/extensions/linear_algebra_functions.html?highlight=ndim#linalg-qr-x-mode-reduced) define the properties that the outputs should have---`Q` orthogonal, `R` upper-triangular---but others such as [`linalg.eigh`](https://data-apis.org/array-api/latest/extensions/linear_algebra_functions.html?linalg-eigh-x-upper-false#linalg-eigh-x-upper-false) do not.

I believe that it would be best to specify the properties that these matrices should have in the definition of each function.

cc @rgommers @mruberry

rgommers:
Makes sense to me to improve the descriptions here, thanks @Lezcano .

PR #220, opened 2021-07-09
Make the writing of the factorisations in linalg more uniform

Lezcano:
Fixes https://github.com/data-apis/array-api/issues/209

I also took the liberty to make the descriptions of the factorisations more uniform.
Happy to change any of them if the reviewers believe that a different option is better.

- Added explicit formulae for the factorisations in the main description

- Added properties (triangular, orthogonal, etc) of the returned matrices in the main description and
the return section

- Uniform language: Prefer "Returns" over "Computes" at the start of every
description.

- Uniform language: Prefer "decomposition" over "factorization"

cc @rgommers @asmeurer

### DIFF
--- a/spec/extensions/linear_algebra_functions.md
+++ b/spec/extensions/linear_algebra_functions.md
@@ -96,7 +96,7 @@ Returns the lower (upper) Cholesky decomposition x = LLᵀ (x = UᵀU) of a symm
 
 -   **out**: _&lt;array&gt;_
 
-    -   an array containing the Cholesky factors for each square matrix. If `upper` is `False`, the array must contain lower-triangular matrices. Otherwise, it must contain upper-triangular matrices. The returned array must have a floating-point data type determined by {ref}`type-promotion` and shape as `x`.
+    -   an array containing the Cholesky factors for each square matrix. If `upper` is `False`, the returned array must contain lower-triangular matrices; otherwise, the returned array must contain upper-triangular matrices. The returned array must have a floating-point data type determined by {ref}`type-promotion` and must have the same shape as `x`.
 
 (function-linalg-cross)=
 ### linalg.cross(x1, x2, /, *, axis=-1)
--- a/spec/extensions/linear_algebra_functions.md
+++ b/spec/extensions/linear_algebra_functions.md
@@ -77,7 +77,7 @@ Accordingly, the standardization process affords the opportunity to reduce inter
 (function-linalg-cholesky)=
 ### linalg.cholesky(x, /, *, upper=False)
 
-Returns the Cholesky decomposition of a symmetric positive-definite matrix (or a stack of symmetric positive-definite matrices) `x`.
+Returns the lower (resp. upper) Cholesky decomposition `x = LLᵀ` (resp. `x = UᵀU`) of a symmetric positive-definite matrix (or a stack of matrices) `x`, where `L` is a lower-triangular matrix or a stack of matrices (resp. `U` is an upper-triangular matrix or a stack of matrices).
 
 <!-- NOTE: once complex numbers are supported, each square matrix must be Hermitian. -->
 
@@ -89,13 +89,13 @@ Returns the Cholesky decomposition of a symmetric positive-definite matrix (or a
 
 -   **upper**: _bool_
 
-    -   If `True`, the result must be the upper-triangular Cholesky factor. If `False`, the result must be the lower-triangular Cholesky factor. Default: `False`.
+    -   If `True`, the result must be the upper-triangular Cholesky factor `U`. If `False`, the result must be the lower-triangular Cholesky factor `L`. Default: `False`.
 
 #### Returns
 
 -   **out**: _&lt;array&gt;_
 
-    -   an array containing the Cholesky factors for each square matrix. The returned array must have a floating-point data type determined by {ref}`type-promotion` and shape as `x`.
+    -   an array containing the Cholesky factors for each square matrix. If `upper` is `False`, the array must contain lower-triangular matrices. Otherwise, it must contain upper-triangular matrices. The returned array must have a floating-point data type determined by {ref}`type-promotion` and shape as `x`.
 
 (function-linalg-cross)=
 ### linalg.cross(x1, x2, /, *, axis=-1)
@@ -125,7 +125,7 @@ Returns the cross product of 3-element vectors. If `x1` and `x2` are multi-dimen
 (function-linalg-det)=
 ### linalg.det(x, /)
 
-Returns the determinant of a square matrix (or stack of square matrices) `x`.
+Returns the determinant of a square matrix (or a stack of matrices) `x`.
 
 #### Parameters
 
@@ -182,9 +182,10 @@ _TODO: this requires complex number support to be added to the specification._
 (function-linalg-eigh)=
 ### linalg.eigh(x, /, *, upper=False)
 
-Returns the eigenvalues and eigenvectors of a symmetric matrix (or a stack of symmetric matrices) `x`.
+Returns the eigenvalues and eigenvectors `x = QLQᵀ` of a symmetric matrix (or a stack of matrices) `x`, where `Q` is an orthogonal matrix (or a stack of matrices) and `L` is a vector (or a stack of vectors).
 
-<!-- NOTE: once complex number support, each matrix must be Hermitian -->
+<!-- NOTE: once complex number support, each matrix must be Hermitian and the returned Q unitary.
+           We might also want to make the dtype of `eigenvalues` unconditionally real -->
 
 #### Parameters
 
@@ -201,9 +202,9 @@ Returns the eigenvalues and eigenvectors of a symmetric matrix (or a stack of sy
 -   **out**: _Tuple\[ &lt;array&gt; ]_
 
     -   a namedtuple (`eigenvalues`, `eigenvectors`) whose
-    
-        -   first element must have the field name `eigenvalues` and must be an array consisting of computed eigenvalues. The array containing the eigenvalues must have shape `(..., M)`.
-        -   second element have have the field name `eigenvectors` and must be an array where the columns of the inner most matrices contain the computed eigenvectors. The array containing the eigenvectors must have shape `(..., M, M)`.
+
+        -   first element must have the field name `eigenvalues` (corresponding to `L` above) and must be an array consisting of computed eigenvalues. The array containing the eigenvalues must have shape `(..., M)`.
+        -   second element have have the field name `eigenvectors` (corresponding to `Q` above) and must be an array where the columns of the inner most matrices contain the computed eigenvectors. These matrices must be orthogonal. The array containing the eigenvectors must have shape `(..., M, M)`.
 
         Each returned array must have the same floating-point data type as `x`.
 
@@ -220,7 +221,7 @@ _TODO: this requires complex number support to be added to the specification._
 (function-linalg-eigvalsh)=
 ### linalg.eigvalsh(x, /, *, upper=False)
 
-Computes the eigenvalues of a symmetric matrix (or a stack of symmetric matrices) `x`.
+Returns the eigenvalues of a symmetric matrix (or a stack of matrices) `x`.
 
 <!-- NOTE: once complex number support, each matrix must be Hermitian -->
 
@@ -253,7 +254,7 @@ Alias for {ref}`function-einsum`.
 (function-linalg-inv)=
 ### linalg.inv(x, /)
 
-Computes the multiplicative inverse of a square matrix (or a stack of square matrices) `x`.
+Returns the multiplicative inverse of a square matrix (or a stack of matrices) `x`.
 
 #### Parameters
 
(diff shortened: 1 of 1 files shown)


spec/extensions/linear_algebra_functions.md
```diff
@@ -77,7 +77,7 @@ Accordingly, the standardization process affords the opportunity to reduce inter
 (function-linalg-cholesky)=
 ### linalg.cholesky(x, /, *, upper=False)
 
-Returns the Cholesky decomposition of a symmetric positive-definite matrix (or a stack of symmetric positive-definite matrices) `x`.
+Returns the lower (resp. upper) Cholesky decomposition `x = LLᵀ` (resp. `x = UᵀU`) of a symmetric positive-definite matrix (or a stack of matrices) `x`, where `L` is a lower-triangular matrix or a stack of matrices (resp. `U` is an upper-triangular matrix or a stack of matrices).
```
asmeurer, line 80: If you're going to use Unicode symbols I think it would be better to avoid wrapping it in a code block. At least for me, whatever font this uses in the rendered version doesn't look very good. And anyway, backticks should only be used for things that are actual code.

spec/extensions/linear_algebra_functions.md
```diff
@@ -125,7 +125,7 @@ Returns the cross product of 3-element vectors. If `x1` and `x2` are multi-dimen
 (function-linalg-det)=
 ### linalg.det(x, /)
 
-Returns the determinant of a square matrix (or stack of square matrices) `x`.
+Returns the determinant of a square matrix (or a stack of matrices) `x`.
```
asmeurer, line 128: IMO it's better to not remove the qualifiers from "stack of matrices" (except for the change in `pinv` where it was wrong).

spec/extensions/linear_algebra_functions.md
Lezcano: I completely agree. I did not know how to handle this one.

It is true that backticks are often used for, code, but I read that the expression `x = USV` was in the SVD docs. If we wanted to go for actual code, that'd have to be te much more convoluted `x = U.matmul(S.matmul(V))`.

Should I just remove the backticks in all these expressions, or what would be the preferred solution?

spec/extensions/linear_algebra_functions.md
Lezcano: The only problem with that is that sometimes it can be a bit overwhelming (think of the case "symmetric positive definite"). Should we still go for qualifiers regardless?

spec/extensions/linear_algebra_functions.md
kgryte: Instead of code blocks, you could make equations italic. In general, would be nice to have equation rendering capabilities in the spec.

Lezcano: The problem with making it italics is that `x`, `L` and `U` are represented later in code, so it'd be a bit odd to have some parts of the text in italics and some others in code.

Having LaTeX support would certainly be helpful. The way we did it in PyTorch is to first write the equations in LaTeX and then saying "the outputs map to the equations above as follows". See for example the return section of [torch.linalg.eig](https://pytorch.org/docs/stable/generated/torch.linalg.eig.html)

kgryte:
```suggestion
Returns the lower (upper) Cholesky decomposition x = LLᵀ (x = UᵀU) of a symmetric positive-definite matrix (or a stack of matrices) `x`, where `L` is a lower-triangular matrix or a stack of matrices (`U` is an upper-triangular matrix or a stack of matrices).
```

spec/extensions/linear_algebra_functions.md
```diff
@@ -170,9 +170,10 @@ Returns the specified diagonals of a matrix (or a stack of matrices) `x`.
 (function-linalg-eigh)=
 ### linalg.eigh(x, /)
 
-Returns the eigenvalues and eigenvectors of a symmetric matrix (or a stack of symmetric matrices) `x`.
+Returns the eigenvalues and eigenvectors `x = QLQᵀ` of a symmetric matrix (or a stack of matrices) `x`, where `Q` is an orthogonal matrix (or a stack of matrices) and `L` is a vector (or a stack of vectors).
```
kgryte, line 173:
```suggestion
Returns the eigenvalues and eigenvectors x = QLQᵀ of a symmetric matrix (or a stack of matrices) `x`, where `Q` is an orthogonal matrix (or a stack of matrices) and `L` is a vector (or a stack of vectors).
```

spec/extensions/linear_algebra_functions.md
```diff
@@ -395,7 +396,7 @@ Computes the (Moore-Penrose) pseudo-inverse of a matrix (or a stack of square ma
 (function-linalg-qr)=
 ### linalg.qr(x, /, *, mode='reduced')
 
-Computes the qr factorization of a matrix (or a stack of matrices), where `q` is an orthonormal matrix (or a stack of matrices) and `r` is an upper-triangular matrix (or a stack of matrices).
+Returns the qr decomposition `x = qr` of a matrix (or a stack of matrices) `x`, where `q` is an orthonormal matrix (or a stack of matrices) and `r` is an upper-triangular matrix (or a stack of matrices).
```
kgryte, line 399:
```suggestion
Returns the qr decomposition x = QR of a matrix (or a stack of matrices) `x`, where `Q` is an orthonormal matrix (or a stack of matrices) and `R` is an upper-triangular matrix (or a stack of matrices).
```

spec/extensions/linear_algebra_functions.md
```diff
@@ -476,7 +477,7 @@ Returns the solution to the system of linear equations represented by the well-d
 (function-linalg-svd)=
 ### linalg.svd(x, /, *, full_matrices=True)
 
-Computes the singular value decomposition `A = USVh` of a matrix (or a stack of matrices) `x`.
+Returns the singular value decomposition `A = USVh` of a matrix (or a stack of matrices) `x` where `U` is a matrix (or a stack of matrices) with orthonormal columns, `S` is a vector of non-negative numbers (or stack of vectors), and `Vh` is a matrix (or a stack of matrices) with orthonormal rows.
```
kgryte, line 480:
```suggestion
Returns the singular value decomposition A = USVh of a matrix (or a stack of matrices) `x` where `U` is a matrix (or a stack of matrices) with orthonormal columns, `S` is a vector of non-negative numbers (or stack of vectors), and `Vh` is a matrix (or a stack of matrices) with orthonormal rows.
```

spec/extensions/linear_algebra_functions.md
kgryte:
```suggestion
Returns the determinant of a square matrix (or a stack of square matrices) `x`.
```

spec/extensions/linear_algebra_functions.md
```diff
@@ -90,13 +90,13 @@ Returns the Cholesky decomposition of a symmetric positive-definite matrix (or a
...
 
-    -   If `True`, the result must be the upper-triangular Cholesky factor. If `False`, the result must be the lower-triangular Cholesky factor. Default: `False`.
+    -   If `True`, the result must be the upper-triangular Cholesky factor `U`. If `False`, the result must be the lower-triangular Cholesky factor `L`. Default: `False`.
 
 #### Returns
 
 -   **out**: _&lt;array&gt;_
 
-    -   an array containing the Cholesky factors for each square matrix. The returned array must have a floating-point data type determined by {ref}`type-promotion` and shape as `x`.
+    -   an array containing the Cholesky factors for each square matrix. If `upper` is `False`, the array must contain lower-triangular matrices. Otherwise, it must contain upper-triangular matrices. The returned array must have a floating-point data type determined by {ref}`type-promotion` and shape as `x`.
```
kgryte, line 99:
```suggestion
    -   an array containing the Cholesky factors for each square matrix. If `upper` is `False`, the returned array must contain lower-triangular matrices; otherwise, the returned array must contain upper-triangular matrices. The returned array must have a floating-point data type determined by {ref}`type-promotion` and must have the same shape as `x`.
```

spec/extensions/linear_algebra_functions.md
```diff
@@ -170,9 +170,10 @@ Returns the specified diagonals of a matrix (or a stack of matrices) `x`.
 (function-linalg-eigh)=
 ### linalg.eigh(x, /)
 
-Returns the eigenvalues and eigenvectors of a symmetric matrix (or a stack of symmetric matrices) `x`.
+Returns the eigenvalues and eigenvectors x = QLQᵀ of a symmetric matrix (or a stack of matrices) `x`, where `Q` is an orthogonal matrix (or a stack of matrices) and `L` is a vector (or a stack of vectors).
 
-<!-- NOTE: once complex number support, each matrix must be Hermitian -->
+<!-- NOTE: once complex number support, each matrix must be Hermitian and the returned Q unitary.
+           We might also want to make the dtype of `eigenvalues` unconditionally real -->
```
kgryte, line 176:
```suggestion
<!-- NOTE: once complex number support, each matrix must be Hermitian and the returned Q unitary. We might also want to make the dtype of `eigenvalues` unconditionally real -->
```

spec/extensions/linear_algebra_functions.md
```diff
@@ -233,7 +234,7 @@ as it requires complex number support.
 (function-linalg-inv)=
 ### linalg.inv(x, /)
 
-Computes the multiplicative inverse of a square matrix (or a stack of square matrices) `x`.
+Returns the multiplicative inverse of a square matrix (or a stack of matrices) `x`.
```
kgryte, line 236:
```suggestion
Returns the multiplicative inverse of a square matrix (or a stack of square matrices) `x`.
```

spec/extensions/linear_algebra_functions.md
```diff
@@ -204,7 +205,7 @@ as it requires complex number support.
 (function-linalg-eigvalsh)=
 ### linalg.eigvalsh(x, /)
 
-Computes the eigenvalues of a symmetric matrix (or a stack of symmetric matrices) `x`.
+Returns the eigenvalues of a symmetric matrix (or a stack of matrices) `x`.
```
kgryte, line 207:
```suggestion
Returns the eigenvalues of a symmetric matrix (or a stack of symmetric matrices) `x`.
```

spec/extensions/linear_algebra_functions.md
```diff
@@ -170,9 +170,10 @@ Returns the specified diagonals of a matrix (or a stack of matrices) `x`.
 (function-linalg-eigh)=
 ### linalg.eigh(x, /)
 
-Returns the eigenvalues and eigenvectors of a symmetric matrix (or a stack of symmetric matrices) `x`.
+Returns the eigenvalues and eigenvectors x = QLQᵀ of a symmetric matrix (or a stack of matrices) `x`, where `Q` is an orthogonal matrix (or a stack of matrices) and `L` is a vector (or a stack of vectors).
```
kgryte, line 173:
```suggestion
Returns the eigenvalues and eigenvectors x = QLQᵀ of a symmetric matrix (or a stack of symmetric matrices) `x`, where `Q` is an orthogonal matrix (or a stack of matrices) and `L` is a vector (or a stack of vectors).
```

spec/extensions/linear_algebra_functions.md
```diff
@@ -306,7 +307,7 @@ Computes the matrix norm of a matrix (or a stack of matrices) `x`.
 (function-linalg-matrix_power)=
 ### linalg.matrix_power(x, n, /)
 
-Raises a square matrix (or a stack of square matrices) `x` to an integer power `n`.
+Raises a square matrix (or a stack of matrices) `x` to an integer power `n`.
```
kgryte, line 309:
```suggestion
Raises a square matrix (or a stack of square matrices) `x` to an integer power `n`.
```

spec/extensions/linear_algebra_functions.md
```diff
@@ -418,15 +419,15 @@ Computes the qr factorization of a matrix (or a stack of matrices), where `q` is
 
     -   a namedtuple `(q, r)` whose
 
-        -   first element must have the field name `q` and must be an array whose shape depends on the value of `mode` and contain orthonormal matrices. If `mode` is `'complete'`, the array must have shape `(..., M, M)`. If `mode` is `'reduced'`, the array must have shape `(..., M, K)`, where `K = min(M, N)`. The first `x.ndim-2` dimensions must have the same size as those of the input `x`.
+        -   first element must have the field name `q` and must be an array whose shape depends on the value of `mode` and contain matrices with orthonormal columns. If `mode` is `'complete'`, the array must have shape `(..., M, M)`. If `mode` is `'reduced'`, the array must have shape `(..., M, K)`, where `K = min(M, N)`. The first `x.ndim-2` dimensions must have the same size as those of the input `x`.
```
kgryte, line 421:
```suggestion
        -   first element must have the field name `q` and must be an array whose shape depends on the value of `mode` and contain matrices with orthonormal columns. If `mode` is `'complete'`, the array must have shape `(..., M, M)`. If `mode` is `'reduced'`, the array must have shape `(..., M, K)`, where `K = min(M, N)`. The first `x.ndim-2` dimensions must have the same size as those of the input array `x`.
```

spec/extensions/linear_algebra_functions.md
```diff
@@ -418,15 +419,15 @@ Computes the qr factorization of a matrix (or a stack of matrices), where `q` is
...
         -   second element must have the field name `r` and must be an array whose shape depends on the value of `mode` and contain upper-triangular matrices. If `mode` is `'complete'`, the array must have shape `(..., M, M)`. If `mode` is `'reduced'`, the array must have shape `(..., K, N)`, where `K = min(M, N)`. The first `x.ndim-2` dimensions must have the same size as those of the input `x`.
 
         Each returned array must have a floating-point data type determined by {ref}`type-promotion`.
 
 (function-linalg-slogdet)=
 ### linalg.slogdet(x, /)
 
-Returns the sign and the natural logarithm of the absolute value of the determinant of a square matrix (or a stack of square matrices) `x`.
+Returns the sign and the natural logarithm of the absolute value of the determinant of a square matrix (or a stack of matrices) `x`.
```
kgryte, line 429:
```suggestion
Returns the sign and the natural logarithm of the absolute value of the determinant of a square matrix (or a stack of square matrices) `x`.
```

kgryte (approved):
LGTM.

PR #221, opened 2021-07-09
Require __pow__ and __truediv__ to have floating-point inputs

asmeurer:
Previously it was any numeric dtype, but this was inconsistent with the
functions `pow()` and `divide()`. Integer inputs are problematic because they
would either always have to be cast to float or use value-based casting
depending on, e.g., the sign of the exponent for `__pow__`. Furthermore, in
order to properly correspond to Python conventions, `__truediv__` should always
return a float (integer division should be implemented by `__floordiv__`).

This was discussed in #173.

### DIFF
--- a/spec/API_specification/array_object.md
+++ b/spec/API_specification/array_object.md
@@ -1018,7 +1018,7 @@ For floating-point operands, let `self` equal `x1` and `other` equal `x2`.
 
     -   array instance whose elements correspond to the exponentiation base. Should have a floating-point data type.
 
--   **other**: _Union\[ int, float, &lt;array&gt; ]_
+-   **other**: _Union\[ float, &lt;array&gt; ]_
 
     -   other array whose elements correspond to the exponentiation exponent. Must be compatible with `self` (see {ref}`broadcasting`). Should have a floating-point data type.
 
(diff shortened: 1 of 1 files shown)


rgommers (changes requested):
This does not look right to me, and not in line with the discussion in gh-173. It should be numeric dtype (so the `pow` and `divide` functions need updating instead), and it should say that `other` must not contain negative integers (or "behavior is undefined for negative integers"). This seems right:
```python
In [1]: x = np.arange(5) - 1

In [2]: x**0
Out[2]: array([1, 1, 1, 1, 1])

In [3]: x**2
Out[3]: array([1, 0, 1, 4, 9])

In [4]: x**-1
ValueError                                Traceback (most recent call last)
<ipython-input-5-13f1892b991f> in <module>
----> 1 x**-1

ValueError: Integers to negative integer powers are not allowed.
```

> Integer inputs are problematic because they would either always have to be cast to float or use value-based casting depending on, e.g., the sign of the exponent for `__pow__`.

Raising an exception is neither of those.

asmeurer:
That is still a value-based check. Although I guess it's similar to `int//int` working except when the denominator is 0.  

For `__trudiv__` there's still the argument that general Python 3 semantics say / is floating-point division and // is integer division. We could make int/int always cast to float, but that's different from other casting rules.

kgryte:
This PR was superseded by #330. Closing...

Issue #217, opened 2021-07-01
Proposal: Remove `upper` argument from `eigh` and `eigvalsh`


Lezcano:
Some functions that accept a symmetric matrix (`eigh`, `eighvals`) have an `upper=False` kwarg. According to the API: 
“If `True`, use the upper-triangular part to compute the eigenvalues and eigenvectors. If `False`, use the lower-triangular part to compute the eigenvalues and eigenvectors.”.

There are a number of reasons why we might want to remove this argument from the API.
1. This kwarg goes against [Point 4 of the design principles](https://data-apis.org/array-api/latest/extensions/linear_algebra_functions.html?linalg-eigh-x-upper-false#design-principles), as it biases the implementations towards “always assuming that the input is well-formed” (symmetric in this case) when it could very much not be the case. This prevents the implementations from throwing meaningful errors when the input is not symmetric.
2. Lack of orthogonality: The API does not offer a parameter of this kind for other functions that also accept a symmetric matrix as inputs such as `cholesky`. In fact, `cholesky` has a boolean parameter also named `upper` which does something completely different to `eigh`'s `upper`.
3. Makes the semantics of autograd counterintuitive: This is not that relevant for this API as this API does not specify differentiation behaviour, but this API will potentially be followed by a number of frameworks that implement differentiation.
It is not clear whether one should assume that the input is symmetric, or that the upper (resp. lower) triangular part of the input matrix input matrix represents a symmetric matrix via the transformation `A.triu() + A.triu(1).T` . This is problematic when computing the gradient, as it is not clear whether the gradient should be symmetric or upper triangular. PyTorch makes the gradient symmetric to be mathematically consistent with the semantics of the operation, rather than with the optimisation that this kwarg suggests, which is semantically wrong but more intuitive. We have not received any issue about this, which hints to the fact that almost no one uses this optimisation to pack two symmetric matrices into one unconstrained matrix (and an extra vector).
4. Lack of Orthogonality 2. It provides an optimisation that can be accomplished via the current API. If a user wants to pack two symmetric matrices with zero diagonal as the upper and lower-triangular part of a square matrix they could write `L1, Q1 = linalg.eigh(A.triu(1) + A.triu(1).T); L2, Q2 = linalg.eigh(A.tril(-1) + A.tril(-1).T)`. Doing this and implementing symmetric gradients (which is what is mathematically correct for `eigh`) would give the correct semantics with respect to the upper (resp. lower) part of `A`.
[...]

PR #224, opened 2021-07-14
Account for when factorizations and solvers are not well-defined

### LINKED ISSUE #207: Some linalg functions are not well-defined
The functions `eigh`, `svd`, `qr` and `lstsq` are not well-defined as written.

## Uniqueness `lstsq`
`linalg.lstsq` docs currently read: 
“Returns the least-squares solution to a linear matrix equation”
but the solution needn't be unique whenever `A` has non-trivial kernel (e.g. `A = 0`).

**Proposed solution**:
- Rewrite it as "Returns a least-square solution [...]"
- Perhaps we want to add a note on how to handle these cases (e.g. the solution with smallest norm is chosen) so that this function is well-defined for every input matrix. Note that this would not allow for the implementation of this function via faster but less robust algorithms such as LAPACK's `gels`, which assumes that `A` is full-rank and performs a QR decomposition.

## Uniqueness `eigh` and `svd`:
- Eigenvectors (resp. singular vectors) of a generic matrix are only defined up to sign in the real case and multiplication by a scalar of norm 1 in the complex case
- When the matrix has repeated eigenvalues (resp. singular values) the associated eigenvectors are just defined up to multiplication by a rotation (or multiplication by a unitary matrix in the complex case).

**Proposed solution**:
- Change the wording from "Returns the eigenvalues and eigenvectors of a symmetric matrix" to "Returns an eigendecomposition of a symmetric matrix".
- Change the wording of `linalg.svd` as well.

## QR decomposition cannot be defined 
The `qr` decomposition is just well-defined whenever the matrix has full column rank. In the singular case pivoting is necessary.

**Proposed solution**: The behaviour of this function should be marked as unspecified whenever the input matrix does not have full column rank.

cc @rgommers @mruberry

Lezcano:
Fixes #207

cc @asmeurer @rgommers @kgryte

### DIFF
--- a/spec/extensions/linear_algebra_functions.md
+++ b/spec/extensions/linear_algebra_functions.md
@@ -182,7 +182,7 @@ _TODO: this requires complex number support to be added to the specification._
 (function-linalg-eigh)=
 ### linalg.eigh(x, /, *, upper=False)
 
-Returns the eigenvalues and eigenvectors of a symmetric matrix (or a stack of symmetric matrices) `x`.
+Returns an eigendecomposition of a symmetric matrix (or a stack of symmetric matrices) `x`.
 
 <!-- NOTE: once complex number support, each matrix must be Hermitian -->
 
@@ -270,7 +270,7 @@ Computes the multiplicative inverse of a square matrix (or a stack of square mat
 (function-linalg-lstsq)=
 ### linalg.lstsq(x1, x2, /, *, rtol=None)
 
-Returns the least-squares solution to a linear matrix equation `Ax = b`.
+Returns the minimum-norm least-squares solution to a linear matrix equation `Ax = b`.
 
 #### Parameters
 
@@ -470,13 +470,13 @@ Computes the (Moore-Penrose) pseudo-inverse of a matrix (or a stack of square ma
 (function-linalg-qr)=
 ### linalg.qr(x, /, *, mode='reduced')
 
-Computes the qr factorization of a matrix (or a stack of matrices), where `q` is an orthonormal matrix (or a stack of matrices) and `r` is an upper-triangular matrix (or a stack of matrices).
+Computes the qr factorization of a full column rank matrix (or a stack of matrices), where `q` is an orthonormal matrix (or a stack of matrices) and `r` is an upper-triangular matrix (or a stack of matrices).
 
 #### Parameters
 
 -   **x**: _&lt;array&gt;_
 
-    -   input array having shape `(..., M, N)` and whose innermost two dimensions form `MxN` matrices. Should have a floating-point data type.
+    -   input array having shape `(..., M, N)` and whose innermost two dimensions form `MxN` matrices of rank equal to `N`. Should have a floating-point data type.
 
 -   **mode**: _str_
 
@@ -551,7 +551,7 @@ Returns the solution to the system of linear equations represented by the well-d
 (function-linalg-svd)=
 ### linalg.svd(x, /, *, full_matrices=True)
 
-Computes the singular value decomposition `A = USV` of a matrix (or a stack of matrices) `x`.
+Computes a singular value decomposition `A = USV` of a matrix (or a stack of matrices) `x`.
 
 #### Parameters
 


kgryte (approved):
LGTM

rgommers (approved):

spec/extensions/linear_algebra_functions.md
```diff
@@ -270,7 +270,7 @@ Computes the multiplicative inverse of a square matrix (or a stack of square mat
 (function-linalg-lstsq)=
 ### linalg.lstsq(x1, x2, /, *, rtol=None)
 
-Returns the least-squares solution to a linear matrix equation `Ax = b`.
+Returns the minimum-norm least-squares solution to a linear matrix equation `Ax = b`.
```
leofang, line 273: Two questions:
1. Following the same argument that led to the other changes in this PR, shouldn't we say "...*a* minimum-norm least-squares solution..." here? Is the solution always unique? 
2. English problem: Doesn't "least-squares" already refer to the L2 norm? So maybe we should just say "... a minimal least-squares solution..."?

spec/extensions/linear_algebra_functions.md
```diff
@@ -465,13 +465,13 @@ Computes the (Moore-Penrose) pseudo-inverse of a matrix (or a stack of square ma
 (function-linalg-qr)=
 ### linalg.qr(x, /, *, mode='reduced')
 
-Computes the qr factorization of a matrix (or a stack of matrices), where `q` is an orthonormal matrix (or a stack of matrices) and `r` is an upper-triangular matrix (or a stack of matrices).
+Computes the qr factorization of a full column rank matrix (or a stack of matrices), where `q` is an orthonormal matrix (or a stack of matrices) and `r` is an upper-triangular matrix (or a stack of matrices).
```
leofang, line 468: [Same problem as in removing "upper"](https://github.com/data-apis/array-api/issues/217#issuecomment-888029418): This "full column rank" addition seems to suggest an array library would do an expansive check before sending the input to the solver. I think in today's call we wanted to rephrase it for the other case such that this interpretation can be avoided (@kgryte will probably update us), so maybe we need to consider a similar phrasing here as well.

spec/extensions/linear_algebra_functions.md
```diff
@@ -465,13 +465,13 @@ Computes the (Moore-Penrose) pseudo-inverse of a matrix (or a stack of square ma
...
 
 #### Parameters
 
 -   **x**: _&lt;array&gt;_
 
-    -   input array having shape `(..., M, N)` and whose innermost two dimensions form `MxN` matrices. Should have a floating-point data type.
+    -   input array having shape `(..., M, N)` and whose innermost two dimensions form `MxN` matrices of rank equal to `N`. Should have a floating-point data type.
```
leofang, line 474: Ditto.

spec/extensions/linear_algebra_functions.md
Lezcano: The minimum norm solution is indeed always unique. It is given by `A^{\dagger}b`, where `A^{\dagger}` is the pseudoinverse of `A` (see the [wiki](https://en.wikipedia.org/wiki/Moore%E2%80%93Penrose_inverse#Applications)).

While least squares most of the times refers to the L2 norm, I think one still needs to explicit that is minimum with respect to a norm. I have never seen the expression "minimum vector".

spec/extensions/linear_algebra_functions.md
Lezcano: Note that that is a slippery slope. If you take that route you may also want to reword [linalg.solve](https://data-apis.org/array-api/latest/extensions/linear_algebra_functions.html?highlight=linalg%20solve#linalg-solve-x1-x2) which assumes that the input is full-rank. Same for [linalg.cholesky](https://data-apis.org/array-api/latest/extensions/linear_algebra_functions.html?highlight=linalg%20solve#linalg-cholesky-x-upper-false) which takes a symmetric positive definite matrix (SPD).

In my opinion, it is perfectly fine to left the array API unspecified for some inputs. That does not mean that the libraries will have to check that the input is well-formed explicitly. This becomes obvious when one tries to implements some of the checks. For example, a way to check whether a matrix is SPD is to perform Cholesky and see if it succeeds. In the same fashion, a way to see if a matrix is full rank is to perform a QR decomposition and see if it succeeds.

Furthermore, leaving unspecified does not mean that the libraries will have to check that the inputs are correct upfront. For example, in PyTorch we are looking into allowing rectangular inputs for `solve`, and these being dispatched directly into `lstsq`. If the rectangular input case is left unspecified, this is a conformant implementation of the API. On the other hand, if the API specifies that the input should have this shape (or these properties) **or throw an error** that would be much more problematic.

Finally, note leaving things unspecified is a common practice in other standards, such as that of C++, where not even the size of ints is specified.

kgryte: A path forward which can assuage both concerns is

1.  Accept the proposed wording change. This would be in-line with `linalg.cholesky` et al in which certain assumptions are stated explicitly.
2.  Add a note to the specification stating that whether an array library explicitly checks whether an input array is full column rank is implementation-defined. Meaning, an array library can choose whether to check or not before passing to solver.
3.  Add a note to the specification stating that how an array library handles input arrays which do not fulfill assumptions (e.g., full column rank) is implementation-defined.

spec/extensions/linear_algebra_functions.md
kgryte: This API has been removed from the spec. Will mark as resolved.

spec/extensions/linear_algebra_functions.md
kgryte: Will address in a follow-up PR.

spec/extensions/linear_algebra_functions.md
kgryte: Will address stated assumptions in a follow-up PR in accordance with my comment above.

kgryte (approved):
LGTM. Will submit a follow-up PR to address assumptions and whether an conforming library must validate those assumptions.

spec/extensions/linear_algebra_functions.md
kgryte: Addressed in https://github.com/data-apis/array-api/pull/302.

PR #225, opened 2021-07-14
Change to full_matrices=False by default

### LINKED ISSUE #214: [Consistency] Default behaviour of SVD vs QR
**Problem**
The current default behaviour of `linalg.svd` computes the full decomposition by setting `full_matrices=True` by default. On the other hand, the default behaviour of `linalg.qr` specifies `mode="reduced"` by default.

**Proposal**
Change de default to `full_matrices=False` in `linalg.svd`. In most cases the reduced SVD is good enough. Even more, it’s better to do less work by default and just make the more expensive operation opt-in.

cc @rgommers @mruberry

Lezcano:
As per the title.

Fixes #214 

cc @asmeurer @rgommers @kgryte


kgryte:
See [gh-214](https://github.com/data-apis/array-api/issues/214#issuecomment-903983134) for discussion.

Given the lack of consensus whether this change is desired based on use cases, keeping the status quo of returning a full SVD decomposition was preferred (as documented in consortium meeting minutes from 2021-07-29). Accordingly, the changes in this PR won't be accepted and will be closed out.

PR #226, opened 2021-07-14
Remove upper keyword from `eigh` and `eigvalsh`.

### LINKED ISSUE #217: Proposal: Remove `upper` argument from `eigh` and `eigvalsh`

Lezcano:
As per the title

Fixes #217 

cc @rgommers @asmeurer @kgryte

### DIFF
--- a/spec/extensions/linear_algebra_functions.md
+++ b/spec/extensions/linear_algebra_functions.md
@@ -180,7 +180,7 @@ Returns the specified diagonals. If `x` has more than two dimensions, then the a
 _TODO: this requires complex number support to be added to the specification._
 
 (function-linalg-eigh)=
-### linalg.eigh(x)
+### linalg.eigh(x, /)
 
 Returns the eigenvalues and eigenvectors of a symmetric matrix (or a stack of symmetric matrices) `x`.
 
@@ -214,7 +214,7 @@ Eigenvalue sort order is left unspecified.
 _TODO: this requires complex number support to be added to the specification._
 
 (function-linalg-eigvalsh)=
-### linalg.eigvalsh(x)
+### linalg.eigvalsh(x, /)
 
 Computes the eigenvalues of a symmetric matrix (or a stack of symmetric matrices) `x`.
 
--- a/spec/extensions/linear_algebra_functions.md
+++ b/spec/extensions/linear_algebra_functions.md
@@ -180,7 +180,7 @@ Returns the specified diagonals. If `x` has more than two dimensions, then the a
 _TODO: this requires complex number support to be added to the specification._
 
 (function-linalg-eigh)=
-### linalg.eigh(x, /, *, upper=False)
+### linalg.eigh(x)
 
 Returns the eigenvalues and eigenvectors of a symmetric matrix (or a stack of symmetric matrices) `x`.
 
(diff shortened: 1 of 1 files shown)


spec/extensions/linear_algebra_functions.md
```diff
@@ -180,7 +180,7 @@ Returns the specified diagonals. If `x` has more than two dimensions, then the a
 _TODO: this requires complex number support to be added to the specification._
 
 (function-linalg-eigh)=
-### linalg.eigh(x, /, *, upper=False)
+### linalg.eigh(x)
```
kgryte, line 183:
```suggestion
### linalg.eigh(x, /)
```

spec/extensions/linear_algebra_functions.md
```diff
@@ -218,7 +214,7 @@ Eigenvalue sort order is left unspecified.
 _TODO: this requires complex number support to be added to the specification._
 
 (function-linalg-eigvalsh)=
-### linalg.eigvalsh(x, /, *, upper=False)
+### linalg.eigvalsh(x)
```
kgryte, line 217:
```suggestion
### linalg.eigvalsh(x, /)
```

kgryte (approved):
LGTM. Made minor edits to ensure positional-only arguments.

kgryte:
This PR can be merged once [gh-243](https://github.com/data-apis/array-api/pull/243) is merged, which adds `tril` and `triu` to the specification.

kgryte:
[gh-243](https://github.com/data-apis/array-api/pull/243) has now been merged. Merging...

PR #230, opened 2021-07-16
Some improvements to type hints

asmeurer:
And other small fixes. The type hint fixes come from https://github.com/numpy/numpy/pull/18585, except for the `Literal` change in qr which comes from me. 

One thing that I didn't address here is the type for the `stream` argument in `__dlpack__`. See this discussion https://github.com/numpy/numpy/pull/18585#discussion_r659026425.

### DIFF
--- a/spec/API_specification/creation_functions.md
+++ b/spec/API_specification/creation_functions.md
@@ -57,7 +57,7 @@ Convert the input to an array.
 
 #### Parameters
 
--   **obj**: _Union\[ float, NestedSequence\[ bool | int | float ], SupportsDLPack, SupportsBufferProtocol ]_
+-   **obj**: _Union\[ &lt;array&gt;, float, NestedSequence\[ bool | int | float ], SupportsDLPack, SupportsBufferProtocol ]_
 
     -   Object to be converted to an array. Can be a Python scalar, a (possibly nested) sequence of Python scalars, or an object supporting DLPack or the Python buffer protocol.
 
--- a/spec/API_specification/creation_functions.md
+++ b/spec/API_specification/creation_functions.md
@@ -57,7 +57,7 @@ Convert the input to an array.
 
 #### Parameters
 
--   **obj**: _Union\[ &lt;array&gt;, float, NestedSequence\[ bool | int | float ], SupportsDLPack, SupportsBufferProtocol ]_
+-   **obj**: _Union\[ &lt;array&gt;, bool, int, float, NestedSequence\[ bool | int | float ], SupportsDLPack, SupportsBufferProtocol ]_
 
     -   Object to be converted to an array. Can be a Python scalar, a (possibly nested) sequence of Python scalars, or an object supporting DLPack or the Python buffer protocol.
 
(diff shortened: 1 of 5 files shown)


kgryte (approved):
LGTM.

spec/API_specification/manipulation_functions.md
```diff
@@ -19,7 +19,7 @@ Joins a sequence of arrays along an existing axis.
 
 #### Parameters
 
--   **arrays**: _Tuple\[ &lt;array&gt;, ... ]_
+-   **arrays**: _Union\[Tuple\[ &lt;array&gt;, ... ], List\[ &lt;array&gt; ] ]_
```
kgryte, line 22: Would it make more sense to use `Sequence`, rather than a `Union`?

spec/API_specification/manipulation_functions.md
```diff
@@ -154,7 +154,7 @@ Joins a sequence of arrays along a new axis.
...
```
kgryte, line 157: Same comment as above regarding `Sequence`.

spec/API_specification/manipulation_functions.md
asmeurer: That was mentioned here https://github.com/numpy/numpy/pull/18585#discussion_r591754462. If we agree Sequence is better we can use that instead.

spec/API_specification/creation_functions.md
```diff
@@ -57,7 +57,7 @@ Convert the input to an array.
 
 #### Parameters
 
--   **obj**: _Union\[ float, NestedSequence\[ bool | int | float ], SupportsDLPack, SupportsBufferProtocol ]_
+-   **obj**: _Union\[ &lt;array&gt;, float, NestedSequence\[ bool | int | float ], SupportsDLPack, SupportsBufferProtocol ]_
```
honno, line 60: I can raise an issue/PR if more appropriate, but while you're at it—as the docs say `obj` can be a Python scalar,  am I correct in saying that `bool` and `int` should be part of the top-level `Union` type hint alongside `float`?

rgommers: Technically `bool` and `int` are already covered by `float` (they sub-types), so the main rule we have used for determining whether or not to add `bool` and/or `int` explicitly is: does it help for clarity?

In this case the output dtype is different for `bool` or `int` scalar input, so makes sense to add explicitly.

asmeurer:
Can this be merged? I'm not clear if we should be using Sequence or Union[Tuple,List].

spec/API_specification/manipulation_functions.md
rgommers: I am fairly happy with requiring a tuple or list. `Sequence` is quite broad (e.g. is an array a sequence?), and I'm not that interested in supporting such flexibility for custom containers, it ends up hurting more than helping often.

rgommers:
> Can this be merged? I'm not clear if we should be using Sequence or Union[Tuple,List].

Let's get this in since it's been pending for a while. If there is follow-up discussion on sequence vs list/tuple we can deal with it separately. Thanks @asmeurer, all.

PR #233, opened 2021-07-21
Add i8 to the mixed-type promotion table

asmeurer:
It was already implied by the graph at the top of the page.


rgommers (approved):
Makes sense, thanks Aaron

Issue #232, opened 2021-07-21
Behavior of reductions on empty arrays should be specified


rgommers:
Following up on https://github.com/pytorch/pytorch/issues/61901#issuecomment-884164536. A design rule that NumPy, PyTorch and JAX seem to follow is:

_Return identity if reduction has one, otherwise return nan for reductions that promote integers to floating point dtypes and raise an error for all other._

This is not specified anywhere, and I can't find a previous discussion on it. For element-wise functions there's a "Special Cases" section, but for reductions there isn't. @asmeurer does the test suite cover empty array input to functions at all?

seberg:
> Return identity if reduction has one, otherwise return nan for reductions that promote integers to floating point dtypes and raise an error for all other.

The note on integers is correct, but gets the logic backwards IMO: The reason for the float promotion is that this is a reduce-like operation (not a proper reduction) which includes a true-division in the definition.  E.g. `mean` is:
```
sum(x) / len(x)  # or similar
```
Since `sum([]) == 0`, `len([]) == 0`, and `0 / 0 == NaN` are all clearly well defined, the result is also well defined.  Of course `std` with correction may divide e.g. by `-1` instead of `0` if unchecked; but it is probably very straight forward to extrapolate.

One problem is that `0 / 0 == NaN` works but integer division `0 //0` should probably just raise an error.

These reduce-likes cannot have an `identity`/`initial` value, only a `default` that usually follows cleanly from their definition.  The behaviour in that case should clearly follow that of integer `floor_divide`.

I believe there was an argument/request for the `NaN` return in `median`, probably saying that `median` should match `mean` for `≤2` points (including 0).  But, I do not remember and opinions might vary.  Note that `quantile` does *not* have this behaviour.

asmeurer:
Assuming all the reduction functions being referred to here are in the "statistical functions" section of the spec, I don't have any tests for them written yet in the test suite, beyond the basic signature tests that apply to every function. I agree that the reductions should either explicitly state what they do to empty arrays or else indicate that they are only specified for nonempty arrays. Right now, it is underspecified. I would also note that the statistical functions don't mention the allowed data types of their inputs (presumably they should all be either "numerical" or "floating-point"). 

[...]

PR #234, opened 2021-07-21
Specify input dtypes for the statistical functions

asmeurer:
Functions like mean() that perform divisions require floating-point inputs.
All other functions require numeric inputs.

This was mentioned in #232 but does not fix the main question of that issue
(the behavior of the statistical functions on empty arrays). I can implement
that as well if we come to an agreement about what it should say.

### DIFF
--- a/spec/API_specification/statistical_functions.md
+++ b/spec/API_specification/statistical_functions.md
@@ -24,7 +24,7 @@ Calculates the maximum value of the input array `x`.
 
 -   **x**: _&lt;array&gt;_
 
-    -   input array.
+    -   input array. Should have a numeric data type.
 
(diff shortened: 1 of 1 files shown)


kgryte:
Not convinced that the input dtypes for mean, median, std, or variance should be floating-point. The output dtype, as already specified for mean, std, variance is floating-point. Meaning, how libraries choose to cast/promote internally should be implementation-defined.

asmeurer:
Happy to change it to numeric. I was basing it off the elementwise functions, which only require floating-point input when the output type has to be floating-point.

kgryte:
@asmeurer Yeah, that is true. In which case, having as `floating-point` makes sense. Punts the responsibility to userland to figure out casting/promotion, which was the rationale behind the element-wise transcendentals.

Issue #227, opened 2021-07-14
Return just the solution in `linalg.lstsq`


Lezcano:
**Description**
`linalg.lstsq` is possibly one of the most algorithmically complex operations that the API supports. On top of that, it returns quite a bit of extra information besides the result: the `residuals`, the `rank`, and the `singular_values`.

The specification of the returns was discussed in [this PR](https://github.com/data-apis/array-api/pull/119).

**Problems**

1) If the libraries wanted to expose a lower-level API that returns more information computed during the solution of the system they are free to do so, but a general API should be as simple and orthogonal as possible.
2) Lack of Orthogonality ([Point 5 of the design principles](https://data-apis.org/array-api/latest/extensions/linear_algebra_functions.html#design-principles)): This function does the work of a number of other functions, namely `linalg.svdvals` and `linalg.matrix_rank` or simply just `linalg.svd`.
3) Lack of orthogonality 2. This function is a more general version of `linalg.solve`, which just returns the solutions. As such, their APIs should be comparable.
4) Usability. This function is, by far, the most complex operation in the whole API. We should provide an API for it that is easy to use by the non-expert.
5) The 4th return value assumes that we are solving the problem via SVD. This needn’t be the case and it goes against Point 4 of the design principles. This problem can be also solved via [QR with pivoting](https://www.netlib.org/lapack/lug/node42.html). In fact, this is actually faster in many practical situations as long as the matrix A is well-conditioned.
6) The second return is there because some LAPACK algorithms return it. Some other LAPACK algorithms (e.g. [gels](https://www.netlib.org/lapack/explore-html/d7/d3b/group__double_g_esolve_ga225c8efde208eaf246882df48e590eac.html)) do not return these values in some cases, and they would need to be computed manually. In the best case, one would need to instantiate a large matrix of zeros, in the worst case, an extra matrix-matrix multiplication would be needed.

**Proposal**
Let `linalg.lstsq` return just the solution of the problem.

**TODO**
Have a look in scipy or other libraries and see how is this function used, whether users do actually use all the returned values or just the solution. It may very well be the case that users are just using the solution in most cases, similar to what happened with [`full_matrices=False` in SVD](https://github.com/data-apis/array-api/issues/214#issuecomment-878262647)

cc @asmeurer @rgommers @kgryte

kgryte:
[...]

PR #235, opened 2021-07-26
Replace lstsq with lstsq_solve

kgryte:
This PR

-   resolves [gh-227](https://github.com/data-apis/array-api/issues/227) by removing `linalg.lstsq` and replacing with `linalg.lstsq_solve`.
-   in contrast to `linalg.lstsq`, `linalg.lstsq_solve` **only** returns the least-squares solution. For singular values, rank, and residuals, users would be required to use other APIs already present in the specification (e.g., `linalg.svdvals`, `linalg.matrix_rank`, etc). This is consistent with linear algebra design principles in which functionality should be orthogonal and not duplicated across APIs.

## Notes

-   the name `lstsq_solve` was suggested by @Lezcano as an alternative name for `lstsq` only returning the solutions which does not conflict with existing API names in NumPy and SciPy (e.g., `least_squares`, `lsqr`, etc).

### DIFF
--- a/spec/extensions/linear_algebra_functions.md
+++ b/spec/extensions/linear_algebra_functions.md
@@ -267,8 +267,8 @@ Computes the multiplicative inverse of a square matrix (or a stack of square mat
 
     -   an array containing the multiplicative inverses. The returned array must have a floating-point data type determined by {ref}`type-promotion` and must have the same shape as `x`.
 
-(function-linalg-lstsq)=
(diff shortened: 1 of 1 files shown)


kgryte:
As discussed during the weekly consortium meetings (2021-07-29), we decided to remove `lstsq` altogether given its relatively low usage and difficulty in standardizing (e.g., consider number of variants in SciPy and elsewhere). Should similar functionality be desired in the future, can be addressed in a future revision of the Array API Specification.

Closing...

Issue #213, opened 2021-07-01
Proposal: split `linalg.norm` into dedicated matrix and vector APIs


kgryte:
[`linalg.norm`](https://github.com/data-apis/array-api/blob/cf5aa666017b8a8a2117aa1cb1770eb1cf841942/spec/extensions/linear_algebra_functions.md#linalgnormx---axisnone-keepdimsfalse-ordnone) was added to the specification in [gh-20](https://github.com/data-apis/array-api/pull/20) and subsequently placed in a `linalg` extension in [gh-182](https://github.com/data-apis/array-api/pull/182).

## Background

The current [API](https://github.com/data-apis/array-api/blob/cf5aa666017b8a8a2117aa1cb1770eb1cf841942/spec/extensions/linear_algebra_functions.md#linalgnormx---axisnone-keepdimsfalse-ordnone) is derived from the comparison of API [signatures](https://github.com/data-apis/array-api-comparison/blob/7260fe1442bd325fe6f821ed8953eabdd52582cb/signatures/linalg/norm.md) across array libraries. The initial intent was to codify existing practice among array libraries to ensure API consistency; less attention was given to whether the `norm` API was sound/desirable to begin with.

## Problem

`norm` is easily one of the most complex APIs in the [specification](https://github.com/data-apis/array-api/blob/cf5aa666017b8a8a2117aa1cb1770eb1cf841942/spec/extensions/linear_algebra_functions.md#linalgnormx---axisnone-keepdimsfalse-ordnone). Depending on the dimensionality of `x` and the values of the function's keyword arguments, many different behaviors are possible. For example,

-   when `axis` is

    -   `None` and `x` is
    
        -   one-dimensional: vector norm.
        -   two-dimensional: matrix norm.
        -   n-dimensional: vector norm over flattened array.

    -   an `int`, then specifies dimension along which to compute the vector norm.

    -   a 2-tuple, then specifies the batch dimensions for computing the matrix norm.

Furthermore, the `ord` keyword is constrained based on whether one is computing a matrix or vector norm (and thus, the dimensionality of `x` and the value of `axis`). For example,

-   the `ord` values `fro` or `nuc` are strictly matrix norms and thus invalid when provided vectors.
-   when `ord` is an `int` value other than `+-1` and `+-2`, matrix norms are not currently supported.

Consequently, from an implementation perspective, disentangling valid combinations can be challenging and portends brittle and hard-to-maintain code.

[...]

PR #236, opened 2021-07-26
Split `linalg.norm` into separate vector and matrix APIs

kgryte:
This PR

-   resolves [gh-213](https://github.com/data-apis/array-api/issues/213) by splitting the `linalg.norm` API into separate matrix and vector APIs for computing matrix and vector norms, respectively.

## Notes

-   For `linalg.vector_norm`, the default `axis` is `None`, in order to maintain NumPy compatibility (i.e., the default behavior is to compute the norm over all array values). Torch also does this at the time of this PR.
-   For `linalg.vector_norm`, `axis` is allowed to be an n-tuple along which to compute batched vector norms.

### DIFF
--- a/spec/extensions/linear_algebra_functions.md
+++ b/spec/extensions/linear_algebra_functions.md
@@ -302,6 +302,57 @@ Returns the least-squares solution to a linear matrix equation `Ax = b`.
 
 Alias for {ref}`function-matmul`.
 
+(function-linalg-matrix-norm)=
+### linalg.matrix_norm(x, /, *, axis=(-2, -1), keepdims=False, ord='fro')
+
+Computes the matrix norm of a matrix (or a stack of matrices) `x`.
+
+#### Parameters
+
+-   **x**: _&lt;array&gt;_
+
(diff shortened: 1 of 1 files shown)


kgryte:
This PR was discussed and encountered no objections during the consortium meeting held on 2021-07-29. If further revisions are necessary, they can be addressed in follow-up PRs based on array library implementation feedback.

arogozhnikov:
Hi @kgryte I see this PR merged a while ago, and I can find function via search in array api website, but I don't see either of norms mentioned here:
https://data-apis.org/array-api/latest/API_specification/linear_algebra_functions.html

lucascolley:
Hi @arogozhnikov , these functions are part of the Linear Algebra extension module, which you can find at https://data-apis.org/array-api/latest/extensions/linear_algebra_functions.html.

Your link goes to the few linear algebra functions which are deemed fundamental enough to be required in the main namespace (see more about extensions at https://data-apis.org/array-api/latest/extensions/index.html).

arogozhnikov:
thanks @lucascolley, now I get the difference

Issue #202, opened 2021-06-16
Specify casting rules and accepted input dtypes for reductions better


rgommers:
Reductions were added in PR gh-17, based on discussion in gh-10. There was quite a bit of discussion in calls as well around reductions (e.g., which ones to support, returning 0-D arrays and not scalars, naming) but not about casting rules and accepted input dtypes. It turns out that this is pretty inconsistent between libraries. Here's a script that compares `sum`, `std` and `prod`:

```python
import numpy as np
import dask.array as da
import torch
import tensorflow as tf
import jax.numpy as jnp
import mxnet
try:
    import cupy as cp
except ImportError:
    # CuPy is GPU-only, so may not be available
    cp = None


def ones(mod, shape):
    # Create a (3, 2)-shaped array of int8 1's
    if mod in (da, mxnet):
        x = mod.ones(shape, dtype=np.int8)  # MXNet doesn't have dtype literals
    else:
        x = mod.ones(shape, dtype=mod.int8)

    return x

def sum(mod, x):
    if mod == tf:
        y = tf.math.reduce_sum(x)
    else:
        y = mod.sum(x)

    return y


def std(mod, x):
    if mod == tf:
        y = tf.math.reduce_std(x)
    elif mod == mxnet:
        y = mxnet.std(x)
    else:
        y = mod.std(x)

    return y


def prod(mod, x):
    if mod == tf:
        y = tf.math.reduce_prod(x)
    else:
        y = mod.prod(x)

    return y


libraries = {
    'numpy': np,
    'pytorch': torch,
    'mxnet': mxnet.np,
    'dask': da,
    'tensorflow': tf,
    'jax': jnp,
}

if cp is not None:
    libraries['cupy'] = cp

results = libraries.copy()

# A separate call to get rid of TF and JAX noise:
shape = (3, 2)
_ = sum(tf, ones(tf, shape))
_ = sum(jnp, ones(jnp, shape))


print("\nsum(int8_array)\n" + "-"*15)
for name, mod in libraries.items():
    dtype = sum(mod, ones(mod, shape)).dtype
    print(f'{name}: {dtype}')

print("\nstd(int8_array)\n" + "-"*15)
for name, mod in libraries.items():
    try:
        dtype = std(mod, ones(mod, shape)).dtype
        print(f'{name}: {dtype}')
    except Exception as e:
        print(f'{name}: {repr(e)}')

print("\nprod(int8_array)\n" + "-"*16)
for name, mod in libraries.items():
[...]

PR #238, opened 2021-07-26
Add `dtype` keyword argument to `sum` and `prod`

kgryte:
This PR

-   adds a `dtype` keyword argument to `sum` and `prod` based on discussions in [gh-202](https://github.com/data-apis/array-api/issues/202) and in consortium meetings. Adding such a keyword aids in preventing overflows during accumulation.
-   specifies that the default behavior (i.e., with `dtype=None`) be that the returned array have the default data type of the same kind as the data type of the input array, so long as the default data type is equal to or larger than the input array data type.

    For example, if provided an `int8` array and the default integer data type is `int32`, the returned array should have data type `int32`. If provided an `int64` array and the default integer data type is `int32`, the returned array should have data type `int64`.

    In the event a user wants to return an array having the same data type as the input array, a user can explicitly specify the return array `dtype`. The default behavior is intended to prevent potential accumulation footguns.

### DIFF
--- a/spec/API_specification/statistical_functions.md
+++ b/spec/API_specification/statistical_functions.md
@@ -24,7 +24,7 @@ Calculates the maximum value of the input array `x`.
 
 -   **x**: _&lt;array&gt;_
 
-    -   input array.
+    -   input array. Should have a numeric data type.
 
 -   **axis**: _Optional\[ Union\[ int, Tuple\[ int, ... ] ] ]_
 
@@ -49,7 +49,7 @@ Calculates the arithmetic mean of the input array `x`.
 
 -   **x**: _&lt;array&gt;_
 
-    -   input array.
+    -   input array. Should have a floating-point data type.
 
 -   **axis**: _Optional\[ Union\[ int, Tuple\[ int, ... ] ] ]_
 
@@ -74,7 +74,7 @@ Calculates the minimum value of the input array `x`.
 
 -   **x**: _&lt;array&gt;_
 
-    -   input array.
+    -   input array. Should have a numeric data type.
 
 -   **axis**: _Optional\[ Union\[ int, Tuple\[ int, ... ] ] ]_
 
@@ -99,7 +99,7 @@ Calculates the product of input array `x` elements.
 
 -   **x**: _&lt;array&gt;_
 
-    -   input array.
+    -   input array. Should have a numeric data type.
 
 -   **axis**: _Optional\[ Union\[ int, Tuple\[ int, ... ] ] ]_
 
@@ -137,7 +137,7 @@ Calculates the standard deviation of the input array `x`.
 
 -   **x**: _&lt;array&gt;_
 
-    -   input array.
+    -   input array. Should have a floating-point data type.
 
 -   **axis**: _Optional\[ Union\[ int, Tuple\[ int, ... ] ] ]_
 
@@ -166,7 +166,7 @@ Calculates the sum of the input array `x`.
 
 -   **x**: _&lt;array&gt;_
 
-    -   input array.
+    -   input array. Should have a numeric data type.
 
 -   **axis**: _Optional\[ Union\[ int, Tuple\[ int, ... ] ] ]_
 
@@ -204,7 +204,7 @@ Calculates the variance of the input array `x`.
 
 -   **x**: _&lt;array&gt;_
 
-    -   input array.
+    -   input array. Should have a floating-point data type.
 
 -   **axis**: _Optional\[ Union\[ int, Tuple\[ int, ... ] ] ]_
 
--- a/spec/API_specification/statistical_functions.md
+++ b/spec/API_specification/statistical_functions.md
@@ -107,7 +107,7 @@ Calculates the product of input array `x` elements.
 
 -   **dtype**: _Optional\[ &lt;dtype&gt; ]_
 
-    -   data type of the returned array. If `None`, the returned array must have the same data type as `x`. If specified, the input array should be cast to the specified data type before computing the product. Default: `None`.
+    -   data type of the returned array. If `None`, the returned array must have the default data type corresponding to the data type "kind" (integer or floating-point) of `x`. If the data type (either specified or resolved) differs from the data type of `x`, the input array should be cast to the specified data type before computing the product. Default: `None`.
 
         ```{note}
 
@@ -170,7 +170,7 @@ Calculates the sum of the input array `x`.
 
 -   **dtype**: _Optional\[ &lt;dtype&gt; ]_
 
-    -   data type of the returned array. If `None`, the returned array must have the same data type as `x`. If specified, the input array should be cast to the specified data type before computing the product. Default: `None`.
+    -   data type of the returned array. If `None`, the returned array must have have the default data type corresponding to the data type "kind" (integer or floating-point) of `x`. If the data type (either specified or resolved) differs from the data type of `x`, the input array should be cast to the specified data type before computing the product. Default: `None`.
 
         ```{note}
 
--- a/spec/API_specification/statistical_functions.md
+++ b/spec/API_specification/statistical_functions.md
@@ -193,7 +193,7 @@ Calculates the sum of the input array `x`.
 
 -   **out**: _&lt;array&gt;_
 
-    -   if the sum was computed over the entire array, a zero-dimensional array containing the sum; otherwise, an array containing the sums.
+    -   if the sum was computed over the entire array, a zero-dimensional array containing the sum; otherwise, an array containing the sums. The returned array must have a data type as described by the `dtype` parameter above.
 
 (function-var)=
 ### var(x, /, *, axis=None, correction=0.0, keepdims=False)
--- a/spec/API_specification/statistical_functions.md
+++ b/spec/API_specification/statistical_functions.md
@@ -169,7 +169,7 @@ Calculates the sum of the input array `x`.
 
 -   **dtype**: _Optional\[ &lt;dtype&gt; ]_
 
-    -   data type of the returned array. If `None`, the returned array must have have the default data type corresponding to the data type "kind" (integer or floating-point) of `x`. If the data type (either specified or resolved) differs from the data type of `x`, the input array should be cast to the specified data type before computing the sum. Default: `None`.
+    -   data type of the returned array. If `None`, the returned array must have the default data type corresponding to the data type "kind" (integer or floating-point) of `x`. If the data type (either specified or resolved) differs from the data type of `x`, the input array should be cast to the specified data type before computing the sum. Default: `None`.
 
         ```{note}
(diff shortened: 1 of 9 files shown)


spec/API_specification/statistical_functions.md
```diff
@@ -105,6 +105,15 @@ Calculates the product of input array `x` elements.
 
     -   axis or axes along which products must be computed. By default, the product must be computed over the entire array. If a tuple of integers, products must be computed over multiple axes. Default: `None`.
 
+-   **dtype**: _Optional\[ &lt;dtype&gt; ]_
+
+    -   data type of the returned array. If `None`, the returned array must have the default data type corresponding to the data type "kind" (integer or floating-point) of `x`. If the data type (either specified or resolved) differs from the data type of `x`, the input array should be cast to the specified data type before computing the product. Default: `None`.
+
+        ```{note}
+
```
leofang, line 113: I don't think we need an extra black line?
```suggestion
```

spec/API_specification/statistical_functions.md
```diff
@@ -105,6 +105,15 @@ Calculates the product of input array `x` elements.
 
     -   axis or axes along which products must be computed. By default, the product must be computed over the entire array. If a tuple of integers, products must be computed over multiple axes. Default: `None`.
 
+-   **dtype**: _Optional\[ &lt;dtype&gt; ]_
```
leofang, line 108: This is a question not related to the PR itself, just a matter of the style consistency across all documents. Do we always use `Optional` for all kw-only arguments? For example, `keepdims` from below does not use it, but `axis` from above does.

spec/API_specification/statistical_functions.md
```diff
@@ -159,6 +168,15 @@ Calculates the sum of the input array `x`.
 
     -   axis or axes along which sums must be computed. By default, the sum must be computed over the entire array. If a tuple of integers, sums must be computed over multiple axes. Default: `None`.
 
+-   **dtype**: _Optional\[ &lt;dtype&gt; ]_
+
+    -   data type of the returned array. If `None`, the returned array must have have the default data type corresponding to the data type "kind" (integer or floating-point) of `x`. If the data type (either specified or resolved) differs from the data type of `x`, the input array should be cast to the specified data type before computing the sum. Default: `None`.
```
leofang, line 173:
```suggestion
    -   data type of the returned array. If `None`, the returned array must have the default data type corresponding to the data type "kind" (integer or floating-point) of `x`. If the data type (either specified or resolved) differs from the data type of `x`, the input array should be cast to the specified data type before computing the sum. Default: `None`.
```

spec/API_specification/statistical_functions.md
```diff
@@ -159,6 +168,15 @@ Calculates the sum of the input array `x`.
...
+
+        ```{note}
+
```
leofang, line 176: Ditto.
```suggestion
```

leofang (changes requested):
LGTM except for a few nitpicks.

spec/API_specification/statistical_functions.md
```diff
@@ -105,6 +105,15 @@ Calculates the product of input array `x` elements.
 
     -   axis or axes along which products must be computed. By default, the product must be computed over the entire array. If a tuple of integers, products must be computed over multiple axes. Default: `None`.
 
+-   **dtype**: _Optional\[ &lt;dtype&gt; ]_
+
+    -   data type of the returned array. If `None`, the returned array must have the default data type corresponding to the data type "kind" (integer or floating-point) of `x`. If the data type (either specified or resolved) differs from the data type of `x`, the input array should be cast to the specified data type before computing the product. Default: `None`.
```
asmeurer, line 110: If the default dtype is 32-bit and the input is 64-bit, won't this mean down casting by default?

spec/API_specification/statistical_functions.md
kgryte: @leofang Thanks for the question. No, we don't always use `Optional` for keyword-only arguments. We use [`Optional`](https://docs.python.org/3/library/typing.html#typing.Optional) for those arguments whose type is the `Union[X, None]`. In which case, for `keepdims`, we would not use `Optional` as `keepdims` cannot be set to `None`, but only a boolean `True` or `False`. Hope this clarifies!

spec/API_specification/statistical_functions.md
kgryte: @asmeurer Thanks for pointing this out. Will reword.

spec/API_specification/statistical_functions.md
leofang: Thanks Athan! That helps a lot!

kgryte: Of course! :)

spec/API_specification/statistical_functions.md
kgryte: I've updated the wording to avoid downcasting.

spec/API_specification/statistical_functions.md
```diff
@@ -113,7 +126,7 @@ Calculates the product of input array `x` elements.
 
 -   **out**: _&lt;array&gt;_
 
-    -   if the product was computed over the entire array, a zero-dimensional array containing the product; otherwise, a non-zero-dimensional array containing the products. The returned array must have the same data type as `x`.
+    -   if the product was computed over the entire array, a zero-dimensional array containing the product; otherwise, a non-zero-dimensional array containing the products.
```
asmeurer, line 129: How about
```suggestion
    -   if the product was computed over the entire array, a zero-dimensional array containing the product; otherwise, a non-zero-dimensional array containing the products. The returned array has a data type as described by the `dtype` parameter above.
```

spec/API_specification/statistical_functions.md
```diff
@@ -167,7 +193,7 @@ Calculates the sum of the input array `x`.
 
 -   **out**: _&lt;array&gt;_
 
-    -   if the sum was computed over the entire array, a zero-dimensional array containing the sum; otherwise, an array containing the sums. The returned array must have the same data type as `x`.
+    -   if the sum was computed over the entire array, a zero-dimensional array containing the sum; otherwise, an array containing the sums.
```
kgryte, line 196:
```suggestion
    -   if the sum was computed over the entire array, a zero-dimensional array containing the sum; otherwise, an array containing the sums. The returned array must have a data type as described by the `dtype` parameter above.
```

asmeurer:
This looks good to me.

asmeurer:
Actually scratch that, I just noticed a potential problem. NumPy's `dtype` argument to `sum` is documented as follows:

>     dtype : dtype, optional
>        The type of the returned array and of the accumulator in which the
>        elements are summed.  The dtype of `a` is used by default unless `a`
>        has an integer dtype of less precision than the default platform
>        integer.  In that case, if `a` is signed then the platform integer
>        is used while if `a` is unsigned then an unsigned integer of the
>        same precision as the platform integer is used.

As written here now, a uint64 array would cast to int64 by default, which would be a downcast (uint64 can represent values that don't fit in the signed int64).

spec/API_specification/statistical_functions.md
```diff
@@ -105,6 +105,19 @@ Calculates the product of input array `x` elements.
 
     -   axis or axes along which products must be computed. By default, the product must be computed over the entire array. If a tuple of integers, products must be computed over multiple axes. Default: `None`.
 
+-   **dtype**: _Optional\[ &lt;dtype&gt; ]_
+
+    -   data type of the returned array. If `None`,
+    
+        -   if the default data type corresponding to the data type "kind" (integer or floating-point) of `x` is smaller than the data type of `x` (e.g., `x` has data type `int64` and the default data type is `int32`), the returned array must have the same data type as `x`.
```
kgryte, line 112:
```suggestion
        -   if the default data type corresponding to the data type "kind" (integer or floating-point) of `x` has a smaller range of values than the data type of `x` (e.g., `x` has data type `int64` and the default data type is `int32`, or `x` has data type `uint64` and the default data type is `int64`), the returned array must have the same data type as `x`.
```

spec/API_specification/statistical_functions.md
```diff
@@ -159,6 +172,19 @@ Calculates the sum of the input array `x`.
 
     -   axis or axes along which sums must be computed. By default, the sum must be computed over the entire array. If a tuple of integers, sums must be computed over multiple axes. Default: `None`.
 
+-   **dtype**: _Optional\[ &lt;dtype&gt; ]_
+
+    -   data type of the returned array. If `None`,
+    
+        -   if the default data type corresponding to the data type "kind" (integer or floating-point) of `x` is smaller than the data type of `x` (e.g., `x` has data type `int64` and the default data type is `int32`), the returned array must have the same data type as `x`.
```
kgryte, line 179:
```suggestion
        -   if the default data type corresponding to the data type "kind" (integer or floating-point) of `x` has a smaller range of values than the data type of `x` (e.g., `x` has data type `int64` and the default data type is `int32`, or `x` has data type `uint64` and the default data type is `int64`), the returned array must have the same data type as `x`.
```

asmeurer:
OK, I think that wording makes it clearer. This does look good now.

kgryte:
@asmeurer I've updated the text to clarify promotion guidance.

kgryte:
This has been reviewed. If any further changes are necessary, we can add them in a follow-up PR.

seberg:
As the quoted docs say, numpy only applies this to integers (and bool).  It does *not* upcast floats, in particular float32.  Is that part intentional, or just a typo?

kgryte:
@seberg Upcasting floats is intentional, as, in principle (and while not currently required by the spec), also applies to `float16` (`bfloat16`) in accelerator libraries, where overflow/precision concerns are still valid, albeit less of common concern than, e.g., `int8`.

seberg:
@kgryte ah OK. I have not thought much about what I would consider ideal, it is/was just a bit surprising.

kgryte:
@seberg Were we to make a distinction between integral and floating dtypes for how promotion is handled relative to the default dtypes, this would add complexity to user mental models in terms of when promotion occurs.

And further, the same concerns motivating promotion to the default dtype for smaller integral dtypes applies to floating dtypes. Namely, that accumulation can lead to overflow and erroneous results. For example, for `float32`, the maximum safe integer is less than `2e7`, leading to potential user footguns when summing and/or multiplying many large positive numbers. If the default floating dtype for a specification-compliant array library is `float64` and `dtype` is `None`, the library should promote to `float64` to avoid overflow and/or precision errors.

When an array library's default floating dtype is `float64`, if a user wants to return a `float32` array when providing a `float32` array, they can explicitly specify via the `dtype` kwarg that they wish to receive an output array of the same dtype. (This same logic applies to a user who provides an `int8` array and wants an `int8` array in return). Whether an array library chooses to accumulate internally in `float32` is, however, an implementation detail.

PR #239, opened 2021-08-12
Fix the logaddexp special cases

asmeurer:
As previously written, they would fail in the test suite because of ambiguity
about the value of logaddexp(nan, inf).


kgryte (approved):
LGTM.

PR #240, opened 2021-08-16
Remove `linalg.lstsq`

kgryte:
This PR

-   removes `linalg.lstsq` (per consensus vote held during the consortium meeting held 2021-07-29). This PR supersedes [gh-235](https://github.com/data-apis/array-api/pull/235), which attempted to replace `lstsq` with a simpler API.
-   closes [gh-227](https://github.com/data-apis/array-api/issues/227) and [gh-235](https://github.com/data-apis/array-api/pull/235).

## Background

Initial discussion on `linalg.lstsq` API design can be found in [gh-227](https://github.com/data-apis/array-api/issues/227). There, it was held that the current API design posed undue burden on specification implementers due to API complexity (returning solutions, ranks, residuals, and singular values) and violated various linear algebra API design principles (particularly orthogonality).

Based on ensuing discussion in consortium meetings, consensus was found in simply removing the `linalg.lstsq` API altogether. This API finds low usage among downstream libraries (see [data](https://github.com/data-apis/array-api-comparison/blob/bb9922d42d2cc9d040bb0b60612f345356442fc2/data/common_apis_ranks.csv#L191)) and has several variants due to specialized use cases (see SciPy).

Should this or similar functionality be found to be desirable in the future, specification can happen in a future revision of the standard.

### DIFF
--- a/spec/extensions/linear_algebra_functions.md
+++ b/spec/extensions/linear_algebra_functions.md
@@ -267,36 +267,6 @@ Computes the multiplicative inverse of a square matrix (or a stack of square mat
 
     -   an array containing the multiplicative inverses. The returned array must have a floating-point data type determined by {ref}`type-promotion` and must have the same shape as `x`.
 
-(function-linalg-lstsq)=
-### linalg.lstsq(x1, x2, /, *, rtol=None)
-
-Returns the least-squares solution to a linear matrix equation `Ax = b`.
-
-#### Parameters
-
--   **x1**: _&lt;array&gt;_
-
-    -   coefficient array `A` having shape `(..., M, N)` and whose innermost two dimensions form `MxN` matrices. Should have a floating-point data type.
-
--   **x2**: _&lt;array&gt;_
-
-    -   ordinate (or "dependent variable") array `b`. If `x2` has shape `(..., M)`, `x2` is equivalent to an array having shape `(..., M, 1)`, and `shape(x2)` must be compatible with `shape(x1)[:-1]` (see {ref}`broadcasting`). If `x2` has shape `(..., M, K)`, each column `k` defines a set of ordinate values for which to compute a solution, and `shape(x2)[:-1]` must be compatible with `shape(x1)[:-1]` (see {ref}`broadcasting`). Should have a floating-point data type.
-
--   **rtol**: _Optional\[ Union\[ float, &lt;array&gt; ] ]_
-
-    -   relative tolerance for small singular values. Singular values less than or equal to `rtol * largest_singular_value` are set to zero. If a `float`, the value is equivalent to a zero-dimensional array having a data type determined by {ref}`type-promotion` (as applied to `x1` and `x2`) and must be broadcast against each matrix. If an `array`, must have a floating-point data type and must be compatible with `shape(x1)[:-2]` (see {ref}`broadcasting`). If `None`, the default value is `max(M, N) * eps`, where `eps` must be the machine epsilon associated with the floating-point data type determined by {ref}`type-promotion` (as applied to `x1` and `x2`). Default: `None`.
-
-#### Returns
-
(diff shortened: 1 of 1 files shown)


kgryte:
As this change was discussed and agreed upon in the consortium meeting held 2021-07-29, will merge. As mentioned in the OP, re-adding to the specification at a future date can be considered as part of a future revision of this standard.

rgommers:
I'll note that this came up in https://github.com/data-apis/array-api/issues/403#issuecomment-1071308357, quoting:

*I realize that `lstsq` was removed as per a previous discussion (and it's understandable given that the API is a bit crufty), but then our code base has 26 unique call sites of that alone, since we're dealing mostly with engineering and stats. One might reasonably assume that backends that already have optimized implementations of that (all 6 do, and torch/tf support batched matrices) will provide it anyway in their array API namespace. However, we do worry that, given that it has been deliberately removed, we can't be sure that some of the existing backends won't be encouraged to minimize their maintenance surface and drop functions like that from their formerly numpy-compatible (soon array API compatible) namespace, forcing users like us to deal with their raw EagerTensor, xla_extension.DeviceArray, or whatever it may be called, and go find it in whichever ancestral tensorflow namespace the functionality may have been buried in before. We're wondering if a tradeoff could be made where e.g., some of the rarely-used outputs could be marked as as "reserved" placeholder and are allowed to hold unspecified values (e.g., None) until perhaps at some future date the API specifies them. There's also the option to go the same route as with `svd`, where some arguments were removed in the interest of simplicity.*

It also came up in SciPy, where it was hit in one of the first APIs on which a conversion was tried (`signal.welch`): https://github.com/scipy/scipy/issues/18286#issuecomment-1542869440

steppi:
To elaborate on why `dgelsd` is better numerically than directly computing the pseudo-inverse with SVD.

Suppose we want to find

$$\min_{x} \left|b - Ax\right|^2$$

where $b$ is a known $n \times 1$ column vector for the right hand side, $A$ is a possibly rank deficient or ill conditioned $m \times n$ matrix, and $x$ is an unknown $m \times 1$ column vector of coefficients.

Rather than directly computing the svd of $A$, $A = U\Sigma V^T$ and using that pseudo-inverse of $A$ is 

$$A^{+} = V\Sigma^{+}u^T$$

where $\Sigma^{+}$ is the easy to compute pseudo-inverse of $\Sigma$

what `dgelsd` does is first use householder reflections to reduce $A$ to bidiagonal form using [dgebrd](https://www.netlib.org/lapack/explore-html/dc/d1c/group__gebrd_ga1314f3a906c316785fe32996698901a8.html#ga1314f3a906c316785fe32996698901a8), finding orthogonal $P$ and $Q$ such that 

$$Q^{T} A P = B$$

for $B$ bidiagonal. The original minimization problem then can be modified to

$$\min_{w} \left|Q^Tb - Bw\right|^2$$

where $w = P^{T}X$.

This bidiagonal least squares problem is solved with the specialized routine [dlalsd](https://www.netlib.org/lapack/explore-html/dc/ded/group__lalsd_ga44ce5c1bbb2dd918c6adb336f5e88e81.html#ga44ce5c1bbb2dd918c6adb336f5e88e81)

which essentially just computes the pseudo-inverse using SVD and does a matmul, but the bidiagonal problem tends to be better behaved numerically than the general problem. One can then recover the solution with $x = Pw = PP^Tx$. There is also a specialized routine [dormbr](https://www.netlib.org/lapack/explore-html/db/d4f/group__unmbr_ga303e3a7dec5764f7856cf65cd5aad0e1.html#ga303e3a7dec5764f7856cf65cd5aad0e1) specifically for the left multiplications by orthogonal matrices produced by `dgebrd`, exploiting the structure of these matrices.

This sidesteps needing to calculate the SVD of $A$, just the easier to compute one of the bidiagonal reduction, leaving us with a numerically better behaved routine.

kgryte:
@steppi And also https://github.com/data-apis/array-api/issues/227

PR #222, opened 2021-07-12
Change defaults of trace and diagonal to `axis1=-2`, `axis2=-1`.

Lezcano:
As per the title.

Fixes https://github.com/data-apis/array-api/issues/215

Also corrected a small errata.

cc @rgommers @asmeurer

Issue #215, opened 2021-07-01
[Consistency] Default axes in `linalg.diagonal`  and `linalg.trace`.


Lezcano:
The functions [`linalg.trace`](https://data-apis.org/array-api/latest/extensions/linear_algebra_functions.html?linalg-eigh-x-upper-false#linalg-trace-x-axis1-0-axis2-1-offset-0) and [`linalg.diagonal`](https://data-apis.org/array-api/latest/extensions/linear_algebra_functions.html?linalg-eigh-x-upper-false#linalg-diagonal-x-axis1-0-axis2-1-offset-0) have as default `axis1=0, axis2=1`.

This goes against the rest of the documentation, in which a batch of matrices is defined as a tensor with shape `(..., M, N)`.

It would be more consistent with the rest of the API to have `axis1=-2, axis2=-1` to handle batches of matrices correctly by default.

leofang:
Wouldn't it be better to just remove `axis1` and `axis2` entirely and always use the last two axes? This makes it more consistent with other APIs. Users who need to compute trace/diagonal along arbitrary two axes can always swap the axes first before calling the API.

Lezcano:
At least in PyTorch, it would be as efficient to transpose the diagonals to the back and then call diagonal than it would be calling it with this API. Now, this API may be convenient for some applications? That I do not know. I have never used diagonal with anything other than `-2, -1`, but my experience is very biased towards working with matrices.

leofang:
In CuPy it's also as efficient, given that transpose/swapaxes only return views. I guess this suggests for both PyTorch and CuPy (and likely NumPy) removing `axis1`/`axis2` is safe. 🙂

rgommers:
`trace` is not heavily used (there's no data in https://github.com/data-apis/python-record-api, and a search in SciPy seems to confirm it's not used there at all). I suspect when it is used, it's indeed with `-2, -1`. 

> Wouldn't it be better to just remove `axis1` and `axis2` entirely and always use the last two axes? This makes it more consistent with other linalg APIs, and is mathematically more intuitive.

+1 makes sense to me

rgommers:
Fixed by gh-241, which removes the keywords. Thanks @Lezcano and @leofang

PR #241, opened 2021-08-16
Remove `axisN` keyword arguments from `linalg.diag` and `linalg.trace`

kgryte:
This PR

-   removes the `axis1` and `axis2` keyword arguments from `linalg.diag` and `linalg.trace`.
-   updates the API to perform batching over the last two dimensions.
-   supersedes [gh-222](https://github.com/data-apis/array-api/pull/222).

## Background

`linalg.diag` and `linalg.trace`, as currently specified, are unique in their parameterization. Namely, for APIs which allow specification of axes, common convention is to use `axis` or `axes`. Furthermore, prior art (NumPy) used `0` and `1` as the default axes, which runs counter to the current prevalent practice of batching along the last (innermost) two dimensions.

[gh-215](https://github.com/data-apis/array-api/issues/215) proposed changing the defaults to `-2` and `-1`, respectively. Subsequent discussion (both in [gh-215](https://github.com/data-apis/array-api/issues/215) and in consortium meeting 2021-07-29) contended that they should be removed altogether.

This PR follows consensus found in the consortium meeting to simply remove the `axisN` keyword arguments. If future needs arise to specify axes in the future, such additions can be made to the specification in a future revision.

### DIFF
--- a/spec/extensions/linear_algebra_functions.md
+++ b/spec/extensions/linear_algebra_functions.md
@@ -140,23 +140,15 @@ Returns the determinant of a square matrix (or stack of square matrices) `x`.
     -   if `x` is a two-dimensional array, a zero-dimensional array containing the determinant; otherwise, a non-zero dimensional array containing the determinant for each square matrix. The returned array must have the same data type as `x`.
 
 (function-linalg-diagonal)=
-### linalg.diagonal(x, /, *, axis1=0, axis2=1, offset=0)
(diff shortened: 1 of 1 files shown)


rgommers (approved):
This was decided a while ago and everyone seems happy. Let's get it in. Thanks @kgryte.

leofang:
Thanks, @kgryte!

Issue #237, opened 2021-07-26
Proposal: add support for returning upper and lower triangles of an array


kgryte:
## Proposal

Add APIs for returning the upper `(triu`) and lower (`tril`) triangles of an array.

## Motivation

Currently, the array API specification does not provide APIs for returning the upper and lower triangles of an array.

These APIs were identified as potential candidates in [gh-187](https://github.com/data-apis/array-api/issues/187) for inclusion in the next version of the standard. However, depending on the outcome of [gh-217](https://github.com/data-apis/array-api/issues/217) may be necessary to include in the current version in order to ensure users can maintain equivalent functionality.

## Background

`triu` and `tril` APIs are common across array libraries (see [here](https://github.com/data-apis/array-api-comparison/blob/bb9922d42d2cc9d040bb0b60612f345356442fc2/data/common_apis.csv#L190)). The notable exception being TensorFlow. MXNet seems to have APIs for returning the triangles of an array, but these are not documented in public documentation, so were not picked up during API comparison analysis.

Both `triu` and `tril` should be straightforward to include in the specification, as the signatures are consistent across libraries.

```text.
tril(m, k=0) → array
triu(m, k=0) → array
```

where `k` specifies the diagonal above/below which to zero elements.

leofang:
@kgryte You're 10 mins ahead of me! I was working on https://github.com/cupy/cupy/pull/5555 and just realized the very same need. I think we need them to be able to handle and return a batch of matrices like all other linalg functions, which is currently not possible with NumPy/CuPy.

kgryte:
@leofang I think handling batches could make sense; however, I don't think any array libraries currently support batching, including Torch. Should be a straightforward extension, as should not conflict with any existing NumPy et al usage patterns.

Lezcano:
The docs of `tril` and  `triu` are wrong. PyTorch does support batched inputs for these functions. I agree with @leofang that it would be good to write the specification for batched inputs.

kgryte:
I've opened [gh-243](https://github.com/data-apis/array-api/pull/243) which adds both `tril` and `triu` to the specification, with support for providing a stack of matrices.

Issue #187, opened 2021-05-27
Common APIs across array libraries (1 year later)


kgryte:
## Overview

Similar to [gh-6](https://github.com/data-apis/array-api/issues/6), this issue looks to identify commonalities across array libraries, but only addresses those APIs which are not already included in the array API specification.

Since [gh-6](https://github.com/data-apis/array-api/issues/6) and its analysis,

1.  Every array library add more APIs.
2.  Continued convergence toward NumPy APIs (CuPy, PyTorch, TensorFlow.experimental, mxnet).
3.  Greater agreement among accelerator libraries (e.g., CuPy, mxnet, Torch, TF) wrt special functions not available in NumPy, but available in SciPy.

## Method

Similar to [gh-6](https://github.com/data-apis/array-api/issues/6), the list was compiled by doing the following:

1.  Generating a list of APIs based on publicly documented array APIs (e.g., by scraping website documentation).
2.  Computing the intersection across the individual datasets.

The following libraries were analyzed:

-   **numpy**
-   **cupy**
-   **dask.array**
-   **jax**
-   **mxnet**
-   **pytorch**
-  **tensorflow**

## APIs

The following APIs were found to be common across the above libraries, but not already included in the array API specification:

```text
cbrt
clip
copysign
count_nonzero

deg2rad
diff

erf (scipy)
erfc (scipy)
erfinv (scipy)
erfcinv (scipy)
exp2

gamma (scipy)
gammaln (scipy)

histogram
hypot

i0 (bessel)

logsumexp (accelerators)

nextafter

pad

rad2deg
reciprocal
repeat
rot90
rsqrt (accelerators)
rcbrt (accelerators)

sigmoid (accelerators)

take
tile
tril
triu
top_k (accelerators+dask)

xlogy (scipy)
```

We can split the APIs into the following categories...

### Array Creation

```text
tril
triu
```

### Array Manipulation

```text
pad
repeat
rot90
tile
```

### Special Functions

```text
cbrt
clip
copysign
deg2rad
erf
erfc
erfcinv
erfinv
exp2
gamma
gammaln
hypot
i0
nextafter
rad2deg
reciprocal
rsqrt
rcbrt
sigmoid
xlogy
```

### Reductions

```text
count_nonzero
histogram
logsumexp
top_k
```

### Indexing

```text
take
```

### Other

```text
diff
```

## Next Steps

-   Identify which APIs could be suitable candidates for standardization in the next version of the specification.

[...]

PR #243, opened 2021-08-23
Add specifications for returning upper (`triu`) and lower (`tril`) triangular matrices

kgryte:
This PR

-   resolves [gh-237](https://github.com/data-apis/array-api/issues/237) by adding specifications for returning the lower (`tril`) and upper (`triu`) triangular matrix for one or more matrices.
-   includes a minor textual change to improve grammar.

## Notes

-   The function signatures were derived by comparing the signatures across array libraries ([`tril`](https://github.com/data-apis/array-api-comparison/blob/a2fdc6b7a7ac4728519785550c6cf25a7ca69973/signatures/creation/tril.md), [`triu`](https://github.com/data-apis/array-api-comparison/blob/a2fdc6b7a7ac4728519785550c6cf25a7ca69973/signatures/creation/triu.md)). The signatures across libraries are consistent, with the exception that PyTorch uses `diagonal` as the keyword argument, rather than `k` as used in NumPy et al. This PR chose `k` given its preponderance.
-   TF was the sole library without `tril` and `triu` APIs; however, these are included in the experimental `numpy` namespace.
-   The proposed APIs deviate from NumPy in allowing providing a batch of matrices (as discussed in [gh-237](https://github.com/data-apis/array-api/issues/237)). Supporting batches is part of a more general theme in this specification (see linear algebra) and follows prior art in the PyTorch API.
-   Both `tril` and `triu` were included in the list of potential APIs to standardize for v2 of the specification (see [gh-187](https://github.com/data-apis/array-api/issues/187)); however, these APIs were found to be needed in the current specification (see [gh-226](https://github.com/data-apis/array-api/pull/226) and discussion [gh-217](https://github.com/data-apis/array-api/issues/217)).
-   Once merged, this should allow merging of [gh-226](https://github.com/data-apis/array-api/pull/226).

### DIFF
--- a/spec/API_specification/creation_functions.md
+++ b/spec/API_specification/creation_functions.md
@@ -396,7 +396,7 @@ The lower triangular part of the matrix is defined as the elements on and below 
 
 -   **out**: _&lt;array&gt;_
 
-    -   an array containing the lower triangular part(s). The returned array must have the same shape and data type as `x`. All elements above the specified diagonal `k` must be zeroed.
+    -   an array containing the lower triangular part(s). The returned array must have the same shape and data type as `x`. All elements above the specified diagonal `k` must be zeroed. The returned array should be allocated on the same device as `x`.
 
 (function-triu)=
 ### triu(x, /, *, k=0)
@@ -424,7 +424,7 @@ The upper triangular part of the matrix is defined as the elements on and above 
 
 -   **out**: _&lt;array&gt;_
 
-    -   an array containing the upper triangular part(s). The returned array must have the same shape and data type as `x`. All elements below the specified diagonal `k` must be zeroed.
+    -   an array containing the upper triangular part(s). The returned array must have the same shape and data type as `x`. All elements below the specified diagonal `k` must be zeroed. The returned array should be allocated on the same device as `x`.
 
 (function-zeros)=
 ### zeros(shape, *, dtype=None, device=None)
--- a/spec/API_specification/creation_functions.md
+++ b/spec/API_specification/creation_functions.md
@@ -391,6 +391,7 @@ The lower triangular part of the matrix is defined as the elements on and below 
 
         ```{note}
         The main diagonal is defined as the set of indices `{(i, i)}` for `i` on the interval `[0, min(M, N) - 1]`.
+        ```
 
 #### Returns
 
@@ -419,6 +420,7 @@ The upper triangular part of the matrix is defined as the elements on and above 
 
         ```{note}
         The main diagonal is defined as the set of indices `{(i, i)}` for `i` on the interval `[0, min(M, N) - 1]`.
+        ```
 
 #### Returns
 
--- a/spec/API_specification/creation_functions.md
+++ b/spec/API_specification/creation_functions.md
@@ -371,7 +371,7 @@ Returns a new array filled with ones and having the same `shape` as an input arr
     -   an array having the same shape as `x` and filled with ones.
 
 (function-tril)=
-### tril(x, /, *, k=0, device=None)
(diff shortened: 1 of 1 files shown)


rgommers (changes requested):
Thanks @kgryte. I see two issues here:
- the `k` keyword is not named `k` in all libraries. E.g. PyTorch uses `diagonal=0`.
- these aren't really creation functions - the input is already an array and this uses a subset of the data. So it looks to me like these don't quite belong in `creation_functions.md` nor should have a `device` keyword.

kgryte:
@rgommers NumPy classifies both `tril` and `triu` as [creation functions](https://numpy.org/doc/stable/reference/routines.array-creation.html#building-matrices). Presumably, this is because NumPy chooses to return a copy of an array for both functions, thus "creating" new arrays. I suppose an array library might choose to return a view, but I am not sure how likely this would be.

rgommers:
> @rgommers NumPy classifies both `tril` and `triu` as [creation functions](https://numpy.org/doc/stable/reference/routines.array-creation.html#building-matrices). Presumably, this is because NumPy chooses to return a copy of an array for both functions, thus "creating" new arrays. I suppose an array library might choose to return a view, but I am not sure how likely this would be.

Hmm. Yes a copy is likely needed, because the view on an upper or lower triangular would be ragged. I'm still not sure that that means a `device` keyword needs to be added though. It seems like one would almost always want the data on the same device that the input array is on, and not on the default device.

kgryte:
Re: copy. Agreed. I think that, in most cases, a copy would be required; although, I could imagine, for certain values of `k`, where a sparse view could be used. In order to avoid confusion, we could be explicit in requiring that a copy be returned.

Re: device. We could change the default `device` behavior from `None` (the default device) to `x.device`, so that the device is inherited from the input array. I do think, however, that `tril` and `triu` should support the `device` keyword for consistency with the other array creation APIs.

My sense is that a similar argument for the default device being `x.device` could also apply to the `*_like` creation functions, where you'd want to also inherit the device of the input array; however, the current default behavior is to use the default device.

rgommers:
> Re: copy. Agreed. I think that, in most cases, a copy would be required; although, I could imagine, for certain values of `k`, where a sparse view could be used. In order to avoid confusion, we could be explicit in requiring that a copy be returned.

We don't do that anywhere else I believe (unless there's an explicit `copy=` keyword), and something like `y = triu(x)[0, :]` could fairly easily be optimized to be a view by any library that has a JIT compiler. So I'd suggest to not do that.

> Re: device. We could change the default `device` behavior from `None` (the default device) to `x.device`, so that the device is inherited from the input array. I do think, however, that `tril` and `triu` should support the `device` keyword for consistency with the other array creation APIs.

Hmm, isn't that a bit circular? I'm not quite seeing it as a creation function, and the most similar function (the one that also takes array input) in https://data-apis.org/array-api/latest/API_specification/creation_functions.html is `meshgrid`, which also does not have a `device` keyword. My preference would be for any function that takes array input to not have a `device` keyword.

> My sense is that a similar argument for the default device being `x.device` could also apply to the `*_like` creation functions, where you'd want to also inherit the device of the input array; however, the current default behavior is to use the default device.

Did we do that on purpose? E.g. https://pytorch.org/docs/stable/generated/torch.ones_like.html uses `x.device`.

kgryte:
As discussed during the consortium meeting held 09-02-2021, updated this PR to remove the `device` keyword argument. If users wish to allocate an upper or lower triangular array on a different device, they can use the `.to_device` method specified in #171.

kgryte:
As review comments have now been addressed, will merge. If further refinements are necessary, we can do so in a follow-up PR.

Issue #228, opened 2021-07-14
The `transpose` function is problematic


rgommers:
There are multiple problems, both related to adoption due to conflicting definitions, and because NumPy's `transpose` does something no one really wants (namely, swap _all_ axes).

The current array API `transpose` may not be correct:
- https://data-apis.org/array-api/latest/API_specification/linear_algebra_functions.html#transpose-x-axes-none says "swap all axes"
- https://github.com/data-apis/array-api/issues/147 claims the intent is to only swap the last two axes, to be consistent with all other `linalg` functions.

PyTorch names the `numpy.transpose` equivalent `permute`, and PyTorch's `transpose` is `np.swapaxes`. Relevant issues PyTorch issues:
  - https://github.com/pytorch/pytorch/issues/50275
  - https://github.com/pytorch/pytorch/issues/51280

Relevant NumPy issues:
- https://github.com/numpy/numpy/issues/9530
- https://github.com/numpy/numpy/issues/7495
- https://github.com/numpy/numpy/issues/13797 (note that there's `np.linalg.linalg.transpose` which is private but _does_ have the desired behavior)

There's multiple ways of dealing with this problem:
1. Accept the status quo
2. Remove `transpose` completely, and use `permute` or some other function name with the behavior we want by default (do the right thing for batches of 2-D matrices)
3. Add both, by adding a `matrix_transpose` function and a corresponding `.mT` (or `.MT`) attribute

I searched the JAX and TensorFlow issue trackers as well and, somewhat surprisingly, didn't find an issue about this topic.

shoyer:
I think separating the functionality of `np.transpose()` into two separate functions/methods would make sense. As you note, there are two distinct use cases:

1. Swap the last two dimensions
2. Permute axes

The only real relation is that (1) can be implemented via (2).

From my perspective, the functionality (1) would be the best fit for the name "transpose", and (2) could use some other name, like "permute axes".

kgryte:
During the consortium meeting on 2021-07-14, the following was considered a path forward:

1.  Limit the `.T` attribute on array instances to 2D. If an array instance has more than two dimensions, an error should be raised. This allows retaining the current behavior for `.T` for the primary use case of taking the transpose of a matrix, while removing the less desired behavior of reversing all axes.
2.  Introduce a `.mT` attribute on array instances which performs a batched matrix transpose for the last two dimensions.
3.  Remove the `transpose` function from the specification and replace with a `permute` API.
[...]

PR #245, opened 2021-08-30
Restrict the `T` attribute to two-dimensional arrays

kgryte:
This PR:

-   restricts the `T` attribute to two-dimensional arrays in agreement with the mathematical definition of the transpose.  For array instances which are not two-dimensional, the specification recommends an error be thrown.
-   includes a note explaining the rationale for limiting to two-dimensions and adds a recommendation to use the `permute` API (**note**: the `permute` specification will be added in a subsequent PR; see [gh-247](https://github.com/data-apis/array-api/pull/247)).
-   removes unnecessary blank lines found in various notes.

## Background

Discussion as to why the current behavior of `T` attribute is problematic can be found in [gh-228](https://github.com/data-apis/array-api/issues/228).

By limiting the attribute to two-dimensional arrays, we retain the current behavior for `.T` for the primary use case of taking the transpose of a matrix, while removing the less desired behavior of reversing all axes.

### DIFF
--- a/spec/API_specification/array_object.md
+++ b/spec/API_specification/array_object.md
@@ -261,7 +261,7 @@ Transpose of the array.
 The array instance must be two-dimensional. If the array instance is not two-dimensional, an error should be raised.
 
 ```{note}
-Limiting the transpose to two-dimensional arrays (matrices) deviates from the NumPy et al practice of reversing all axes for arrays having more than two-dimensions. This is intentional, as reversing all axes was found to be problematic (e.g., conflicting with the mathematical definition of a transpose which is limited to matrices; not operating on batches of matrices; et cetera). In order to reverse all axes, one is recommended to use the functional `permute_axes` interface found in this specification.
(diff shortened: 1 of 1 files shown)


rgommers:
> The array instance must be two-dimensional.

(making the same comment twice today, xref https://github.com/pytorch/pytorch/pull/64180): gh-228 seemed to say to raise an error for >2-D only. Cc @Lezcano

kgryte:
Yeah, I am a bit conflicted here. Yes, taking the transpose of a 1D array could be a no-op, but, as discussed in https://github.com/numpy/numpy/issues/7495, it is questionable whether this is actually what you want. In which case, opted to be more restrictive in limiting to only 2D arrays.

kgryte:
Based on discussions in the consortium meeting held 09-02-2021, no objections were made to limiting to 2-dimensional arrays. 

While consensus determined that the transpose of a 1-dimensional array is ambiguous (does a user expect a no-op or actually want to convert a row vector to a column vector or vice versa?), less ambiguous is the behavior of taking the transpose of a 0-dimensional array (a no-op).

However, the stance of this PR is that a user should be explicit in taking transpose of an empty two-dimensional array, rather than using a zero-dimensional shorthand, as not clear to a reader whether an author actually intended for a no-op or this is a bug. Accordingly, strictly limiting to 2D and not allowing 0D avoids this runtime ambiguity.

kgryte:
Merging based on discussions during the 09-02-2021 consortium meeting.

PR #246, opened 2021-08-30
Add specification for performing a batched matrix transpose on an array instance via an `.mT` attribute

kgryte:
This PR:

-   adds a specification for the `.mT` attribute for performing a batched matrix transpose on an array instance.

## Background

Discussion regarding the need for an `.mT` attribute can be found in [gh-228](https://github.com/data-apis/array-api/issues/228).

Notably, we cannot use the existing `.T` attribute for applying a transpose operation to a stack of matrices, as this would introduce an undesirable incompatibility in the array library ecosystem, where existing practice (see NumPy et al) is to reverse all dimensions for arrays having more than two-dimensions. Accordingly, in order to support applying a batched matrix transpose on an array instance, a new attribute is necessary.

### DIFF
--- a/spec/API_specification/array_object.md
+++ b/spec/API_specification/array_object.md
@@ -215,6 +215,19 @@ Hardware device the array data resides on.
 
     -   a `device` object (see {ref}`device-support`).
 
+(attribute-mT)=
+### mT
+
+Transpose of a matrix (or a stack of matrices).
+
+If an array instance has fewer than two dimensions, an error should be raised.
+
+#### Returns
+
+-   **out**: _&lt;array&gt;_
+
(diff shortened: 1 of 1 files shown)


kgryte:
Adding an `.mT` attribute has been discussed in consortium meetings (most recently on 09-02-2021). During these meetings, no objections were raised against the addition of this attribute, as `.mT` achieves the desirable behavior of a batched matrix transpose for arrays having more than two-dimensions.

As the `.T` attribute is now restricted to 2D (see [gh-245](https://github.com/data-apis/array-api/pull/245)), `.mT` extends the transpose to arrays having more than two dimensions via a batched matrix transpose without breaking backward compatibility (i.e., reversing axes).

Further, the `.mT` attribute is complementary to the `matrix_transpose` API (added in [gh-248](https://github.com/data-apis/array-api/pull/248)).

Merging...

PR #248, opened 2021-08-30
Add specification for computing the matrix transpose (`matrix_transpose`)

kgryte:
This PR:

-   adds a specification for computing the transpose of a matrix (or a stack of matrices).

## Background

The addition of a `matrix_transpose` API is meant to replace the use of the `transpose` API. The `transpose` API was found to be problematic due to its API supporting more general axes permutation (see [gh-228](https://github.com/data-apis/array-api/issues/228)).

Similar to other linear algebra APIs, the proposed `matrix_transpose` API operates on stack of matrices and thus more succinctly and explicitly allows expressing matrix operations in accordance with mathematical intuition and obviates the need for doing `transpose(x, axes=[-2,-1])` whenever wanting to perform a batched matrix transpose. Batching is the default behavior.

The `matrix_transpose` API serves as a functional complement to the proposed [`.mT`](https://github.com/data-apis/array-api/pull/246) attribute (see [gh-246](https://github.com/data-apis/array-api/pull/246)).


kgryte:
Adding the `matrix_transpose` was discussed in the consortium meeting held 09-02-2021. No objections were raised, as this PR provides the desired behavior of batched matrix transpose.

Merging...

PR #251, opened 2021-09-01
Clarify some text in the logical_* functions

asmeurer:
Only boolean data types are required in these functions, so the zero-nonzero
bool equivalence is only for implementations that choose to implement
non-boolean inputs.

Alternatively, we could just remove these sentences from here.

### DIFF
--- a/spec/API_specification/elementwise_functions.md
+++ b/spec/API_specification/elementwise_functions.md
@@ -968,7 +968,11 @@ For floating-point operands,
 (function-logical_and)=
 ### logical_and(x1, x2, /)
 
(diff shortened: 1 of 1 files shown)


spec/API_specification/elementwise_functions.md
```diff
@@ -968,7 +968,7 @@ For floating-point operands,
 (function-logical_and)=
 ### logical_and(x1, x2, /)
 
-Computes the logical AND for each element `x1_i` of the input array `x1` with the respective element `x2_i` of the input array `x2`. Zeros are considered the equivalent of `False`, while non-zeros are considered the equivalent of `True`.
+Computes the logical AND for each element `x1_i` of the input array `x1` with the respective element `x2_i` of the input array `x2`. Only boolean data types are required, but if non-boolean data types are implemented, zeros are considered the equivalent of `False`, while non-zeros are considered the equivalent of `True`.
```
kgryte, line 971: May be better to move this sentence to a note, rather than as the main description of the API description.

kgryte (approved):
LGTM.

PR #253, opened 2021-09-03
Fix output array dtype when computing the remainder

asmeurer:
It stated that the return type should be floating-point, but it accepts
integer inputs (and the remainder of two integers is always an integer).

Somewhat related, I noticed that `remainder` doesn't specify any special cases, but it perhaps should specify the behavior for remainder by 0, and infinites. Also should integer remainder by 0 be unspecified?


kgryte (approved):
LGTM. Thanks for catching this!

kgryte:
Regarding the special cases mentioned in the OP, we should open a separate issue.

PR #258, opened 2021-09-10
Fix wrong function name in array_object.py

soraros:


oleksandr-pavlyk:
Perhaps the link should be changed to point to `positive`. The `positive` function is simpler than `bitwise_or`.

kgryte (approved):
LGTM. Thanks for catching this!

PR #259, opened 2021-09-13
Address a few device related issues

### LINKED ISSUE #156: Device object and device `kwarg` for creation
Hi,

When reading the section regarding [devices](https://data-apis.org/array-api/latest/design_topics/device_support.html), I stumbled upon the following note:

> The only way to obtain a Device object is from the .device property on the array object, hence there is no Device object in the array API itself that can be instantiated to point to a specific physical or logical device.

I understood this, as users can never instantiate a `Device` object themselves.
But in the paragraph inmediately above there are sentences like:

> A device=None keyword for array creation functions, which takes an instance of a Device object.

> A .to_device(device) method on the array object, with device again being a Device object, to move an array to a different device.

And both require a `Device` object, that according to the note, can't be directly instantiated, or be obtained from elsewhere than a created array.

### LINKED ISSUE #256: The spec of `to_device()` method is missing
Here we mentioned there's a `to_device()` method to enable data copy/movement across devices:
https://github.com/data-apis/array-api/blob/09410675703ee275e6e5746db3aded0d37bc4b96/spec/design_topics/device_support.md#L48-L49
But we haven't defined the signature and semantics yet.

leofang:
Close #256. Close #156.

This PR addresses a few things:
- Add `stream` to `.to_device()` and allow any library-specific stream object to be used
- Clarify that `.to_device()` is to *copy* (not move) data across devices 
- Clarify that the standard is not concerned with (a)synchronous transfer
- Clarify that any library-specific device management is permitted

### DIFF
--- a/spec/API_specification/array_object.md
+++ b/spec/API_specification/array_object.md
@@ -461,6 +461,7 @@ Exports the array for consumption by {ref}`function-from_dlpack` as a DLPack cap
         ```{note}
         Support for a `stream` value other than `None` is optional and implementation-dependent.
         ```
+
         Device-specific notes:
 
         :::{admonition} CUDA
--- a/spec/API_specification/array_object.md
+++ b/spec/API_specification/array_object.md
@@ -1191,3 +1191,33 @@ Evaluates `self_i ^ other_i` for each element of an array instance with the resp
 
 Element-wise results must equal the results returned by the equivalent element-wise function [`bitwise_xor(x1, x2)`](elementwise_functions.md#bitwise_xorx1-x2-).
 ```
+
+(method-to_device)=
+### to_device(self, device, /, *, stream=None)
+
+Copy the array from the device it currently resides to the specified `device`.
+
+#### Parameters
+
+-   **self**: _&lt;array&gt;_
+
+    -   array instance.
+
+-   **device**: _&lt;device&gt;_
+
+    -   device to place the copied array on.
+
+-   **stream**: _Optional\[ Union\[ int, Any ]]_
+
+    -   stream object to use during copy. See {ref}`method-__dlpack__`.
+
+#### Returns
+
+-   **out**: _&lt;array&gt;_
+
+    -   an identical copy of the source array placed on the target `device`.
+
+```{note}
+
+Whether the copy is performed synchronously or asynchronously is up to the array library. As a result, if any synchronization (which is out of scope of this standard) is required to guarantee data safety, the library should explain to its users.
+```
--- a/spec/API_specification/array_object.md
+++ b/spec/API_specification/array_object.md
@@ -1217,7 +1217,7 @@ Element-wise results must equal the results returned by the equivalent element-w
 (method-to_device)=
 ### to\_device(self, device, /, *, stream=None)
 
-Copy the array from the device it currently resides to the specified `device`.
+Copy the array from the device on which it currently resides to the specified `device`.
 
 #### Parameters
 
@@ -1227,7 +1227,7 @@ Copy the array from the device it currently resides to the specified `device`.
 
 -   **device**: _&lt;device&gt;_
 
-    -   a `device` object (see {ref}`device-support`)
+    -   a `device` object (see {ref}`device-support`).
 
 -   **stream**: _Optional\[ Union\[ int, Any ]]_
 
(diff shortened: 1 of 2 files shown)


leofang:
Let's merge #171 first.

kgryte:
@leofang Following up here: are there things related to this PR that we should discuss and resolve during the next consortium meeting?

spec/design_topics/device_support.md
```diff
@@ -48,13 +48,20 @@ cross-device data transfer:
...
 
 ```{note}
-The only way to obtain a `Device` object is from the `.device` property on
-the array object, hence there is no `Device` object in the array API itself
-that can be instantiated to point to a specific physical or logical device.
+In the current API standard, the only way to obtain a `Device` object is from the
+`.device` property on the array object. The standard does **not** include a universal
+`Device` object recognized by all compliant libraries. Accordingly, the standard does
+not provide a means of instantiating a `Device` object to point to a specific physical or
+logical device.
```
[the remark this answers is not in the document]
kgryte: I think we can add a note here stating that this may be revisited in the future.

spec/API_specification/array_object.md
```diff
@@ -458,6 +458,8 @@ Exports the array for consumption by {ref}`function-from_dlpack` as a DLPack cap
 
         For other device types which do have a stream, queue or similar synchronization mechanism, the most appropriate type to use for `stream` is not yet determined. E.g., for SYCL one may want to use an object containing an in-order `cl::sycl::queue`. This is allowed when libraries agree on such a convention, and may be standardized in a future version of this API standard.
 
+        Support for any `stream` value other than `None` is optional and up to each library in question.
+
```
kgryte, line 462:
```suggestion
        ```{note}
        Support for a `stream` value other than `None` is optional and implementation-dependent.
        ```
```

```

File: spec/API_specification/array_object.md [hunk 3]
```diff
@@ -1227,8 +1229,16 @@ Move the array to the given device.
 
     -   a `device` object (see {ref}`device-support`).
 
+-   **stream**: _Optional\[ Union\[ int, Any ]]_
+
+    -   stream object to use during copy. In addition to the supported types as discussed in {ref}`method-__dlpack__`, any library-specific stream object is also allowed to be used here, with the caveat that using such an object would make the code non-portable.
```
kgryte, line 1234:
```suggestion
    -   stream object to use during copy. In addition to the types supported in {ref}`method-__dlpack__`, implementations may choose to support any library-specific stream object with the caveat that any code using such an object would not be portable.
```

spec/API_specification/array_object.md
```diff
@@ -1227,8 +1229,16 @@ Move the array to the given device.
...
+
 #### Returns
 
 -   **out**: _&lt;array&gt;_
 
-    -   an array with the same data and dtype, located on the specified device.
+    -   an array with the same data and data type as `self` and located on the specified `device`.
+
+```{note}
+If `stream` is given, the copy operation should be enqueued on the provided `stream`; otherwise, the copy operation should be enqueued on the default stream/queue. Whether the copy is performed synchronously or asynchronously is library dependent. As a result, if synchronization is required to guarantee data safety, this should be clearly explained in a conforming library's documentation.
```
kgryte, line 1243:
```suggestion
If `stream` is given, the copy operation should be enqueued on the provided `stream`; otherwise, the copy operation should be enqueued on the default stream/queue. Whether the copy is performed synchronously or asynchronously is implementation-dependent. As a result, if synchronization is required to guarantee data safety, this should be clearly explained in a conforming library's documentation.
```

spec/design_topics/device_support.md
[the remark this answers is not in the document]
kgryte:
```suggestion
not provide a means of instantiating a `Device` object to point to a specific physical or
logical device.

The choice to not include a standardized `Device` object may be revisited in a future revision of this standard.
```

spec/API_specification/array_object.md
```diff
@@ -458,6 +458,9 @@ Exports the array for consumption by {ref}`function-from_dlpack` as a DLPack cap
 
         For other device types which do have a stream, queue or similar synchronization mechanism, the most appropriate type to use for `stream` is not yet determined. E.g., for SYCL one may want to use an object containing an in-order `cl::sycl::queue`. This is allowed when libraries agree on such a convention, and may be standardized in a future version of this API standard.
 
+        ```{note}
+        Support for a `stream` value other than `None` is optional and implementation-dependent.
+        ```
```
kgryte, line 463:
```suggestion
        ```

```

kgryte (approved):
LGTM. Any additional changes can be made in follow-up PRs. Thanks, @leofang!

PR #260, opened 2021-09-13
Clarify output dtype guidance for `mean`, `var`, and `std`

kgryte:
This PR

-   clarifies output dtype guidance for `mean`, `var`, and `std`, as discussed in [gh-202](https://github.com/data-apis/array-api/issues/202#issuecomment-863405754).

    Notably, the PR adds notes which clarify the output array dtype when the input array has an integer dtype. While integer dtypes are not explicitly supported by the standard (due to mixed-type promotion being undefined), this PR adds language to require that, if an array having an integer dtype is provided, the returned array must have the default floating-point dtype.

    For floating-point input arrays, the returned array must have the same dtype as the input array.

### DIFF
--- a/spec/API_specification/statistical_functions.md
+++ b/spec/API_specification/statistical_functions.md
@@ -63,7 +63,9 @@ Calculates the arithmetic mean of the input array `x`.
 
 -   **out**: _&lt;array&gt;_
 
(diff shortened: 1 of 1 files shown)


rgommers (approved):
LGTM and should be uncontroversial, so merging. Thanks @kgryte

Issue #197, opened 2021-06-10
Undefined behavior for overflowing `.full()`


pmeier:
Currently the specification stipulates nothing about what should happen in case [`full`](https://data-apis.org/array-api/latest/API_specification/creation_functions.html#full-shape-fill-value-dtype-none-device-none) (or `full_like`) get passed a `fill_value` that overflows the `dtype`. For example numpy emits a `RuntimeWarning` and uses `inf` as a substitute 

```python
>>> np.full((), np.finfo(np.float64).max * 2)
<input>:1: RuntimeWarning: overflow encountered in double_scalars
array(inf)
```

while PyTorch bails out:

```python
>>> torch.full((), torch.finfo(torch.float32).max * 2)
RuntimeError: value cannot be converted to type float without overflow: 6.80565e+38
```

I hit this discrepancy while running the API test suite against PyTorch. It uses the [`hypothesis`' `floats` strategy](https://hypothesis.readthedocs.io/en/latest/data.html?highlight=floats#hypothesis.strategies.floats), which includes values larger than PyTorch's default dtype (float32) can handle, together with `full` to [generate numeric scalars](https://github.com/data-apis/array-api-tests/blob/6e3684b7948564b3ecd69f5b066aaf9c722120ec/array_api_tests/hypothesis_helpers.py#L104). cc @asmeurer 

Thus, we should either specify the behavior or fix the test suite to not rely on a specific behavior. I vote for the former.

asmeurer:
Probably should be unspecified (meaning the test suite needs to be fixed).

rgommers:
Agreed with Aaron. In general we try not to specify detailed warning and exception behavior, not for things like floating-point overflows (it's hardware and compiler-dependent often) nor for things like invalid inputs to functions. There's a couple of reasons for that: it's not practical to do, libraries can be opinionated here, or even offer APIs to change this behavior (e.g., `numpy.errstate` context manager)

PR #261, opened 2021-09-13
Clarify `full` and `full_like` behavior when fill value exceeds precision

kgryte:
This PR

-   resolves [gh-197](https://github.com/data-apis/array-api/issues/197).
-   clarifies `full` and `full_like` behavior when the `fill_value` exceeds the precision of the default integer/floating-point data type. Notably, this PR specifies that behavior is left unspecified and, thus, implementation-defined. (see also [comment](https://github.com/data-apis/array-api/issues/197#issuecomment-860038261))


rgommers (approved):
LGTM, thanks @kgryte

Issue #231, opened 2021-07-16
__index__ method


asmeurer:
I just noticed that the `__index__` method is missing from the array object specification. See https://www.python.org/dev/peps/pep-0357/.

`__index__` should return `int(self)` if the input is a 0-D array with integer dtype, and raise TypeError otherwise (similar to `__int__` except it doesn't work on floating-point arrays). This will allow 0-D integer arrays to be used as indices. While this is not strictly necessary for an array type to support itself as an index, it does mean that things like indexing Python lists with 0-D arrays will work correctly, and it's generally a good Python practice to do this for integer-like types.

rgommers:
I just did a quick check that NumPy, PyTorch, JAX and TensorFlow all support this behavior. So seems reasonable. Size-1 lists don't support it, so it's not obvious that it should work. But may as well, since it's common practice.

PR #263, opened 2021-09-13
Add `__index__` to array object

kgryte:
This PR

-   adds `__index__` to the array object, which is called to implement [`operator.index()`](https://docs.python.org/3/reference/datamodel.html#object.__index__) (see also [PEP 0357](https://www.python.org/dev/peps/pep-0357/)).
-   allows a zero-dimensional array to be used as an index when indexing lists and other built-in Python objects supporting slices.
-   resolves [gh-231](https://github.com/data-apis/array-api/issues/231).


rgommers (approved):
LGTM, thanks @kgryte

PR #262, opened 2021-09-13
Specify behavior of statistical reductions over the empty set

kgryte:
This PR

-   specifies the behavior of statistical reductions when the number of elements over which to compute a reduction is zero and resolves [gh-232](https://github.com/data-apis/array-api/issues/232).
-   Notably, this PR proposes

     -   `max`: implementation-defined (due to no consistent maximum value across all supported data types).
     -   `mean`: return `NaN`.
     -   `min`: implementation-defined (due to no consistent minimum value across all supported data types).
     -   `prod`: return `1` (empty product).
     -   `std`: return `NaN`.
     -   `sum`: return `0` (empty sum).
     -   `var`: return `NaN`.

PR #265, opened 2021-09-15
Revert "Specify behavior of statistical reductions over the empty set"

Psoffie009:
Reverts data-apis/array-api#262

### DIFF
--- a/spec/API_specification/statistical_functions.md
+++ b/spec/API_specification/statistical_functions.md
@@ -20,10 +20,6 @@ A conforming implementation of the array API standard must provide and support t
 
 Calculates the maximum value of the input array `x`.
 
-```{note}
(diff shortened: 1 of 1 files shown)


kgryte:
@Psoffie009 Perhaps you can explain a bit more the rationale behind this PR.

kgryte:
This appears to be a bot.

asmeurer:
The same bot has also left some spammy comments on other issues. I would suggest blocking and reporting them to GitHub. 

There seems to be a strange influx of unusual bots recently. Another bot on NumPy a few weeks ago exactly copied one of @honno's issues word for word. I'm not really sure what the point of such bots is. My best guess is that they are trying to create a bunch of innocuous comments on GitHub which they can later edit to include SEO spam.

rgommers:
> The same bot has also left some spammy comments on other issues. I would suggest blocking and reporting them to GitHub.

done, and will delete the other comments.

PR #268, opened 2021-09-20
Update device guidance for `*_like` functions

kgryte:
This PR

-   updates the output array device guidance for `*_like` creation functions to infer the output array device from the input array, if the `device` kwarg is `None`. This behavior is consistent with `dtype` and `shape` inference and consistent with, e.g., [Torch](https://pytorch.org/docs/1.9.0/generated/torch.ones_like.html?highlight=ones_like#torch.ones_like).

This change was discussed and agreed upon in the 2021-09-16 consortium meeting.


kgryte:
As this PR was discussed and agreed upon during the 2021-09-16 consortium meeting, merging...

PR #269, opened 2021-09-20
Remove `einsum`

kgryte:
This PR

-   removes the `einsum` TODO. The addition of `einsum` is delayed until the 2022 specification revision.


kgryte:
This change was discussed and agreed upon during the consortium meeting held on 2021-09-09.

PR #250, opened 2021-09-01
Indicate that abs() of the minimum integer value for signed dtypes is undefined

asmeurer:

PR #270, opened 2021-09-20
Add note to `negative` for signed integer dtype behavior

kgryte:
This PR

-   adds a note to `negative` concerning the behavior of computing the numerical negative of the minimum representable integer value for arrays having a signed integer data type. Namely, that behavior is implementation-dependent. This is similar to [gh-250](https://github.com/data-apis/array-api/pull/250).


kgryte:
This was discussed in [gh-250](https://github.com/data-apis/array-api/pull/250).

PR #274, opened 2021-09-27
Update guidance for output array dtype inference in `full_like`

kgryte:
This PR

-   updates the guidance for output array dtype inference in `full_like`.

## Discussion

In [gh-167](https://github.com/data-apis/array-api/commit/e6d827afdbeb8db0f18288526231aa1904e63a5c#diff-f8b551edf45aca0f0a5926e0a6f4408c45f7a43ce0773d42179f5a16c1a99df9R238), guidance was changed (see [gh-31](https://github.com/data-apis/array-api/pull/31/files#diff-f8b551edf45aca0f0a5926e0a6f4408c45f7a43ce0773d42179f5a16c1a99df9R150)) to infer the dtype from the `fill_value`. However, this seems to have been a mistake, as scalar values do not have an array data type (only `int` or `float`) and do not participate in type promotion.

This PR reverts this change, but also includes an explicit note that, if the `fill_value` is not of the same data type "kind" as the resolved output array data type, then behavior is left unspecified (e.g., if `x.dtype` is `int64` and `fill_value` is `float`; e.g., `fill_value = 3.5`).

### DIFF
--- a/spec/API_specification/creation_functions.md
+++ b/spec/API_specification/creation_functions.md
@@ -239,10 +239,14 @@ Returns a new array filled with `fill_value` and having the same `shape` as an i
 
 -   **dtype**: _Optional\[ &lt;dtype&gt; ]_
 
-    -   output array data type. If `dtype` is `None`, the output array data type must be inferred from `fill_value` (see {ref}`function-full`). Default: `None`.
+    -   output array data type. If `dtype` is `None`, the output array data type must be inferred from `x`. Default: `None`.
 
         ```{note}
-        If `dtype` is `None` and the `fill_value` exceeds the precision of the resolved default output array data type, behavior is left unspecified and, thus, implementation-defined.
+        If `dtype` is `None` and the `fill_value` exceeds the precision of the resolved output array data type, behavior is unspecified and, thus, implementation-defined.
+        ```
+
+        ```{note}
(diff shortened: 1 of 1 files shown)


asmeurer:
CC @honno

asmeurer:
> However, this seems to have been a mistake, as scalar values do not have an array data type (only int or float) and do not participate in type promotion.

FWIW, this matched the behavior of `full()`, which infers the data type from `fill_value` when `dtype=None`.

kgryte:
@asmeurer Correct, but `full_like` has the advantage of additional information needed to make a better guess as to the desired output array data type. So `full` should probably not serve as a basis for `full_like`.

asmeurer:
Also, just to clarify, for all the `*_like` functions, the dtype of `x` should be completely ignored if `dtype` is provided, even if it isn't compatible? For instance, `ones_like(array(..., dtype=bool), dtype=int8)` (this was another point that came up in https://github.com/data-apis/array-api-tests/pull/18).

asmeurer (approved):
I've mentioned before that I never quite understood the point of the `*_like` functions, but it does seem reasonable to me that they should try to make an array as "like" `x` as possible. And more to the point, this now matches the NumPy behavior. So this looks good to me.

kgryte:
> Also, just to clarify, for all the *_like functions, the dtype of x should be completely ignored if dtype is provided, even if it isn't compatible?

Yes, it should be ignored. Compatibility does not apply for the `*_like` functions, as the functions infer meta data about the array and do nothing with the actual values of `x` themselves.

I will update the guidance, however, to make this a bit more explicit.

kgryte:
This has been approved and simply reverts to previous behavior. Merging...

Issue #212, opened 2021-06-30
Address polymorphic return value for `unique`?


rgommers:
`unique` was added very early on in gh-25. The review discussions have focused on

- whether return values should be sorted or not (see gh-40), and
- (b) the issue with value-dependent output shape (see gh-84).

There is another issue that was not discussed: `unique` is the only function whose return type is polymorphic. It can return an array, or a length-2, 3 or 4 tuple of arrays:

```python
unique(x, /, *, return_counts=False, return_index=False, return_inverse=False)
```
API doc: https://data-apis.org/array-api/latest/API_specification/set_functions.html#unique-x-return-counts-false-return-index-false-return-inverse-false

This issue is painful for, e.g., libraries that have to support it in their JIT compiler. For `svd`, which previously was polymorphic too, we decided that it was better to introduce `svdvals` to get rid of the polymorphism. There's even a design principle that says as much in https://data-apis.org/array-api/latest/extensions/linear_algebra_functions.html: _"In general, interfaces should avoid polymorphic return values (e.g., returning an array or a namedtuple, dependent on, e.g., an optional keyword argument)."_

Given that there are 3 boolean keywords, it's not feasible to split `unique` into separate functions (there'd be 8 of them). An alternative may be:
- return a custom object with four named fields: (`unique`, `indices`, `inverse`, `counts`).
- no tuple unpacking
- if data is returned, its type is array
- for fields where the boolean keyword was `False`, accessing the field is undefined behavior

Another question that may be relevant is: how often are the boolean keywords each used? The data at https://raw.githubusercontent.com/data-apis/python-record-api/master/data/typing/numpy.py may help there:

<details>

```python
@overload
def unique(ar: numpy.ndarray, return_inverse: bool):
    """
    usage.scipy: 14
    usage.skimage: 5
    usage.sklearn: 125
    usage.statsmodels: 18
    usage.xarray: 3
    """
    ...


@overload
def unique(ar: numpy.ndarray):
    """
    usage.dask: 7
    usage.matplotlib: 6
    usage.orange3: 16
    usage.scipy: 39
    usage.seaborn: 5
    usage.skimage: 48
    usage.sklearn: 385
    usage.statsmodels: 45
    usage.xarray: 10
    """
    ...


@overload
def unique(ar: numpy.ndarray, return_inverse: bool, return_counts: bool):
    """
    usage.skimage: 3
    usage.sklearn: 2
    usage.statsmodels: 1
    """
    ...


@overload
[...]

PR #275, opened 2021-09-27
Split `unique` into separate functions

kgryte:
This PR

-   resolves [gh-212](https://github.com/data-apis/array-api/issues/212) by splitting `unique` into separate functions with a fixed number of arguments, thus addressing the concerns regarding `unique`'s polymorphic behavior.

-   splits `unique` in the following functions:

    -   `unique(x, /)`: returns (1) unique values, (2) indices of first occurrence, (3) reverse indices, and (4) counts. This is the equivalent of the previous API with `return_index`, `return_inverse`, and `return_counts` kwargs as `True`.
    -   `unique_inverse(x,/)`: returns (1) unique values and (2) reverse indices. As discussed in the [OP](https://github.com/data-apis/array-api/issues/212#issue-934066289) for [gh-212](https://github.com/data-apis/array-api/issues/212), this addresses the fairly common use case of setting `return_inverse` to `True` in the previous API.
    -    `unique_values(x,/)`: only returns (1) unique values. This is the most common use of the previous API.

-   chose to retain `return_index` as a return value in `unique` as this returned array does have value in being able to construct the array of unique values and because deriving this returned array is somewhat tedious to construct via the objects in this specification otherwise.

-   eliminates polymorphic return behavior by removing kwargs from the respective APIs which affect which and how many arrays are returned.

-   adds guidance on expected output array shapes.

-   requires that named tuples be returned.

### DIFF
--- a/spec/API_specification/set_functions.md
+++ b/spec/API_specification/set_functions.md
@@ -13,12 +13,12 @@ A conforming implementation of the array API standard must provide and support t
 <!-- NOTE: please keep the functions in alphabetical order -->
 
 (function-unique)=
-### unique(x, /, *, return_counts=False, return_index=False, return_inverse=False)
+### unique(x, /)
 
 :::{admonition} Data-dependent output shape
 :class: important
 
-The shapes of one or more of output arrays for this function depend on the data values in the input array; hence, array libraries which build computation graphs (e.g., JAX, Dask, etc.) may find this function difficult to implement without knowing array values. Accordingly, such libraries may choose to omit this function. See {ref}`data-dependent-output-shapes` section for more details.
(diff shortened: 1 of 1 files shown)


rgommers:
I suggest not mixing the "are NaN's unique" question with the discussion on `unique` being polymorphic and hence this PR to split it into three separate functions. They are orthogonal problems. Could you move your question to the NaN-unique issue @honno?

kgryte:
This PR was discussed during the 7 October 2021 consortium meeting. No objections were raised and the proposed changes were seen as positive. Will merge. Any needed changes can be addressed in subsequent PRs.

PR #276, opened 2021-09-27
Remove the `Sequence` encapsulation of variadic parameters

BvB93:
Xref https://github.com/data-apis/array-api/issues/143 and https://github.com/numpy/numpy/pull/19969.

Variadic parameters were previously all encapsulated in a `Sequence` type, implying that each individual parameter must be a `Sequence` (rather than the encapsulated type).

Examples
---------
``` python
from typing import List

def func1(*args: List[int]) -> None: ...
func1([1], [2], [3])

def func2(*args: int) -> None: ...
func2(1, 2, 3)
```

### DIFF
--- a/spec/API_specification/creation_functions.md
+++ b/spec/API_specification/creation_functions.md
@@ -304,7 +304,7 @@ Returns coordinate matrices from coordinate vectors.
 
 #### Parameters
 
--    **arrays**: _Sequence\[ &lt;array&gt; ]_
+-    **arrays**: _&lt;array&gt;_
 
      -   one-dimensional arrays representing grid coordinates. Must have a numeric data type.
 
--- a/spec/API_specification/creation_functions.md
+++ b/spec/API_specification/creation_functions.md
@@ -306,7 +306,7 @@ Returns coordinate matrices from coordinate vectors.
 
 -    **arrays**: _&lt;array&gt;_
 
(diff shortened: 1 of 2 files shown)


rgommers (commented):
Interesting, that looks correct - but also counterintuitive, the annotation is a bit unhelpful as documentation.

It may be useful to then modify the sentences below each annotation to say "Sequence of ... arrays ..." for clarity.

BvB93:
> It may be useful to then modify the sentences below each annotation to say "Sequence of ... arrays ..." for clarity.

I'm personally not huge fan of the term `Sequence` in this context. 
The term `Sequence` works fine-ish if, for example, the input is passed via unpacking a list of arrays (`out = func(*ar_list)`), 
but makes less sense if the arrays are explicitly passed as individual arguments (`out = func(ar1, ar2, ar3)`).

How about something along the lines of "variable" or "arbitrary" number of arrays? 
I'd say this leaves less room for ambiguity.

rgommers:
> How about something along the lines of "variable" or "arbitrary" number of arrays?
> I'd say this leaves less room for ambiguity.

Yes that sounds great to me.

rgommers (approved):
LGTM now, thanks @BvB93!

asmeurer:
Actually better wording would be "one or more" instead of "an arbitrary number", since none of these functions accept zero arguments, even though the *args syntax would allow it.

PR #277, opened 2021-09-27
Small fixes to the linalg extension

asmeurer:


spec/API_specification/linear_algebra_functions.md
```diff
@@ -2,6 +2,12 @@
 
 > Array API specification for linear algebra functions.
 
+```{note}
```
kgryte, line 5: Is this note necessary given L11 which states that

> A conforming implementation of the array API standard must provide and support the following functions adhering to the following conventions.

This already implies that these linear algebra functions are required.

kgryte (approved):
LGTM apart from what I'd consider a duplicate note.

PR #278, opened 2021-09-28
Disallow `k=None` for the `eye` function

BvB93:
Xref https://github.com/data-apis/array-api/issues/143 and https://github.com/numpy/numpy/pull/19969.

The spec currently claims that the `k` parameter of [`eye`](https://data-apis.org/array-api/latest/API_specification/creation_functions.html#eye-n-rows-n-cols-none-k-0-dtype-none-device-none) should accept either `None` or an integer, however not a single mention is made of `None` in the actual parameter-description. In fact, support for `k=None` seems to be absent from the limes of `np.eye` as well.

Based on this it seems the inclusion of the `k: Optional[int] = ...` annotation may have been accidental; hence the `Optional` encapsulation is removed in this PR. If this was a deliberate choice I'd instead propose to update the docs with a more explicit mention of `None` and its expected behavior.

### DIFF
--- a/spec/API_specification/creation_functions.md
+++ b/spec/API_specification/creation_functions.md
@@ -150,7 +150,7 @@ Returns a two-dimensional array with ones on the `k`th diagonal and zeros elsewh
 
     -   number of columns in the output array. If `None`, the default number of columns in the output array is equal to `n_rows`. Default: `None`.
 
--   **k**: _Optional\[ int ]_
+-   **k**: _int_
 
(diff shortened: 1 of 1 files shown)


asmeurer:
This looks good to me, but others should confirm that this is indeed what was intended.

rgommers (approved):
Thanks @BvB93. This seems like a simple copy-paste mistake, or assuming "optional" means "don't have to specify" rather than "`| None`". This issue wasn't discussed in the original PR (gh-31). In it goes.

Issue #257, opened 2021-09-08
Clarify whether single-axis indices can be tuples i.e. `x[(i,)]`


honno:
Currently `numpy.array_api` supports this behaviour. For example:

```
>>> x = xp.zeros(5)
>>> x[(0,)]
Array(0)
```

Also notably @asmeurer tests for `(i,)` indices in the compliance suite's multiaxis tests.

I can't work out whether the indexing rules specify this is allowed or not (i.e. not explicitly mentioned in either the single or multi-axis sections), but I would imagine this is expected. Before this was mentioned by @Zac-HD in HypothesisWorks/hypothesis#3065, I had just assumed that multi meant "one-or-more" and so this is allowed.

Let me know if I'm wrong and you can guarantee this with the current spec, although that raises the question if we could still purposely note this.

Clarification either way would be nice, but if this isn't actually agreed upon then I'd just add that allowing `(i,)` indices would make index tooling (like in Hypothesis) much simpler.

asmeurer:
I understand that as a multi-axis index that selects 1 axis. Maybe the multi-axis indexing section could say something like "a multi-axis index indexes between 0 and N indices (inclusive)".

kgryte:
The following note in the spec implies that single-element tuples are the equivalent of single-axis indices:

> In Python, `x[(exp1, exp2, ..., expN)]` is equivalent to `x[exp1, exp2, ..., expN]`; the latter is syntactic sugar for the former.

However, I've submitted [gh-280](https://github.com/data-apis/array-api/pull/280) to provide some additional guidance.

PR #280, opened 2021-10-04
Add clarifying examples to note concerning tuple usage in multi-axis indexing

kgryte:
This PR

-   resolves https://github.com/data-apis/array-api/issues/257 by expanding the note regarding tuple usage for multi-axis indexing. Notably, the added examples include a one-element tuple example demonstrating the equivalence with a single-axis index.

### DIFF
--- a/spec/API_specification/indexing.md
+++ b/spec/API_specification/indexing.md
@@ -137,8 +137,6 @@ Multi-dimensional arrays must extend the concept of single-axis indexing to mult
     In Python, `A[(exp1, exp2, ..., expN)]` is equivalent to `A[exp1, exp2, ..., expN]`; the latter is syntactic sugar for the former.
 
     Accordingly, if `A` has rank `1`, then `A[(2:10,)]` must be equivalent to `A[2:10]`. If `A` has rank `2`, then `A[(2:10, :)]` must be equivalent to `A[2:10, :]`. And so on and so forth.
-
-    To access the value in a zero-dimensional array, one must provide an empty tuple; e.g., `A[()]`.
     ```
 
 -   Providing a single nonnegative integer `i` as a single-axis index must index the same elements as the slice `i:i+1`.
--- a/spec/API_specification/indexing.md
+++ b/spec/API_specification/indexing.md
@@ -17,21 +17,18 @@ To index a single array axis, an array must support standard Python indexing rul
 -   **Valid** nonnegative indices must reside on the half-open interval `[0, n)`.
 
     ```{note}
-
     This specification does not require bounds checking. The behavior for out-of-bounds integer indices is left unspecified.
     ```
 
 -   Negative indices must count backward from the last array index, starting from `-1` (i.e., negative-one-based indexing, where `-1` refers to the last array index).
 
     ```{note}
-
     A negative index `j` is equivalent to `n-j`; the former is syntactic sugar for the latter, providing a shorthand for indexing elements that would otherwise need to be specified in terms of the axis (dimension) size.
     ```
 
 -   **Valid** negative indices must reside on the closed interval `[-n, -1]`.
 
     ```{note}
-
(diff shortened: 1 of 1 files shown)


asmeurer (approved):

spec/API_specification/indexing.md
```diff
@@ -134,7 +134,11 @@ Multi-dimensional arrays must extend the concept of single-axis indexing to mult
 -   Each axis may be independently indexed via single-axis indexing by providing a comma-separated sequence ("selection tuple") of single-axis indexing expressions (e.g., `A[:, 2:10, :, 5]`).
 
     ```{note}
-    In Python, `x[(exp1, exp2, ..., expN)]` is equivalent to `x[exp1, exp2, ..., expN]`; the latter is syntactic sugar for the former.
+    In Python, `A[(exp1, exp2, ..., expN)]` is equivalent to `A[exp1, exp2, ..., expN]`; the latter is syntactic sugar for the former.
+
+    Accordingly, if `A` has rank `1`, then `A[(2:10,)]` must be equivalent to `A[2:10]`. If `A` has rank `2`, then `A[(2:10, :)]` must be equivalent to `A[2:10, :]`. And so on and so forth.
+
+    To access the value in a zero-dimensional array, one must provide an empty tuple; e.g., `A[()]`.
```
honno, line 141: I had just assumed the Array API didn't support 0d indexing, so just incase I'll ask—are we sure about allowing empty tuples to index 0d arrays?

asmeurer: This does work, because the empty tuple indexes zero axes. 

However, I don't like this sentence "To access the value in a zero-dimensional array" now that I look at it. In the spec, indexing returns an array, so `a[()]` would just return `a` unchanged (regardless of its shape). "Access the value" seems to imply some sort of NumPy-like scalar objects.

kgryte: Right. Okay. The preferred means of accessing the value in a 0D array would be, e.g., `float(x)` and `int(x)`, correct?

If so, I'll remove that sentence, and add a sentence stating that indexing zero-dimensional arrays is not allowed.

asmeurer: > The preferred means of accessing the value in a 0D array would be, e.g., float(x) and int(x), correct?

If you want a Python scalar yes. The 0-D array itself ought to just work as a scalar value in most relevant contexts inside of the array API. 

> If so, I'll remove that sentence, and add a sentence stating that indexing zero-dimensional arrays is not allowed.

`A[()]` and `A[...]` should be allowed on 0-dimensional arrays, because in both cases the indices index 0 axes.

spec/API_specification/indexing.md
```diff
@@ -143,10 +144,20 @@ Multi-dimensional arrays must extend the concept of single-axis indexing to mult
...
+    ```
+
 -   Providing a slice must retain array dimensions (i.e., the array rank must remain the same; `rank(A) == rank(A[:])`).
 
 -   Providing [ellipsis](https://docs.python.org/3/library/constants.html#Ellipsis) must apply `:` to each dimension necessary to index all dimensions (e.g., if `A` has rank `4`, `A[1:, ..., 2:5] == A[1:, :, :, 2:5]`). Only a single ellipsis must be allowed. An `IndexError` exception must be raised if more than one ellipsis is provided.
 
+-   Providing an empty tuple or an ellipsis to an array of rank `0` must result in an array of the same rank (i.e., if `A` has rank `0`, `A == A[()]` and `A == A[...]`).
+
+    ```{note}
+    This behavior differs from NumPy where providing an empty tuple to an array of rank `0` returns a Python scalar.
```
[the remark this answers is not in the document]
kgryte: I've fixed the text and added a note regarding selections of single values returning zero-dimensional arrays in the section on "single axis indexing".

kgryte:
As this PR does not include any breaking changes and simply clarifies existing specification behavior, will merge. Any further changes not included in this PR can be addressed in a follow-up PR.

PR #281, opened 2021-10-04
Update the default `axis` value for `vecdot` to be `-1`

kgryte:
This PR

-   resolves a discussion raised on NumPy (see https://github.com/numpy/numpy/pull/19980/files#r718067352).
-   changes the `axis` default from `None` to `-1`. Behavior is still preserved, in that, by default, the function computes the vector dot product over the last dimension.
-   supporting `None` was probably never intended, given that its associated behavior is equivalent to `axis=-1`.


asmeurer (approved):

kgryte:
As this has been approved, merging...

PR #289, opened 2021-10-21
Specify shape behavior

kgryte:
This PR

-   resolves [gh-97](https://github.com/data-apis/array-api/issues/97) and follows the guidance discussed [therein](https://github.com/data-apis/array-api/issues/97#issuecomment-948325555). **Update**: supporting unknown ranks is no longer included in this proposal (see [comment](https://github.com/data-apis/array-api/issues/97#issuecomment-961342592)).
-   allows `x.shape` to return a tuple with `None` elements to indicate that a dimension is unknown.
-   tightens guidance specifying that `x.shape` should return a tuple. A custom shape object is allowed so long as it behaves like a tuple.
-   removes `__len__` from the specification. NumPy, TF, and others return the size of the first dimension when `len()` (which calls `__len__`) is invoked on an array. In Python, [`__len__`](https://docs.python.org/3/reference/datamodel.html#object.__len__) should return integer `>=0`. Given the above changes accommodating unknown shapes/dimensions, returning an integer cannot be guaranteed as the first dimension could be unknown and thus `None`. Accordingly, this PR elects to remove `__len__` from the specification.

### DIFF
--- a/spec/API_specification/array_object.md
+++ b/spec/API_specification/array_object.md
@@ -270,7 +270,7 @@ This must equal the product of the array's dimensions.
 
 -   **out**: _Optional\[ int ]_
 
-    -   number of elements in an array. The returned value must be `None` if and only if an array dimension is unknown.
+    -   number of elements in an array. The returned value must be `None` if and only if one or more array dimensions are unknown.
 
 ```{note}
 For array libraries having graph-based computational models, an array may have unknown dimensions due to data-dependent operations.
--- a/spec/API_specification/array_object.md
+++ b/spec/API_specification/array_object.md
@@ -251,7 +251,7 @@ Array dimensions.
 
 -   **out**: _Optional\[ Tuple\[ Optional\[ int ], ... ] ]_
 
-    -   array dimensions. An array dimension must be `None` if and only if a dimension is unknown.
+    -   array dimensions. The returned value must be `None` if and only if the rank of an array is unknown. An array dimension must be `None` if and only if a dimension is unknown.
 
 ```{note}
 For array libraries having graph-based computational models, array dimensions may be unknown due to data-dependent operations (e.g., boolean indexing; `A[:, B > 0]`) and thus cannot be statically resolved without knowing array contents.
--- a/spec/API_specification/array_object.md
+++ b/spec/API_specification/array_object.md
@@ -234,13 +234,9 @@ Number of array dimensions (axes).
 
 #### Returns
 
--   **out**: _Optional\[ int ]_
+-   **out**: _int_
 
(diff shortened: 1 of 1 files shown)


kgryte:
For visibility, @shoyer @agarwal-ashish @jakirkham @alextp.

shoyer:
TensorFlow currently alows arrays with unknown numbers of dimensions, but in my opinion the use-cases (beyond backwards compatibility) are pretty marginal. On the other hand, arrays with unknown shape are much more common (e.g., at least in TensorFlow, JAX and Dask).

So I would lean towards typing `array.shape` as `Tuple[Optional[int], ...]]`.

I don't love the idea of allowing `array.shape` itself to return an array, because it's useful to be able to _statically_ check for fully known dimension sizes, which allow for some types of optimizations.

Instead, we could consider a separate function like [`tf.shape`](https://www.tensorflow.org/api_docs/python/tf/shape) for returning array shape as a 1D array.

kgryte:
@shoyer Agreed. We've discussed adding a `shape` API (MXNet also has `shape_array`) to support returning an array, but have pushed that to the 2022 version of the standard to gather additional requirements, with the focus on nailing down `x.shape` expectations for 2021.

I'm certainly open to modifying the current proposal to always return a tuple (i.e., disallowing supporting unknown rank).

rgommers:
> TensorFlow currently alows arrays with unknown numbers of dimensions, but in my opinion the use-cases (beyond backwards compatibility) are pretty marginal. On the other hand, arrays with unknown shape are much more common (e.g., at least in TensorFlow, JAX and Dask).
> 
> So I would lean towards typing `array.shape` as `Tuple[Optional[int], ...]]`.

This sounds about right to me - strikes a good balance between the needs of libraries using delayed evaluation and usability.

- It's one thing to explain to users that `x.shape` may have `None` for one or more elements of the tuple, so they should check for that - this is a small-ish cost to pay, because algorithms using `x.shape[1] == SPECIAL_LENGTH: ...` are not common.  
- It's quite another thing to say `x.ndim` may be `None`, so code like `x.ndim == 2: ...` (quite common) is now invalid.

rgommers:
IIRC we never really had any data on the "unknown rank" use cases. I thought our main remaining decision here was whether to represent the unknown dimension _inside_ `.shape` by `None` or `nan`.

kgryte:
Shape guidance has now been updated to remove support for unknown ranks, in alignment with @shoyer's proposal and as discussed on today's call.

kgryte:
As this has been discussed and agreed upon during consortium meetings and both here and in https://github.com/data-apis/array-api/issues/97, will merge. Any further changes can be addressed in follow-up PRs.

PR #296, opened 2021-11-01
Add note concerning the casting of boolean input arrays

kgryte:
This PR

-   adds a clarifying note regarding the casting of boolean input arrays. Namely, that `True` and `False` must cast to `1` and `0` numeric values, respectively, and `0` and non-zero numeric values must cast to `False` and `True`, respectively.
-    this was discussed during the consortium meeting held on 21 October 2021.


kgryte:
As this was discussed during the 21 October 2021 consortium meeting, will merge.

PR #297, opened 2021-11-02
Include entry points as an optional feature

### LINKED ISSUE #244: Proposal: use entrypoints to make array-api submodules discoverable
Over in https://github.com/HypothesisWorks/hypothesis/pull/3065, we're working on test tooling that takes an array-api implementation (e.g. `numpy.array_api`, etc) and returns test helpers which use that implementation.  For ergonomics we'd also like to automatically find the array-api implementation, if there is one, for a given module - but as I understand it there's no way to discover that mapping.  Yet!

I therefore propose that as part of the Array API standard, packages implementing the API should register an entry point declaring which (sub)module contains the implementation.  For example, Numpy would [register](https://github.com/numpy/numpy/blob/90ddaa5f88cab3b9143877aa1bb78f80fc38a9ac/setup.py#L411-L413):

```python
        entry_points={
            'console_scripts': f2py_cmds,
            'array_api': ['numpy = numpy.array_api'],  # new: maps top-level name to array api implementation
        },
```

Implementations can then be found using `importlib.metadata` ([see here for < py3.7](https://github.com/HypothesisWorks/hypothesis/blob/bf2c4be6f86ccb88a496373b24e92ad8e270bc01/hypothesis-python/src/hypothesis/entry_points.py)) as follows:

```python
import numpy.array_api
from importlib.metadata import entry_points

try:
    eps = entry_points(group="array_api")
except TypeError:
    # Load-time selection requires Python >= 3.10 or importlib_metadata >= 3.6, so fallback:
    eps = importlib_metadata.entry_points().get("array_api", [])
implementations = {entry.name: entry for entry in eps}

npa = implementations["numpy"].load()
assert npa is numpy.array_api
```

This allows us to avoid hard-coded lists of known libraries and locations, which will inevitably be both incomplete and out of date as the ecosystem evolves.  It also makes it easy to find all installed array-api implementations, supporting a much wider range of purposes than testing 🎉

honno:
Resolves #244 by adding a subsection that goes over entry points as an optional thing to adopt in the "How to adopt this API" section.

I thought to give an example of how it would be used, like how `xp = x.__array_namespace__()` is shown above. It's a bit awkward due to the [dict interface deprecation](https://importlib-metadata.readthedocs.io/en/latest/history.html#v3-9-0), but it seemed better to address it in the code example than in a separate note.

I also thought to give an example of defining an entry point in setup.py i.e. `entry_points={'array_api': ['package_name = package_name.array_api']}`, but that is a bit out-of-scope really.

### DIFF
--- a/spec/purpose_and_scope.md
+++ b/spec/purpose_and_scope.md
@@ -369,6 +369,38 @@ recommended not to add other functions or objects, because that may make it
 harder for users to write code that will work with multiple array libraries.
 
 
+### Discoverability
+
+Libraries that wish to [create arrays](../API_specification/creation_functions.md)
+using multiple array libraries would still need to explicitly import each
+library, as [`__array_namespace__()`](method-__array_namespace__) only makes an
+API namespace accessible when consuming arrays.
+
+To address this issue, an {pypa}`entry point <specifications/entry-points/>` may
+be provided by a conforming implementation to make its API namespace
+discoverable:
+
+```python
+from importlib.metadata import entry_points
+
+try:
+    eps = entry_points()['array_api']
+    ep = next(ep for ep in eps if ep.name == 'package_name')
+except TypeError:
+    # The dict interface for entry_points() is deprecated in py3.10,
+    # supplanted by a new select interface.
+    ep = entry_points(group='array_api', name='package_name')
+
+xp = ep.load()
+```
+
+The properties of an entry point should have:
+
+- `array_api` as the **group**
+- The package name as the **name**
+- The import path to the API namespace as the **object reference**
+
+
 * * *
 
 ## Conformance
--- a/spec/purpose_and_scope.md
+++ b/spec/purpose_and_scope.md
@@ -371,7 +371,7 @@ harder for users to write code that will work with multiple array libraries.
 
 ### Discoverability
 
-To assist array-consuming libraries which need to create arrays originating from multiple conforming array implementations, conforming implementations may provide an {pypa}`entry point <specifications/entry-points/>` in order to make an array API namespace discoverable:
+To assist array-consuming libraries which need to create arrays originating from multiple conforming array implementations, conforming implementations may provide an {pypa}`entry point <specifications/entry-points/>` in order to make an array API namespace discoverable. For example,
 
 ```python
 from importlib.metadata import entry_points
--- a/spec/purpose_and_scope.md
+++ b/spec/purpose_and_scope.md
@@ -371,7 +371,7 @@ harder for users to write code that will work with multiple array libraries.
 
 ### Discoverability
 
-To assist array-consuming libraries which need to create arrays originating from multiple conforming array implementations, conforming implementations may provide a {pypa}`entry point <specifications/entry-points/>` in order to make an array API namespace discoverable:
+To assist array-consuming libraries which need to create arrays originating from multiple conforming array implementations, conforming implementations may provide an {pypa}`entry point <specifications/entry-points/>` in order to make an array API namespace discoverable:
 
 ```python
 from importlib.metadata import entry_points
--- a/spec/purpose_and_scope.md
+++ b/spec/purpose_and_scope.md
@@ -371,14 +371,7 @@ harder for users to write code that will work with multiple array libraries.
 
 ### Discoverability
 
-Libraries that wish to [create arrays](../API_specification/creation_functions.md)
-using multiple array libraries would still need to explicitly import each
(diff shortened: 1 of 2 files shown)


spec/purpose_and_scope.md
```diff
@@ -369,6 +369,38 @@ recommended not to add other functions or objects, because that may make it
...
+### Discoverability
+
+Libraries that wish to [create arrays](../API_specification/creation_functions.md)
+using multiple array libraries would still need to explicitly import each
+library, as [`__array_namespace__()`](method-__array_namespace__) only makes an
+API namespace accessible when consuming arrays.
+
+To address this issue, an {pypa}`entry point <specifications/entry-points/>` may
+be provided by a conforming implementation to make its API namespace
+discoverable:
```
kgryte, line 381:
```suggestion
To assist array-consuming libraries which need to create arrays originating from multiple conforming array implementations, conforming implementations may provide an {pypa}`entry point <specifications/entry-points/>` in order to make an array API namespace discoverable:
```

spec/purpose_and_scope.md
```diff
@@ -369,6 +369,38 @@ recommended not to add other functions or objects, because that may make it
...
+    ep = entry_points(group='array_api', name='package_name')
+
+xp = ep.load()
+```
+
+The properties of an entry point should have:
+
+- `array_api` as the **group**
+- The package name as the **name**
+- The import path to the API namespace as the **object reference**
```
kgryte, line 401:
```suggestion
An entry point must have the following properties:

-   **group**: equal to `array_api`.
-   **name**: equal to the package name.
-   **object reference**: equal to the array API namespace import path.
```

spec/purpose_and_scope.md
```diff
@@ -369,6 +369,31 @@ recommended not to add other functions or objects, because that may make it
 harder for users to write code that will work with multiple array libraries.
 
 
+### Discoverability
+
+To assist array-consuming libraries which need to create arrays originating from multiple conforming array implementations, conforming implementations may provide an {pypa}`entry point <specifications/entry-points/>` in order to make an array API namespace discoverable:
```
kgryte, line 374:
```suggestion
To assist array-consuming libraries which need to create arrays originating from multiple conforming array implementations, conforming implementations may provide a {pypa}`entry point <specifications/entry-points/>` in order to make an array API namespace discoverable:
```

spec/purpose_and_scope.md
```diff
@@ -369,6 +369,31 @@ recommended not to add other functions or objects, because that may make it
 harder for users to write code that will work with multiple array libraries.
 
 
+### Discoverability
+
+To assist array-consuming libraries which need to create arrays originating from multiple conforming array implementations, conforming implementations may provide a {pypa}`entry point <specifications/entry-points/>` in order to make an array API namespace discoverable:
```
kgryte, line 374:
```suggestion
To assist array-consuming libraries which need to create arrays originating from multiple conforming array implementations, conforming implementations may provide an {pypa}`entry point <specifications/entry-points/>` in order to make an array API namespace discoverable:
```

kgryte (approved):
LGTM. Thanks, @honno!

If we need to make further changes, we can do so in a follow-up PR.

honno (commented):
@kgryte as per slack—yep your changes all look good.

spec/purpose_and_scope.md
```diff
@@ -369,6 +369,31 @@ recommended not to add other functions or objects, because that may make it
 harder for users to write code that will work with multiple array libraries.
 
 
+### Discoverability
+
+To assist array-consuming libraries which need to create arrays originating from multiple conforming array implementations, conforming implementations may provide an {pypa}`entry point <specifications/entry-points/>` in order to make an array API namespace discoverable. For example,
```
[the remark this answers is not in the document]
kgryte: Thanks for looking over!

spec/purpose_and_scope.md
```diff
@@ -369,6 +369,31 @@ recommended not to add other functions or objects, because that may make it
 harder for users to write code that will work with multiple array libraries.
 
 
+### Discoverability
+
```
rgommers, line 373: This is a bit cryptic for a reader, like why does it exist? what goes wrong if an entrypoint is not implemented? The one word _may_ is not enough, I'll try to add a note with rationale. Especially because this is in the "Purpose and scope" section, that is important.

kgryte: Thanks!

PR #298, opened 2021-11-02
Arrays must share the same dtype in `meshgrid()`

honno:
This PR clarifies that input arrays must share the same dtype.

Allowing mixed dtypes complicates both user expectations and developer implementations for the returned arrays, which is currently outlined as all having the same promoted dtype (differing from both NumPy and PyTorch behaviour). It also does not seem like anyone would use or desire input arrays of mixed dtypes.

```python
>>> x1 = np.arange(5, dtype=np.int32)
>>> x2 = np.arange(5, dtype=np.int64)
>>> out1, out2 = np.meshgrid(x1, x2)
>>> out1.dtype
int32  # same as x1.dtype
>>> out2.dtype
int64  # same as x2.dtype
...
>>> t1 = torch.arange(5, dtype=torch.int32)
>>> t2 = torch.arange(5, dtype=torch.int64)
>>> torch.meshgrid(t1, t2)
RuntimeError: meshgrid expects all tensors to have the same dtype
```

### DIFF
--- a/spec/API_specification/creation_functions.md
+++ b/spec/API_specification/creation_functions.md
@@ -322,7 +322,7 @@ Returns coordinate matrices from coordinate vectors.
 
 -    **arrays**: _&lt;array&gt;_
 
-     -   an arbitrary number of one-dimensional arrays representing grid coordinates. Must have numeric data types.
+     -   an arbitrary number of one-dimensional arrays representing grid coordinates. Each array must have the same numeric data type.
 
 -    **indexing**: _str_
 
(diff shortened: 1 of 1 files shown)


spec/API_specification/creation_functions.md
```diff
@@ -322,7 +322,7 @@ Returns coordinate matrices from coordinate vectors.
 
 -    **arrays**: _&lt;array&gt;_
 
-     -   an arbitrary number of one-dimensional arrays representing grid coordinates. Must have numeric data types.
+     -   an arbitrary number of one-dimensional arrays representing grid coordinates. Each array must have the same numeric data type.
```
asmeurer, line 325: We should use "should" here instead of "must". That way libraries can support mixed dtype behavior if they want, but the spec only specifies behavior when they match.

honno: Makes sense, thanks!

I also changed the must to should in the modified return part of meshgrid. I *think* it's appropriate too?

```diff
- Each returned array must have the same data type as the input arrays.
+ Each returned array should have the same data type as the input arrays.
```

asmeurer: I think so, since any sort of type promotion would only occur when the inputs have different dtypes anyway, which is outside of the scope of the spec.

kgryte (approved):
LGTM.

kgryte:
Thanks, @honno!

PR #301, opened 2021-11-03
Remove DLPack support from asarray

### LINKED ISSUE #155: Do we need from_dlpack when DLPack is supported by asarray?
I guess one possible advantage of `from_dlpack()` is that it's more explicit? If that's desired, then perhaps we should add other explicit constructors like `from_buffer()`. Otherwise, it seems redundant with the generic `asarray()`, and I would lean towards removing it.

kgryte:
This PR

-   resolves #155 by removing DLPack support from `asarray`. To create an array via DLPack, users can use `from_dlpack`.
-   the changes in this PR were discussed and approved during the consortium meeting held on 21 October 2021.

### DIFF
--- a/spec/API_specification/creation_functions.md
+++ b/spec/API_specification/creation_functions.md
@@ -57,12 +57,11 @@ Convert the input to an array.
 
 #### Parameters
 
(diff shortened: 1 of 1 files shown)


kgryte:
As the changes were already discussed and approved, will merge...

leofang:
Thanks @kgryte!

kgryte:
@leofang Of course! 😃  Thanks for all your input!

PR #302, opened 2021-11-04
Add notes concerning input array validation

kgryte:
This PR

-   addresses concerns raised in #224 (see [comment](https://github.com/data-apis/array-api/pull/224#discussion_r679622056)) by stating whether conforming array libraries must validate input arrays based on API assumptions. In all instances, this PR indicates that whether validation occurs is implementation-defined.


kgryte:
The content of this PR has already been discussed during consortium meetings (circa July 2021) and on PR [threads](https://github.com/data-apis/array-api/pull/224#discussion_r679622056). Accordingly, will merge. If anything needs to change, we can address in a follow-up PR.

PR #303, opened 2021-11-04
Update `slogdet` guidance for when the determinant is zero

kgryte:
This PR

-   updates the guidance in `linalg.slogdet` for when the determinant is zero. The original language stemmed from NumPy; however, the guidance there is more conceptual than grounded in practice due to errors in numerical precision (see https://github.com/numpy/numpy/issues/20114). Accordingly, this PR loosens restrictions around return values when the determinant is `0`. Namely, the result **should** be sign `0` and logabsdet `-inf`, but may not be due to numerical precision errors.


leofang (approved):

kgryte:
Thanks @leofang! 😅

Issue #216, opened 2021-07-01
Definition of `rtol` in  `matrix_rank`, `lstsq` and `pinv`


Lezcano:
This can be seen as a follow-up of https://github.com/data-apis/array-api/issues/211.

The current definition of `rtol` in these functions reads:
> rtol: relative tolerance for small singular values. Singular values less than or equal to `rtol * largest_singular_value` are set to zero

This rules out the implementation of these functions via algorithms based on the QR decomposition, which goes against [Design Principle 4: Implementation Agnosticism](https://data-apis.org/array-api/latest/extensions/linear_algebra_functions.html?linalg-eigh-x-upper-false#design-principles).

Note that in LAPACK, the meaning of `rcond` in [SVD-based](https://www.netlib.org/lapack/explore-html/d7/d3b/group__double_g_esolve_ga94bd4a63a6dacf523e25ff617719f752.html) lstsq solvers and [QR-solvers](http://www.netlib.org/lapack/explore-html/d7/d3b/group__double_g_esolve_ga385713b8bcdf85663ff9a45926fac423.html) are somewhat equivalent. The one using a QR-decomposition defines `RCOND` for the decomposition
```
A * P = Q * [ R11 R12 ]
            [  0  R22 ]
```
as the smallest `k` for which `\sigma_1(R11) / \sigma_{k+1}(R11) >= 1/RCOND`, that is, the `k` for which `RCOND\sigma_1(R11) >= \sigma_{k+1}(R11)`. In this case they assume `R22 = 0`. Note that this is an approximation of the definition for the SVD, for which if `RCOND\sigma_1(A) >=\sigma_{k+1}(A)` then `\sigma_r(A)` is considered zero for `r = k+1,...,n`.

For this reason, a solution to this problem might go through rewording the definition of `rtol` in this functions in terms of condition numbers, although it is not clear to me how to write this down in a simple way.

A simpler but less satisfactory way to address this would be to simply define `rtol` as
> Singular values **approximately** less than or equal to rtol * largest_singular_value are set to zero.

I do not completely like this solution, because it sort of biases the definition to be based on SVD-like algorithms, and it takes a few seconds to realise that this is also what LAPACK's QR with pivoting is doing .

cc @rgommers @mruberry

PR #304, opened 2021-11-04
Update guidance for linalg `rtol` tolerance cutoff to accommodate non-SVD-like algorithms

kgryte:
This PR

-   resolves https://github.com/data-apis/array-api/issues/216 by updating the guidance for `rtol` to accommodate non-SVD-like algorithms. Namely, instead of a strict cutoff of less than or equal to, this PR updates the guidance to say "approximately less than or equal to" to allow conforming implementations some wiggle room should they choose a non-SVD-like algorithm for implementing `pinv`. For `matrix_rank`, the API is already defined in terms of singular values, so, in practice, an implementation is likely to be SVD-based; in which case, the more strict guidance is likely to hold.
-   in general, the previous guidance should hold true, given the prevalent usage of LAPACK; however, array libraries should be able to use alternative algorithms in accordance with the design principle to be "implementation agnostic". A future refinement could be to define the tolerance in terms of condition numbers, but that is left to future consideration.

### DIFF
--- a/spec/extensions/linear_algebra_functions.md
+++ b/spec/extensions/linear_algebra_functions.md
@@ -345,7 +345,7 @@ Returns the rank (i.e., number of non-zero singular values) of a matrix (or a st
 
 -   **rtol**: _Optional\[ Union\[ float, &lt;array&gt; ] ]_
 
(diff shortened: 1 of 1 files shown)


leofang (approved):

kgryte:
Thanks @leofang!

Issue #285, opened 2021-10-15
linalg.solve broadcasting behavior is ambiguous


asmeurer:
The spec for the [`linalg.solve`](https://data-apis.org/array-api/latest/extensions/linear_algebra_functions.html#linalg-solve-x1-x2) function seems ambiguous. In `solve(x1, x2)`, `x1` has shape `(..., M, M)` and `x2` either has shape `(..., M)` or `(..., M, K)`. In either case, the `...` parts should be broadcast compatible.

This is ambiguous. For example, if `x1` is shape `(2, 2, 2)` and `x2` is shape `(2, 2)`, should this be interpreted as `x2` is `(2,)` stack of a `(2,)` vector, i.e., the result would be `(2, 2, 2, 1)` after broadcasting, or as a single stack of a 2x2 matrix, i.e., resulting in `(2, 2, 2, 2)`. 

- Relevant pytorch issue about this: https://github.com/pytorch/pytorch/issues/52915
- Relevant NumPy issue: https://github.com/numpy/numpy/issues/15349
- `torch.linalg.solve` docs: https://pytorch.org/docs/stable/generated/torch.linalg.solve.html
- `numpy.linalg.solve` docs: https://numpy.org/doc/stable/reference/generated/numpy.linalg.solve.html#numpy.linalg.solve

Regarding NumPy, it seems to sometimes pick one over the other, even when only the other one makes sense. For example

```py
>>> x1 = np.eye(1)
>>> x2 = np.asarray([[0.], [0.]])
>>> x1.shape
(1, 1)
>>> x2.shape
(2, 1)
>>> np.linalg.solve(x1, x2)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<__array_function__ internals>", line 5, in solve
  File "/Users/aaronmeurer/anaconda3/envs/array-apis/lib/python3.9/site-packages/numpy/linalg/linalg.py", line 393, in solve
    r = gufunc(a, b, signature=signature, extobj=extobj)
ValueError: solve: Input operand 1 has a mismatch in its core dimension 0, with gufunc signature (m,m),(m,n)->(m,n) (size 2 is different from 1)
```

Here it wants to treat `x2` as a single 2x1 matrix, which is shape incompatible with the 1x1 `x1`, but it could also treat it a `(2,)` stacks of length 1 vectors.

[...]

PR #305, opened 2021-11-04
Disallow providing a stack of vectors to `linalg.solve`

kgryte:
This PR

-   resolves https://github.com/data-apis/array-api/issues/285 by disallowing providing stacks of one-dimensional vectors. This resolves an ambiguity when `M=K` and determination of broadcasting semantics.

### DIFF
--- a/spec/extensions/linear_algebra_functions.md
+++ b/spec/extensions/linear_algebra_functions.md
@@ -484,7 +484,7 @@ Whether an array library explicitly checks whether an input array is full rank i
 
 -   **x2**: _&lt;array&gt;_
 
-    -   ordinate (or "dependent variable") array `B`. If `x2` has shape `(..., M)`, `x2` is equivalent to an array having shape `(..., M, 1)`, and `shape(x2)` must be compatible with `shape(x1)[:-1]` (see {ref}`broadcasting`). If `x2` has shape `(..., M, K)`, each column `k` defines a set of ordinate values for which to compute a solution, and `shape(x2)[:-1]` must be compatible with `shape(x1)[:-1]` (see {ref}`broadcasting`). Should have a floating-point data type.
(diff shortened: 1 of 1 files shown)


spec/extensions/linear_algebra_functions.md
```diff
@@ -484,7 +484,7 @@ Whether an array library explicitly checks whether an input array is full rank i
 
 -   **x2**: _&lt;array&gt;_
 
-    -   ordinate (or "dependent variable") array `B`. If `x2` has shape `(..., M)`, `x2` is equivalent to an array having shape `(..., M, 1)`, and `shape(x2)` must be compatible with `shape(x1)[:-1]` (see {ref}`broadcasting`). If `x2` has shape `(..., M, K)`, each column `k` defines a set of ordinate values for which to compute a solution, and `shape(x2)[:-1]` must be compatible with `shape(x1)[:-1]` (see {ref}`broadcasting`). Should have a floating-point data type.
+    -   ordinate (or "dependent variable") array `B`. If `x2` has shape `(M)`, `x2` is equivalent to an array having shape `(..., M, 1)`. If `x2` has shape `(..., M, K)`, each column `k` defines a set of ordinate values for which to compute a solution, and `shape(x2)[:-1]` must be compatible with `shape(x1)[:-1]` (see {ref}`broadcasting`). Should have a floating-point data type.
```
leofang, line 487: nitpick:
```suggestion
    -   ordinate (or "dependent variable") array `B`. If `x2` has shape `(M,)`, `x2` is equivalent to an array having shape `(..., M, 1)`. If `x2` has shape `(..., M, K)`, each column `k` defines a set of ordinate values for which to compute a solution, and `shape(x2)[:-1]` must be compatible with `shape(x1)[:-1]` (see {ref}`broadcasting`). Should have a floating-point data type.
```

kgryte:
Thanks, @leofang!

PR #306, opened 2021-11-04
Remove obsolete note

kgryte:
This PR

-   removes an obsolete note concerning adoption. The note referenced NEP 37; however, that proposal has not seen the requisite movement necessary to change either the specification's use of `__array_namespace__` or the experimental implementation in NumPy. Accordingly, this PR chooses to remove the note to avoid creating any uncertainty regarding the preferred mechanism of exposing a conforming array implementation as a separate namespace.

cc @rgommers for visibility.

### DIFF
--- a/spec/purpose_and_scope.md
+++ b/spec/purpose_and_scope.md
@@ -349,24 +349,10 @@ request a specific API version:
 xp = x.__array_namespace__(api_version='2020.10')
 ```
 
-```{note}
-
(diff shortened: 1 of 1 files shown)


kgryte:
Will merge. If further changes are necessary, they can be made in follow-up PRs.

rgommers (commented):
Agreed with removal. NEP 37 was useful inspiration, but it never went anywhere so there's no point in referencing it.

PR #307, opened 2021-11-04
Disallow scalar operands when using the matmul operator

kgryte:
This PR

-   clarifies that the matmul `@` operator should not support Python scalar types. See https://github.com/data-apis/array-api/issues/98.
-   removes linebreaks in the document in order to be consistent--some lines broke; others did not.
-   updates spelling and rephrases select sentences.


kgryte:
cc @asmeurer for visibility.

PR #308, opened 2021-11-04
Group operators into separate classes

kgryte:
This PR

-   follows up https://github.com/data-apis/array-api/issues/98 which recommended grouping the operators into distinct classes.
-   for each class, defines the data types on which the respective operators should operate.
-   updates copy.


kgryte:
cc @asmeurer for visibility.

Issue #249, opened 2021-09-01
Specify NaN behaviour in `unique()`


honno:
Say we have an array with multiple NaN values:
```
>>> x = xp.full(5, xp.nan)
>>> x
Array([nan, nan, nan, nan, nan])
```
Should NaNs be treated as unique to one another?
```
>>> xp.unique(x)
Array([nan, nan, nan, nan, nan])
```
Or should they be treated as the same?
```
>>> xp.unique(x)
Array([nan])
```

I have no dog in the race but you might want to refer to some discussion bottom of numpy/numpy#19301 and in the [NumPy mailing list](https://mail.python.org/pipermail/numpy-discussion/2021-August/081995.html) that relates to a recent accidental "regression" in how `np.unique()` deals with NaNs.

In either case, just specificity would prevent creating wrapper methods to standardise this behaviour. For example I created this (admittedly scrappy) utility method for HypothesisWorks/hypothesis#3065 to work for both scenarios, which @asmeurer noted was not ideal and probably a failure of the Array API spec.

```python
def count_unique(array):
    n_unique = 0
    nan_index = xp.isnan(array)
    for isnan, count in zip(*xp.unique(nan_index, return_counts=True)):
        if isnan:
            n_unique += count
            break
    filtered_array = array[~nan_index]
    unique_array = xp.unique(filtered_array)
    n_unique += unique_array.size
    return n_unique
```

And if deterministic NaN behaviour cannot be part of the API, a note should be added to say this behaviour is out of scope.

pmeier:
IMO this depends on the definition of "unique". If it means "value equality", it should return all occurrences of NaN, since `assert float("nan") != float("nan")`. If it means "identity", it should return only one NaN, since `assert float("nan") is float("nan")`.

If I'm not mistaken, in this context "unique" means value equality, since we are dealing with values and not objects (is `np.unique` defined for the `object` dtype?). To satisfy both requirements, we could add an `equal_nan` flag that, if `True`, treats NaN's as equal. This is similar to what most testing functions do, e.g. [`np.testing.assert_allclose`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose).

rgommers:
[...]

PR #310, opened 2021-11-04
Require that NaNs be considered distinct when returning unique elements

kgryte:
This PR

-   resolves https://github.com/data-apis/array-api/issues/249 by requiring uniqueness be determined based on value equality (`==`).
-   requires that NaN values be considered distinct when returning the unique elements of an array. Accordingly, if an array contains multiple NaN values, the returned array will also contain multiple NaN values.
-   specifies that signed zeros be treated as equal. As a consequence, reconstructing an array containing signed zeros is not guaranteed to return the exact same array. Treating signed zeros as equal is the current prevailing behavior among array libraries (NumPy, MXNet, TF, ...).


kgryte:
The changes introduced in this PR were discussed and approved during the consortium meeting held on 21 October 2021. Will merge. Any further changes can be addressed in follow-up PRs.

Issue #199, opened 2021-06-11
Specify zero division behavior


pmeier:
numpy and PyTorch diverge on the zero division behavior of integral `dtype`'s:

```python
>>> a1 = np.array(1, dtype=np.int)
>>> a2 = np.array(0, dtype=np.int)
>>> a1 // a2
0
>>> a1 % a2
0
>>> t1 = torch.tensor(1, dtype=torch.int)
>>> t2 = torch.tensor(0, dtype=torch.int)
>>> t1 // t2
RuntimeError: ZeroDivisionError
>>> t1 % t2
RuntimeError: ZeroDivisionError
```

We should either specify the behavior or fix the test suite to exclude zero as second input when testing these functions. cc @asmeurer

rgommers:
The numpy code does give:
```python
>>> a1 // a2
<ipython-input-42-136abf062072>:1: RuntimeWarning: divide by zero encountered in floor_divide
  a1 // a2
0
>>> a1 % a2
<ipython-input-43-bdabd79cdfbc>:1: RuntimeWarning: divide by zero encountered in remainder
  a1 % a2
0
```
There's an issue about it here: https://github.com/numpy/numpy/issues/5150

I don't think this should be specified in the array API spec, it falls under the warnings/exceptions behavior which is library-specific.

rgommers:
And for the record: there's something to say for either option, but I personally prefer the PyTorch behavior.

pmeier:
> I don't think this should be specified in the array API spec, it falls under the warnings/exceptions behavior which is library-specific.

In that case, we need to exclude it as second input for these functions in the test suite.

asmeurer:
The spec should be updated first. The test suite should be easy to update. We already have similar code for the bitwise shift operators, which only accept nonnegative values for the second argument.

pmeier:
@asmeurer From https://github.com/data-apis/array-api/issues/199#issuecomment-860041496

> I don't think this should be specified in the array API spec

Thus, I think we can go ahead and adapt the test suite.

asmeurer:
I disagree with that. The spec should specify what range of inputs are required for a specific function. If this confused you, it will confuse other people too.

kgryte:
@asmeurer Not clear to me how you think the specification should be changed. Are you looking for a note that saying zero-division behavior is undefined (i.e., library-specific)?

asmeurer:
[...]

PR #311, opened 2021-11-04
Add note concerning division by zero for integer dtypes in floor_divide and remainder

kgryte:
This PR

-   resolves https://github.com/data-apis/array-api/issues/199 by adding a note to `floor_divide` and `remainder` concerning expected behavior when input arrays promote to an integer dtype. Namely, that the result of division by zero is implementation defined (e.g., throwing an error, emitting a warning and returning zero, or something else).

cc @asmeurer


kgryte:
As the proposed changes in this PR do not change the status quo of the standard, but merely clarified unspecified behavior, will merge. Any subsequent revisions can be raised in a subsequent issue and addressed in a follow-up PR.

PR #312, opened 2021-11-04
Add guidance for when rounding special floating-point values

kgryte:
This PR

-   clarifies rounding behavior when encountering special floating-point values, such as `NaN`, `+-infinity`, and `+-0`. The lack of guidance was raised in https://github.com/data-apis/array-api/issues/199.
-   adds special cases to `ceil`, `floor`, `round`, and `trunc`.

cc @asmeurer


kgryte:
As this PR clarifies special case behavior for floating-point operands and does not introduce any controversial changes (all array libraries following IEEE 754 should already follow the guidance introduced in this PR), will merge.

PR #314, opened 2021-11-05
Remove obsolete TODOs

kgryte:
This PR

-   removes a few obsolete TODO items from the "Purpose and Scope" section of the specification. The first TODO pertains to the initial RFC and can be removed now that the standard is moving toward official release. The second and third TBDs concern optional extensions. As the `linalg` extension has been added, it provides a cowpath for subsequent extensions and demonstrates how extensions are included in the standard.

### DIFF
--- a/spec/purpose_and_scope.md
+++ b/spec/purpose_and_scope.md
@@ -48,14 +48,6 @@ functionality that's not present in this document?_ This:
 - may indicate that that functionality, if present in a particular array
   library, is unlikely to be present in all other libraries
 
-```{note}
-
-This document is ready for wider community review, but still contains a
-number of TODOs, and is expected to change and evolve before a first
-official release. See {ref}`future-API-evolution` for proposed
-versioning.
-```
-
 ### History
 
 The first library for numerical and scientific computing in Python was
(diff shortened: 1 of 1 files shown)


kgryte:
cc @rgommers One thing that may need to change is figure in the ["in scope" section](https://data-apis.org/array-api/latest/purpose_and_scope.html#in-scope). This currently says `TBD`.

And one consideration not addressed in the specification is whether extensions can be versioned independently of the main (non-extension) namespace. Atm, the `linalg` extension resides in a separate sub-namespace, and the specification is not clear on whether the `linalg` sub-namespace should also include a [version](https://data-apis.org/array-api/latest/future_API_evolution.html#versioning). We may want to make that more clear.

rgommers:
> One thing that may need to change is the figure in the ["in scope" section](https://data-apis.org/array-api/latest/purpose_and_scope.html#in-scope). This currently says `TBD`.

I'll update that figure.

> And one consideration not addressed in the specification is whether extensions can be versioned independently of the main (non-extension) namespace. Atm, the `linalg` extension resides in a separate sub-namespace, and the specification is not clear on whether the `linalg` sub-namespace should also include a [version](https://data-apis.org/array-api/latest/future_API_evolution.html#versioning). We may want to make that more clear.

I'd say no, it's just optional but it's not a completely separate thing with its own versioning scheme.

kgryte:
> I'll update that figure.

Thanks!

Issue #279, opened 2021-10-02
Explicit rules about zero-dimensional tensors


arogozhnikov:
Documentation mentions in several places zero-dimensional tensors, but it's not clear to me if those are required to be supported.

While it sounds very reasonable to assume that e.g. xp.sum(x) returns scalar, some of frameworks don't provide support for zero-dimensional arrays, and fake it with 1-element 1-dimensional arrays. Examples:
- mxnet fails on creation `mx.nd.zeros([])` I know API will be reworked in mxnet 2, but did not check actual support there
- paddle also lacks zero-dim arrays, it does not seem to be in any priority

rgommers:
Thanks for pointing that out @arogozhnikov! 0-D arrays are required, so we should make that more explicit.

rgommers:
I'm not sure we need a whole separate section, just a note near the top of https://data-apis.org/array-api/latest/API_specification/array_object.html should be fine I think?

kgryte:
Okay. Will add.

PR #315, opened 2021-11-05
Add guidance requiring support for zero-dimensional arrays

kgryte:
This PR

-   resolves https://github.com/data-apis/array-api/issues/279 by adding explicit guidance requiring support for zero-dimensional arrays in order for an array implementation to be compliant.


leofang (approved):

rgommers (approved):
LGTM too.

Cc @szha for visibility, because the issue explicitly mentions that MXNet currently does not support 0-D arrays. Probably on the MXNet roadmap, would be good to know for sure.

kgryte:
Thanks @leofang and @rgommers!

Issue #288, opened 2021-10-21
Clarify floating-point sort order for ascending and descending sort


kgryte:
Currently, the specification does not indicate the behavior for sorting the IEEE 754 floating-point values: `NaN`, `-0`, and `+0`.

In NumPy, `NaN` values are sorted to the end and signed zeros are not sorted:

```python
In [1]: np.sort([2.0, 0.0, 1.0, -0.0, np.nan, np.nan, 3.0])                                                                                   
Out[1]: array([ 0., -0.,  1.,  2.,  3., nan, nan])
```

When sorting lists in Python, `NaN` values are left in place and signed zeros are not sorted:

```python
In [1]: list = [2.0,0.0,1.0,-0.0,float('nan'),float('nan'),3.0,-1.0];                                                                         

In [2]: list.sort()                                                                                                                           

In [3]: list                                                                                                                                  
Out[3]: [-1.0, 0.0, -0.0, 1.0, 2.0, nan, nan, 3.0]
```

Other languages have made different design decisions. E.g., Julia sorts signed zeros and places `NaN` values at the end.

```julia
julia> sort([2.0,1.0,0.0,-0.0,NaN,NaN,3.0,-1.0])
8-element Array{Float64,1}:
  -1.0
  -0.0
   0.0
   1.0
   2.0
   3.0
 NaN  
 NaN
```

IMO, we should clarify the sort order for ascending and descending sort. Preferably,

-   sorting signed zeros
-   sorting `NaN` values to the end when sorting in **ascending** order and to the beginning when sorting in **descending** order

PR #316, opened 2021-11-05
Add note concerning float sort order

kgryte:
This PR

-   resolves https://github.com/data-apis/array-api/issues/288 by adding a note concerning the sort order IEEE 754 floating-point values (`NaN`, `+0`, and `-0`). Currently, array libraries are not consistent in how they sort these values. Accordingly, this PR simply adds a note indicating that the sort order of these values is left unspecified and is thus implementation-dependent.


leofang (approved):

kgryte:
Thanks @leofang!

PR #317, opened 2021-11-05
Add `unique_counts()` and fix the description of `unique_all()`

### LINKED ISSUE #291: Add `unique_counts()`?
This is a follow-up of #275. I've found `numpy.unique(x, return_counts=True)` quite useful. It allows us to build a histogram for the input array `x`. I'd suggest us to add the following function:
```python
def unique_counts(x, /):
   """
   return a tuple of two arrays (unique values and counts)
   """
```

leofang:
Close #291.

This PR:
- adds a new API `unique_counts` as suggested in #291 to return unique values and their corresponding counts
- fixes the description of `unique` APIs:
  - the description of `unique_all()` was copied/pasted, which is now properly expanded
  - ~~the "default integer type" mentions are not accurate if the index/count is too large. Now we allow type promotion as needed.~~ **UPDATE**: discussion moved to #319.

### DIFF
--- a/spec/API_specification/set_functions.md
+++ b/spec/API_specification/set_functions.md
@@ -21,7 +21,7 @@ A conforming implementation of the array API standard must provide and support t
 The shapes of two of the output arrays for this function depend on the data values in the input array; hence, array libraries which build computation graphs (e.g., JAX, Dask, etc.) may find this function difficult to implement without knowing array values. Accordingly, such libraries may choose to omit this function. See {ref}`data-dependent-output-shapes` section for more details.
 :::
 
-Returns the unique elements of an input array `x`.
+Returns the unique elements of an input array `x`, the first occurring indices for each unique element in `x`, the indices from the set of unique elements that reconstruct `x`, and the corresponding counts for each unique element in `x`.
 
 ```{note}
 Uniqueness should be determined based on value equality (i.e., `x_i == x_j`). For input arrays having floating-point data types, value-based equality implies the following behavior.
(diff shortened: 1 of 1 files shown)


leofang:
hmmm why does this PR only have 1 check instead of 3?

spec/API_specification/set_functions.md
```diff
@@ -47,9 +47,50 @@ Each `nan` value should have a count of one, while the counts for signed zeros s
...
+        -   second element must have the field name `indices` and must be an array containing the indices (first occurrences) of `x` that result in `values`. The array must have the same shape as `values` and must have the default integer data type (or promoted if needed).
+        -   third element must have the field name `inverse_indices` and must be an array containing the indices of `values` that reconstruct `x`. The array must have the same shape as `x` and must have the default integer data type (or promoted if needed).
+        -   fourth element must have the field name `counts` and must be an array containing the number of times each unique element occurs in `x`. The returned array must have same shape as `values` and must have the default integer data type (or promoted if needed).
+
+        ```{note}
+        The order of unique elements is not specified and may vary between implementations.
+        ```
+
+(function-unique-count)=
```
kgryte, line 58: Bikeshed: `unique_count` or `unique_counts`?

kgryte: Not sure whether we want alignment with _unique_value**s**_ here.

leofang: I actually meant to add `unique_counts`, so it's a typo on my side 😅 But let's make the decision first...

spec/API_specification/set_functions.md
```diff
@@ -47,9 +47,50 @@ Each `nan` value should have a count of one, while the counts for signed zeros s
...
 
         -   first element must have the field name `values` and must be an array containing the unique elements of `x`. The array must have the same data type as `x`.
-        -   second element must have the field name `indices` and must be an array containing the indices (first occurrences) of `x` that result in `values`. The array must have the same shape as `values` and must have the default integer data type.
-        -   third element must have the field name `inverse_indices` and must be an array containing the indices of `values` that reconstruct `x`. The array must have the same shape as `x` and must have the default integer data type.
-        -   fourth element must have the field name `counts` and must be an array containing the number of times each unique element occurs in `x`. The returned array must have same shape as `values` and must have the default integer data type.
+        -   second element must have the field name `indices` and must be an array containing the indices (first occurrences) of `x` that result in `values`. The array must have the same shape as `values` and must have the default integer data type (or promoted if needed).
```
kgryte, line 50: How would promotion work? Would an implementing library need to internally copy to a new array upon overflow?

Here, would not seem that an implementing library would be able to statically know what the output dtype should be until actually processing the data. Suppose the input array has `2**33` elements (`>2**31-1`) and the default integer data type is `int32`. If all elements are unique, then counts for all would be `1`; while if all are the same, then we'd have a single count which would trigger promotion.

Instead of promotion, we could say something to the effect of "must have an `int64` dtype (i.e., the largest integer data type supported by the standard)".

spec/API_specification/set_functions.md
kgryte: I think `counts` makes sense to me. :)

kgryte:
@leofang Not sure why it has only one check. 🧐

spec/API_specification/set_functions.md
leofang: Fixed!

leofang:
> Not sure why it has only one check. 🧐

My guess is the other two CIs only run for this repo's branches, not for branches from forks. There is a piece of permission bit that needs to be set (at least it's the case for GH actions).

spec/API_specification/set_functions.md
kgryte: Conformant libraries are already expected to support `int64`: https://github.com/data-apis/array-api/blob/aa6c89c47c6f6f37fa0b53934b3c5784985d0628/spec/API_specification/data_types.md#data-types

This have require host emulation for 32-bit systems, but support is still expected.

If need be, we could add a disclaimer, but the spec does not currently allow leeway for conforming libraries to elide certain dtypes.

kgryte:
Ah, right. This was probably intentional in order to avoid drive-by PRs consuming resources.

spec/API_specification/set_functions.md
leofang: > How would promotion work? Would an implementing library need to internally copy to a new array upon overflow?
>
> Here, would not seem that an implementing library would be able to statically know what the output dtype should be until actually processing the data. 

Yes, but most likely `unique_*` will be implemented in two passes (so splitting `unique` gives some room for optimization), so by the time we need to allocate space for indices/counts, it's statically known.

> Suppose the input array has `2**33` elements (`>2**31-1`) and the default integer data type is `int32`. If all elements are unique, then counts for all would be `1`; while if all are the same, then we'd have a single count which would trigger promotion.

Yes.

> Instead of promotion, we could say something to the effect of "must have an `int64` data type (i.e., the largest integer data type supported by the standard)".

We could say that too, I don't have strong opinions here, my point is saying "default integer type" is not sufficient 🙂

kgryte: Originally, when writing the `unique` specification, the output dtype was the "default index data type". I wonder if we need to revive that distinction. Namely, that an array library should have three default data types:

1.  floating-point data type
2.  integer data type
3.  index data type

Here, having a default index data type makes sense, as counts should align accordingly (i.e., a count should never exceed the maximum array index).

Furthermore, while it may often be the case that indices will have the same dtype as the default integer dtype, this need not be the case. For example, indices may be `int64`, while the default integer dtype could be `int32` due to better target hardware support (e.g., GPUs).

Thoughts, @rgommers?

leofang: @kgryte I have a suggestion. How about removing all changes on the returned integer type, and moving the discussion to a new issue? I suppose the decision made here will affect all routines returning the indices of array elements satisfying a certain condition. On top of my head I know `argsort` is one of them (in fact it still says "default array index data type").

leofang: Did more homework about "array index type" and raised an issue in #319. I believe this was an overlook during a doc refactoring.

leofang: > How about removing all changes on the returned integer type, and moving the discussion to a new issue? This way the rest of this PR is unambiguous.

Cleaned up in f1557a1.

leofang:
I have reverted all changes on the returned integer type (discussion continued in #319) to make this PR easier to review 🙂

spec/API_specification/set_functions.md
kgryte: Thanks, @leofang!

spec/API_specification/set_functions.md
```diff
@@ -55,6 +55,47 @@ Each `nan` value should have a count of one, while the counts for signed zeros s
...
+Returns the unique elements of an input array `x` and the corresponding counts for each unique element in `x`.
+
+```{note}
+Uniqueness should be determined based on value equality (i.e., `x_i == x_j`). For input arrays having floating-point data types, value-based equality implies the following behavior.
+
+-   As `nan` values compare as `False`, `nan` values should be considered distinct.
+-   As `-0` and `+0` compare as `True`, signed zeros should not be considered distinct, and the corresponding unique element will be implementation-dependent (e.g., an implementation could choose to return `-0` if `-0` occurs before `+0`).
+
+As signed zeros are not distinct, using `inverse_indices` to reconstruct the input array is not guaranteed to return an array having the exact same values.
+
```
kgryte, line 76:
```suggestion
```

spec/API_specification/set_functions.md
```diff
@@ -55,6 +55,45 @@ Each `nan` value should have a count of one, while the counts for signed zeros s
...
+#### Parameters
+
+-   **x**: _&lt;array&gt;_
+
+    -   input array. If `x` has more than one dimension, the function must flatten `x` and return the unique elements of the flattened array.
+
+#### Returns
+
+-   **out**: _Tuple\[ &lt;array&gt;, &lt;array&gt;, &lt;array&gt;, &lt;array&gt; ]_
```
kgryte, line 86:
```suggestion
-   **out**: _Tuple\[ &lt;array&gt;, &lt;array&gt; ]_
```

kgryte (approved):
LGTM. Thanks, @leofang!

leofang:
Thanks, @kgryte, good catch!

PR #320, opened 2021-11-08
Machine readability fixes for operator special cases

honno:
`__truediv__` had a consistency issue, and `__pow__` was not updated with `xp.pow()` in #108. This does not conflict with #221.


kgryte (approved):
LGTM. Thanks, @honno!

Issue #318, opened 2021-11-05
axis argument to matrix_norm


asmeurer:
The matrix_norm function has an axis argument to specify the axes that the matrices are stacked on, defaulting to the last two axes. This is currently the only matrix function that allows specifying this. All other matrix functions assume matrices are stacked on the last two axes. To work with stacks on other axes, the array axes must first be transposed using something like `permute_dims`.

The axis/axes argument is present in vector_norm, cross, and tensordot, but the first two of those operate on 1-D vectors and the last operates on arbitrary tensors. The axis argument to matrix_norm is a holdover from the split of `norm`, which handled both matrix and vector norms at the same time.

kgryte:
@Lezcano You have thoughts on this?

kgryte:
This may be an instance where we want to remove the `axis` argument from the spec and consider its addition in the 2022 revision of the standard. This would not disbar PyTorch from continuing to support a `dim` kwarg, but would allow more time for array conforming libraries to access whether they would also prefer to support the ability to specify axes.

Lezcano:
I think @kgryte 's comment is on point. Removing things from the API and revisit these points in a later revision of the API is always easier.

kgryte:
Okay. Will submit a PR removing support for the `axis` argument. Thanks, @Lezcano, @leofang, and @asmeurer!

PR #321, opened 2021-11-08
Remove `axis` kwarg support from `linalg.matrix_norm`

kgryte:
This PR

-   resolves https://github.com/data-apis/array-api/issues/318 by removing support for the `axis` kwarg in `linalg.matrix_norm`. The current specification derives its signature from a previously unified `norm` API which had to cater to both vector and matrix use cases. In the current specification, `matrix_norm` is unique among `linalg` APIs in supporting an `axis` kwarg for an API operating on stacks of matrices. And as this API is relatively new among array libraries, seems most prudent to hold off including an `axis` kwarg until array libraries have come to a consensus that providing `axis` support is desirable. In general, easier to add to the standard, rather than take something away.

### DIFF
--- a/spec/extensions/linear_algebra_functions.md
+++ b/spec/extensions/linear_algebra_functions.md
@@ -261,7 +261,7 @@ Returns the multiplicative inverse of a square matrix (or a stack of square matr
 Alias for {ref}`function-matmul`.
 
 (function-linalg-matrix-norm)=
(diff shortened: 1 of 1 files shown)


kgryte:
cc @asmeurer for visibility.

kgryte:
Will merge. We can revisit the inclusion of an `axis` kwarg in the 2022 revision of the standard.

PR #322, opened 2021-11-08
Update guidance for default data types

kgryte:
This PR

-   resolves https://github.com/data-apis/array-api/issues/319 by updating guidance for default data types. Namely, this PR  adds guidance for a default array index data type which can be specified independently of the default integer data type.

PR #323, opened 2021-11-08
Update return data types for those APIs returning indices and counts

kgryte:
This PR 

-   is a follow-up PR to https://github.com/data-apis/array-api/pull/322, which adds guidance for conforming implementations to specify a "default array index data type". This PR subsequently updates the return data types for those APIs returning indices and/or counts (e.g., `unique*`, `argsort`, `argmin`, `argmax`, `nonzero`).

**Note**: the return types for APIs listed above which are not a `unique*` variant are already spec'd to return the default array index data type.

### DIFF
--- a/spec/API_specification/set_functions.md
+++ b/spec/API_specification/set_functions.md
@@ -47,9 +47,9 @@ Each `nan` value should have a count of one, while the counts for signed zeros s
     -   a namedtuple `(values, indices, inverse_indices, counts)` whose
 
         -   first element must have the field name `values` and must be an array containing the unique elements of `x`. The array must have the same data type as `x`.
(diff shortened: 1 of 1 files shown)


rgommers (changes requested):
There should be some more functions that return indices. `argsort` comes to mind for example, and `unique_inverse`. Can you review all APIs?

kgryte:
@rgommers Those are already spec'd to return the default array index data type. :)

kgryte:
As only the `unique*` APIs need to be updated, will go ahead and merge.

leofang:
I think nonzero is another function.

kgryte:
@leofang [`nonzero`](https://github.com/data-apis/array-api/blob/7525316eb26ac382dbed45c10d122703bdfa03d7/spec/API_specification/searching_functions.md#nonzerox-) is already spec'd to return the default array index data type.

PR #328, opened 2021-11-09
Fix `matmul` broadcasting guidance

kgryte:
This PR

-   addresses errors in `matmul` broadcasting guidance. Namely, only non-matrix dimensions should be broadcasted. Previous guidance suggests that matrix dimensions should also broadcast, but this is not correct (e.g., `(10, 2, 1) @ (10, 2, 2)` should not work). This is discussed further in [PEP 465](https://www.python.org/dev/peps/pep-0465/).


asmeurer (approved):

kgryte:
Thanks, @asmeurer!

PR #329, opened 2021-11-09
Add floating-point special cases for `floor_divide`

kgryte:
This PR

-   adds guidance for special cases involving floating-point operands in `floor_divide`. The lack of guidance was initially raised in [gh-199](https://github.com/data-apis/array-api/issues/199) (see [comment](https://github.com/data-apis/array-api/issues/199#issuecomment-909708561)). Upon investigation, NumPy's `floor_divide` is both inconsistent with `floor(divide(x,y))` behavior and built-in Python behavior and also diverges from behavior found in other libraries (e.g., TensorFlow). Accordingly, this PR seeks to clarify behavior for floating-point operands and ensure consistency with `floor(divide(x,y))` to avoid downstream user surprises. The documented special cases largely follow those of `divide`.
-   allows for array libraries to deviate in certain special cases involving infinities in order to match Python behavior.

### DIFF
--- a/spec/API_specification/array_object.md
+++ b/spec/API_specification/array_object.md
@@ -635,7 +635,7 @@ For input arrays which promote to an integer data type, the result of division b
 #### Special Cases
 
 ```{note}
-Floor division was introduced in Python via [PEP 238](https://www.python.org/dev/peps/pep-0238/) with the goal to disambiguate "true division" (i.e., computing an approximation to the mathematical operation of division) from rounding the result of division toward negative infinity (i.e., "floor division"). The former was computed when one of the operands was a `float`, while the latter was computed when both operands were `int`s. Overloading the `/` operator to support both behaviors led to subtle numerical bugs when integers are possible, but not expected.
+Floor division was introduced in Python via [PEP 238](https://www.python.org/dev/peps/pep-0238/) with the goal to disambiguate "true division" (i.e., computing an approximation to the mathematical operation of division) from "floor division" (i.e., rounding the result of division toward negative infinity). The former was computed when one of the operands was a `float`, while the latter was computed when both operands were `int`s. Overloading the `/` operator to support both behaviors led to subtle numerical bugs when integers are possible, but not expected.
 
 To resolve this ambiguity, `/` was designated for true division, and `//` was designated for floor division. Semantically, floor division was [defined](https://www.python.org/dev/peps/pep-0238/#semantics-of-floor-division) as equivalent to `a // b == floor(a/b)`; however, special floating-point cases were left ill-defined.
 
--- a/spec/API_specification/array_object.md
+++ b/spec/API_specification/array_object.md
@@ -632,6 +632,33 @@ Evaluates `self_i // other_i` for each element of an array instance with the res
 For input arrays which promote to an integer data type, the result of division by zero is unspecified and thus implementation-defined.
 ```
 
+#### Special Cases
+
+For floating-point operands, let `self` equal `x1` and `other` equal `x2`.
+
+-   If either `x1_i` or `x2_i` is `NaN`, the result is `NaN`.
+-   If `x1_i` is either `+infinity` or `-infinity` and `x2_i` is either `+infinity` or `-infinity`, the result is `NaN`.
+-   If `x1_i` is either `+0` or `-0` and `x2_i` is either `+0` or `-0`, the result is `NaN`.
+-   If `x1_i` is `+0` and `x2_i` is greater than `0`, the result is `+0`.
+-   If `x1_i` is `-0` and `x2_i` is greater than `0`, the result is `-0`.
+-   If `x1_i` is `+0` and `x2_i` is less than `0`, the result is `-0`.
+-   If `x1_i` is `-0` and `x2_i` is less than `0`, the result is `+0`.
+-   If `x1_i` is greater than `0` and `x2_i` is `+0`, the result is `+infinity`.
+-   If `x1_i` is greater than `0` and `x2_i` is `-0`, the result is `-infinity`.
+-   If `x1_i` is less than `0` and `x2_i` is `+0`, the result is `-infinity`.
+-   If `x1_i` is less than `0` and `x2_i` is `-0`, the result is `+infinity`.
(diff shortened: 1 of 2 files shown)


spec/API_specification/array_object.md
```diff
@@ -632,6 +632,33 @@ Evaluates `self_i // other_i` for each element of an array instance with the res
...
+-   If `x1_i` is less than `0` and `x2_i` is `+0`, the result is `-infinity`.
+-   If `x1_i` is less than `0` and `x2_i` is `-0`, the result is `+infinity`.
+-   If `x1_i` is `+infinity` and `x2_i` is a positive (i.e., greater than `0`) finite number, the result is `+infinity`.
+-   If `x1_i` is `+infinity` and `x2_i` is a negative (i.e., less than `0`) finite number, the result is `-infinity`.
+-   If `x1_i` is `-infinity` and `x2_i` is a positive (i.e., greater than `0`) finite number, the result is `-infinity`.
+-   If `x1_i` is `-infinity` and `x2_i` is a negative (i.e., less than `0`) finite number, the result is `+infinity`.
+-   If `x1_i` is a positive (i.e., greater than `0`) finite number and `x2_i` is `+infinity`, the result is `+0`.
+-   If `x1_i` is a positive (i.e., greater than `0`) finite number and `x2_i` is `-infinity`, the result is `-0`.
+-   If `x1_i` is a negative (i.e., less than `0`) finite number and `x2_i` is `+infinity`, the result is `-0`.
+-   If `x1_i` is a negative (i.e., less than `0`) finite number and `x2_i` is `-infinity`, the result is `+0`.
```
asmeurer, line 657: Just want to confirm that we want to deviate from NumPy on these (except for the `+0` ones).

```py
>>> np.floor_divide(np.inf, 1.)
nan
>>> np.floor_divide(np.inf, -1.)
nan
>>> np.floor_divide(-np.inf, 1.)
nan
>>> np.floor_divide(-np.inf, -1.)
nan
>>> np.floor_divide(1, np.inf)
0.0
>>> np.floor_divide(1, -np.inf)
-1.0
>>> np.floor_divide(-1, np.inf)
-1.0
>>> np.floor_divide(-1, -np.inf)
0.0
```

asmeurer:
These imported into the test suite just fine. I ran the tests and it found that NumPy deviates on those I mentioned.

spec/API_specification/array_object.md
kgryte: IMO, yes, we should deviate from NumPy here. I think you are correct when you stated that deviating from `floor(divide(...))` is surprising and certain results in the above are very unintuitive. E.g.,

```python
>>> np.floor_divide(1, -np.inf)
-1.0
>>> np.floor_divide(-1, np.inf)
-1.0
```

However, it seems that NumPy's behavior is intentional. Namely, it attempts to be complementary to `remainder` and abide by the relation

```
b = a % b + b * (a // b)
```

https://github.com/numpy/numpy/pull/7258/files#diff-7355537178d1b6f85a40acf970ac3317b184721c75828f33ed136ea0f6c92ae9R282 changed `floor_divide` to no longer use `floor( a / b )`, but to use `divmod`.

The above relation fails for the aforementioned special cases, except for `np.floor_divide( -1, np.inf )`.

I am not of the opinion that the spec should follow the above relation, as NumPy attempts to do, but perhaps @rgommers has other thoughts.

As the OP states, libraries diverge in what they return, so this is not a matter of simply going with consensus. E.g., TF seems to abide by the behavior proposed in this PR.

kgryte:
Thanks for reviewing and confirming @asmeurer!

spec/API_specification/array_object.md
seberg: In case nobody realized and it matters: NumPy follows Python.

kgryte: @seberg Indeed, that was mentioned here: https://github.com/data-apis/array-api/issues/199#issuecomment-964620063.

kgryte: Prior to the NumPy's PR 7258, NumPy seemed to follow PEP 238's ["semantics of floor division"](https://www.python.org/dev/peps/pep-0238/#semantics-of-floor-division). Unfortunately, PEP 238 is a bit light on explaining any rationale or expected behavior for floating-point edge cases.

kgryte: Based on my reading of this [discussion](https://bugs.python.org/issue22444), `floor_divide` is (perhaps obviously) ill-suited for floating-point operands.

In the spec, [`floor_divide`](https://github.com/data-apis/array-api/blob/02fae2cb00449c0e8bfbb27aa5cf707e5658f34a/spec/API_specification/elementwise_functions.md#floor_dividex1-x2-) is expected to return an array having a floating-point dtype when provided floating-point operands. This, perhaps, diverges from the historical origin story of `//` which was to return `int(floor(a/b))` (at least as per [PEP 3141](https://www.python.org/dev/peps/pep-3141/)). This stated, when providing floating-point operands to `//`, Python does return a float.

In short, I think there seems to be some historical baggage associated with `//` which this spec should avoid in order to ensure consistent behavior among APIs within the spec. `floor_divide` is currently included in the spec in order to mirror the equivalent dunder method (per specification conventions). I would expect that end users would be less surprised to see the spec diverge from Python built-in behavior, than to discover divergent results when using `floor(divide(...))` via spec-defined APIs.

seberg: Cool then. Yeah, it seems plausible that the Python behaviour just stems from divmod but the corner cases were never vetted (no idea).  I wonder if it is worth pushing it to the Python issue tracker, but dunno.

rgommers: Can we just add a note to the docstring that says something along the lines of "note, `floor_divide` wasn't really designed for floating-point input and therefore may have discrepencies for these special cases"?

kgryte: I've added a note providing context for discrepancies and then added notes to each special case which may deviate from expected/specified behavior.

kgryte:
The changes introduced in this PR have been reviewed and were discussed at the consortium meeting on 18 November 2021. Will merge, and any subsequent edits can be made in follow-up PRs.

@asmeurer @honno You'll want to note that array libraries are allowed to deviate from behavior included in the spec for a subset of the special cases. These special cases are demarcated accordingly.

PR #330, opened 2021-11-10
Update guidance for `pow` and `divide` concerning arrays with integer dtypes

kgryte:
This PR

-   supersedes https://github.com/data-apis/array-api/pull/221
-   adds guidance for `pow` and `divide` for arrays having integer data types, as discussed in https://github.com/data-apis/array-api/pull/173.
-   adds a note that, when providing arrays having integer data types to `pow`, the result of `x1_i**x2_i` is implementation-defined for `x2_i < 0` (e.g., `5**(-1)` as the result is a fraction).
-   specifies that the result of "true division" be an array having a floating-point data type.
-   makes input data type guidance consistent between `__pow__`/`pow` and `__truediv__`/`divide`.

cc @asmeurer @rgommers


kgryte:
As the changes introduced in this PR have been discussed elsewhere, will go ahead and merge. If further refinements are necessary, they can be addressed in a follow-up PR.

PR #331, opened 2021-11-10
`expand_dims()` signature change (and slight doc fix)

honno:
In #167 @rgommers gave us the current `expand_dims()` signature:
```python
expand_dims(x, /, *, axis)
```
I believe it makes sense to remove the keyword-only requirement of `axis`, given it has no default value:
```python
expand_dims(x, /, axis)
```
This keeps in line with the related function `squeeze()`. Notably all other manipulation functions that have arguments after `*` have defaults. Personally I've never seen a keyword-only argument that has no default, and I'm wondering if this behaviour could confuse developers and users.


kgryte (approved):
LGTM. Thanks, @honno!

PR #335, opened 2021-11-15
Special case when NaNs present in statistical functions

honno:
Adds a special case for the two functions that any NaNs present in computed elements should result in NaN. Hopefully I've used the correct language here.

Per discussion with @kgryte, this is expected behaviour.
```python
>>> min([0., float("nan")])
0.0
>>> np.min(np.asarray([0., float("nan")]))
nan
>>> torch.min(torch.as_tensor([0., float("nan")]))
tensor(nan)
```

### DIFF
--- a/spec/API_specification/statistical_functions.md
+++ b/spec/API_specification/statistical_functions.md
@@ -26,7 +26,7 @@ When the number of elements over which to compute the maximum value is zero, the
 
 #### Special Cases
 
-Let `x_i` be an element over which to compute the maximum value. For floating-point operands,
+For floating-point operands,
 
 -   if `x_i` is `NaN`, the maximum value is `NaN` (i.e., `NaN` values propagate).
 
@@ -57,14 +57,10 @@ Calculates the arithmetic mean of the input array `x`.
 
 #### Special Cases
 
-For a floating-point input array `x`, let `N` equal the number of elements over which to compute the arithmetic mean and
+Let `N` equal the number of elements over which to compute the arithmetic mean.
 
 -   if `N` is `0`, the arithmetic mean is `NaN`.
-
-Let `x_i` be an element over which to compute the arithmetic mean. For floating-point operands,
-
 -   if `x_i` is `NaN`, the arithmetic mean is `NaN` (i.e., `NaN` values propagate).
-
 
 #### Parameters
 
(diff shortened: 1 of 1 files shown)


rgommers:
> Do note this does mean PyTorch is wrong.

I don't see that in the example. `tensor(nan)` (as a 0-D array) looks like the expected answer to me (reduction over 1-D array -> 0-D array). The `numpy.array_api` answer should match that, but it may not (?).

honno:
> > Do note this does mean PyTorch is wrong.
> 
> I don't see that in the example. `tensor(nan)` (as a 0-D array) looks like the expected answer to me (reduction over 1-D array -> 0-D array). The `numpy.array_api` answer should match that, but it may not (?).

Oops I was thinking of a different discussion (I didn't even read my own example, oops!)

honno:
@kgryte I suppose this special case applies to all statistical functions? If that's the case, I can update this PR.

spec/API_specification/statistical_functions.md
```diff
@@ -24,6 +24,10 @@ Calculates the maximum value of the input array `x`.
 When the number of elements over which to compute the maximum value is zero, the maximum value is implementation-defined. Specification-compliant libraries may choose to error, return a sentinel value (e.g., if `x` is a floating-point input array, return `NaN`), or return the minimum possible value for the input array `x` data type (e.g., if `x` is a floating-point array, return `-infinity`).
 ```
 
+#### Special Cases
+
+-   if `NaN` in elements over which to compute the maximum value, the result is `NaN`.
```
kgryte, line 29:
```suggestion
Let `x_i` be an element over which to compute the maximum value. For floating-point operands,

-   if `x_i` is `NaN`, the maximum value is `NaN` (i.e., `NaN` values propagate).
```

spec/API_specification/statistical_functions.md
```diff
@@ -88,6 +92,10 @@ Calculates the minimum value of the input array `x`.
 When the number of elements over which to compute the minimum value is zero, the minimum value is implementation-defined. Specification-compliant libraries may choose to error, return a sentinel value (e.g., if `x` is a floating-point input array, return `NaN`), or return the maximum possible value for the input array `x` data type (e.g., if `x` is a floating-point array, return `+infinity`).
 ```
 
+#### Special Cases
+
+-   if `NaN` in elements over which to compute the minimum value, the result is `NaN`.
```
kgryte, line 97: Same changes as proposed above, except `s/maximum/minimum`.

kgryte (changes requested):

kgryte:
@honno Yes, that is correct. `NaN` values should propagate for the other statistical functions. Feel free to update this PR. Thanks!

kgryte (approved):
LGTM. Thanks, @honno!

PR #338, opened 2021-11-16
Changes Implementation section of data-interchange to close #337

### LINKED ISSUE #337: data-interchange description of DLPack
The [implementation section](https://data-apis.org/array-api/latest/design_topics/data_interchange.html#implementation) of the `data_interchange` document contains some incorrect statements:

  1.  > The `__dlpack__` method will produce a `PyCapsule` containing a `DLPackManagedTensor`,  ....

The `DLPackManagedTensor` should be `DLManagedTensor` consistent with [dlpack.h](https://github.com/dmlc/dlpack/blob/main/include/dlpack/dlpack.h#L206) and the DLPack diagram.

I also think it would be beneficial to make it more precise:

> The `__dlpack__` method will produce a `PyCapsule` object containing a reference to `DLPackManagedTensor` struct,  ....

 2. > The consumer must set the `PyCapsule` name to `"used_dltensor"` , and call the `deleter` of the `DLPackManagedTensor` when it no longer needs the data.

The `DLPackManagedTensor` here also should be `DLManagedTensor`, but more importantly, the consumer should not be calling the `deleter` function. The `DLManagedTensor.deleter` must be called by the `PyCapsule_Destructor` function provided by the producer. 

I also would like to call the mandatory nature of capsule renaming from `"dltensor"` to `"used_dltensor"` into question. With proper memory management the capsule can be consumed more than once, creating multiple views into the same allocation. 

I would therefore like to propose to change that statement from a requirement to a recommendation.

oleksandr-pavlyk:
Changes to the implementation section of `data_interchange.md` per #337.

@fschlimb

### DIFF
--- a/spec/design_topics/data_interchange.md
+++ b/spec/design_topics/data_interchange.md
@@ -133,14 +133,15 @@ _Note that while this API standard largely tries to avoid discussing implementat
 _DLPack diagram. Dark blue are the structs it defines, light blue struct members, gray text enum values of supported devices and data types._
 
 The `__dlpack__` method will produce a `PyCapsule` containing a
-`DLPackManagedTensor`, which will be consumed immediately within
+`DLManagedTensor`, which will be consumed immediately within
 `from_dlpack` - therefore it is consumed exactly once, and it will not be
 visible to users of the Python API.
 
-The producer must set the PyCapsule name to ``"dltensor"`` so that it can
-be inspected by name.
-The consumer must set the PyCapsule name to `"used_dltensor"`, and call the
-`deleter` of the `DLPackManagedTensor` when it no longer needs the data.
+The producer must set the `PyCapsule` name to ``"dltensor"`` so that it can
+be inspected by name, and set `PyCapsule_Destructor` that calls the `deleter`
+of the `DLManagedTensor` when the capsule is no longer needed.
+The consumer may change the `PyCapsule` name, e.g. to `"used_dltensor"` to
+prevent the capsule from being consumed more than once.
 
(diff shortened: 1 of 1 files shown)


spec/design_topics/data_interchange.md
```diff
@@ -126,24 +126,38 @@ unnecessary so asynchronous execution is enabled.
...
-Note: the capsule names ``"dltensor"`` and `"used_dltensor"` must be statically
-allocated.
+The consumer must transer ownership of the `DLManangedTensor` from the
+capsule to its own object. It does so by renaming the capsule to
+`"used_dltensor"` to ensure that `PyCapsule_Destructor` will not get
+called (ensured if `PyCapsule_Destructor` calls `deleter` only for
+capsules whose name is `"dltensor"`), but the `deleter` of the
+`DLManagedTensor` will be called by the destructor of the consumer
+library object created to own the `DLManagerTensor` obtained from the
+capsule.
```
fschlimb, line 157: Why is the ref-count of the capsule not enough?

oleksandr-pavlyk: My understanding is because it is an interlibrary data exchange protocol, and if some libraries transfer ownership to their internal objects in their existing implementations, discrepancy in lifetime management approaches may result in double free.

leofang (approved):
LGTM, thanks @oleksandr-pavlyk!

kgryte:
Thanks for reviewing @leofang!

Will merge. If anything further needs updating, we can address in a follow-up PR. Thanks, @oleksandr-pavlyk!

PR #340, opened 2021-11-29
Update guidance for unsigned integer data types for `sum` and `prod`

kgryte:
This PR

-   updates the guidance for `sum` and `prod` when `x` has an unsigned integer data type with a smaller range of values than the default integer data type and the `dtype` kwarg is `None`. Namely, when providing a `uint8` array, the returned array should have a `uint32` data type when the default integer data type is `int32`. This is in contrast to current guidance which requires that the returned array have an `int32` data type. In short, the returned array should have a data type whose "signedness" matches the input array. Returning a signed integer array when providing an unsigned integer array is likely to violate user expectations.

The changes introduced in this PR match NumPy's current behavior.


honno (approved):
LGTM

kgryte:
Thanks for reviewing, @honno!

Will merge, as this was an initial oversight and aligns with current NumPy behavior and thus with established precedent.

PR #341, opened 2021-12-06
Add note concerning subnormal numbers

### LINKED ISSUE #339: Improve subnormal expectations
Currently the spec doesn't mention subnormal floats, so assumedly the IEEE 754 behaviour stands.

For typical CuPy builds subnormals are flushed to zero, as explained by @leofang in https://github.com/numpy/numpy/pull/18536#issuecomment-801150072.

> Just FYI. In CuPy we turn flush-to-zero on by default (via setting a compiler flag `-ftz=true`), see [floating.py#L56-L58](https://github.com/cupy/cupy/blob/eafd82bfb80b5e7061cfd2eebe56bb2e4de658c5/cupy/_math/floating.py#L56-L58). The reason is [there could be significant computational cost on accelerators like GPU](https://developer.nvidia.com/blog/cuda-pro-tip-flush-denormals-confidence/). So, I suspect that CuPy is not the only array library that does this.

It does seem bad if IEEE 754 subnormal behaviour is expected but will be seemingly forever violated by one of its adopters (when compiled with default settings). **So I wonder if subnormal behaviour could just be specified as out-of-scope.** What would be the ramifications?

Notably in #131 and numpy/numpy#18536 there was discussion of a `smallest_subnormal` property for [`finfo`](https://data-apis.org/array-api/latest/API_specification/data_type_functions.html#finfo-type), but it was seen as a pretty awkward fit. It might be useful if some kind of information could tell the user that subnormals are not supported, e.g. `smallest_subnormal == smallest_normal`.

kgryte:
This PR

-   resolves #339.
-   relaxes IEEE 754 requirements for supporting subnormal numbers.
-   stems from the need to allow conforming implementations to disable subnormal number support for performance and/or security reasons.

### DIFF
--- a/spec/API_specification/data_types.md
+++ b/spec/API_specification/data_types.md
@@ -50,14 +50,19 @@ IEEE 754 single-precision (32-bit) binary floating-point number (see IEEE 754-20
 
 IEEE 754 double-precision (64-bit) binary floating-point number (see IEEE 754-2019).
 
+```{note}
+IEEE 754-2019 requires support for subnormal (a.k.a., denormal) numbers, which are useful for supporting gradual underflow. However, hardware support for subnormal numbers is not universal, and many platforms (e.g., accelerators) and compilers support toggling denormals-are-zero (DAZ) and/or flush-to-zero (FTZ) behavior to increase performance and to guard against timing attacks.
+
+Accordingly, subnormal behavior is left unspecified and, thus, implementation-defined. Conforming implementations may vary in their support for subnormal numbers.
+```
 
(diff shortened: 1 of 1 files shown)


honno (approved):
LGTM. Happy to see this PR :sweat_smile:

kgryte:
Thanks, @honno!

kgryte:
The changes in this PR have been discussed elsewhere. Will merge. Any further changes can be addressed in follow-up PRs.

leofang:
cc: @oleksandr-pavlyk for vis

PR #342, opened 2021-12-08
Specify numeric arrays for `argmin` and `argmax`

honno:
Remains consistent with `min` and `max`.


kgryte (approved):
LGTM. Thanks, @honno!

PR #344, opened 2021-12-14
Remove shape parity for inverse indices in set functions

honno:
I still get confused with inverse indices, but I think the spec is wrong to state the returned array must have the same shape... how would you reconstruct the input array `x` otherwise heh. This follows NumPy behaviour (and indeed `numpy.array_api`).


honno:
Nevermind I misread this, it says same shape as `x` :facepalm:

Issue #355, opened 2021-12-31
linalg.vecdot and vecdot have (midly) diverging defaults.


Lezcano:
Their signatures are:
`linalg.vecdot(x1, x2, /, *, axis=None)`

`vecdot(x1, x2, /, *, axis=-1)`

I'd certainly prefer to settle for `axis=-1`.

rgommers:
Ouch, that seems like a bug to me, good catch @Lezcano.

asmeurer:
I think this was just an oversight https://github.com/data-apis/array-api/pull/281

kgryte:
This has been addressed in https://github.com/data-apis/array-api/pull/358.

PR #358, opened 2022-01-05
Fix signature for linalg.vecdot

kgryte:
This PR:

-   resolves https://github.com/data-apis/array-api/issues/355. [gh-281](https://github.com/data-apis/array-api/pull/281) failed to include updating the default for `axis` to `-1`.


kgryte:
As this PR addresses an oversight and the change simply makes consistent `linalg.vecdot` with `vecdot`, will go ahead and merge.

PR #362, opened 2022-01-07
Clarify that the results of division operations on integer array data types resulting in floating-point outputs is implementation-defined 

### LINKED ISSUE #361: Divide functions have ambiguous specification for integer inputs
[`divide()`](https://data-apis.org/array-api/latest/API_specification/elementwise_functions.html#divide-x1-x2) (and `__truediv__`/`__itruediv__`) allows `x1` and `x2` to be integer arrays, but the ouput is specified as a floating array determined by type promotion, which int types can't promote to.

The spec should either clarify behaviour for both mixed int/float array and two int array scenarios, or note they are out-of-scope similiar to [`pow()`](https://data-apis.org/array-api/latest/API_specification/array_object.html#pow-self-other).

honno:
Resolve #361 by noting any instance of int as out-of-scope.

### DIFF
--- a/spec/API_specification/array_object.md
+++ b/spec/API_specification/array_object.md
@@ -1249,6 +1249,10 @@ Element-wise results must equal the results returned by the equivalent element-w
 
 Evaluates `self_i / other_i` for each element of an array instance with the respective element of the array `other`.
 
+```{note}
+If one or both of `self` and `other` have integer data types, the result is unspecified and thus implementation-dependent.
+```
+
 #### Special Cases
 
 For floating-point operands, let `self` equal `x1` and `other` equal `x2`.
--- a/spec/API_specification/array_object.md
+++ b/spec/API_specification/array_object.md
@@ -1250,7 +1250,9 @@ Element-wise results must equal the results returned by the equivalent element-w
 Evaluates `self_i / other_i` for each element of an array instance with the respective element of the array `other`.
 
 ```{note}
-If one or both of `self` and `other` have integer data types, the result is unspecified and thus implementation-dependent. This is because type promotion between data type “kinds” (i.e. integer versus floating-point) is unspecified.
+If one or both of `self` and `other` have integer data types, the result is implementation-dependent. This is because type promotion between data type “kinds” (i.e. integer versus floating-point) is unspecified.
+
+Specification-compliant libraries may choose to raise an error or return an array containing the element-wise results. If an array is returned, it must have a floating-point data type.
 ```
 
 #### Special Cases
--- a/spec/API_specification/array_object.md
+++ b/spec/API_specification/array_object.md
@@ -1115,7 +1115,7 @@ Calculates an implementation-dependent approximation of exponentiation by raisin
 ```{note}
 If both `self` and `other` have integer data types, the result of `__pow__` when `other_i` is negative (i.e., less than zero) is unspecified and thus implementation-dependent.
 
-If `self` has an integer data type and `other` has a floating-point data type, behavior is implementation-dependent. This is because type promotion between data type “kinds” (i.e. integer versus floating-point) is unspecified.
+If `self` has an integer data type and `other` has a floating-point data type, behavior is implementation-dependent, as type promotion between data type "kinds" (e.g., integer versus floating-point) is unspecified.
 ```
 
 #### Special Cases
@@ -1250,9 +1250,9 @@ Element-wise results must equal the results returned by the equivalent element-w
 Evaluates `self_i / other_i` for each element of an array instance with the respective element of the array `other`.
 
 ```{note}
-If one or both of `self` and `other` have integer data types, the result is implementation-dependent. This is because type promotion between data type “kinds” (i.e. integer versus floating-point) is unspecified.
+If one or both of `self` and `other` have integer data types, the result is implementation-dependent, as type promotion between data type "kinds" (e.g., integer versus floating-point) is unspecified.
 
-Specification-compliant libraries may choose to raise an error or return an array containing the element-wise results. If an array is returned, it must have a floating-point data type.
+Specification-compliant libraries may choose to raise an error or return an array containing the element-wise results. If an array is returned, the array must have a floating-point data type.
 ```
 
 #### Special Cases
--- a/spec/API_specification/array_object.md
+++ b/spec/API_specification/array_object.md
@@ -1250,7 +1250,7 @@ Element-wise results must equal the results returned by the equivalent element-w
 Evaluates `self_i / other_i` for each element of an array instance with the respective element of the array `other`.
 
 ```{note}
-If one or both of `self` and `other` have integer data types, the result is unspecified and thus implementation-dependent.
+If one or both of `self` and `other` have integer data types, the result is unspecified and thus implementation-dependent (type promotion between data type “kinds” (integer versus floating-point) is unspecified).
 ```
 
 #### Special Cases
--- a/spec/API_specification/array_object.md
+++ b/spec/API_specification/array_object.md
@@ -1115,7 +1115,7 @@ Calculates an implementation-dependent approximation of exponentiation by raisin
 ```{note}
 If both `self` and `other` have integer data types, the result of `__pow__` when `other_i` is negative (i.e., less than zero) is unspecified and thus implementation-dependent.
 
-If `self` has an integer data type and `other` has a floating-point data type, behavior is implementation-dependent (type promotion between data type "kinds" (integer versus floating-point) is unspecified).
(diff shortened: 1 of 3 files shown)


asmeurer:
This is what I wanted to do in https://github.com/data-apis/array-api/pull/221, but it was rejected. I still think it's a good idea.

spec/API_specification/elementwise_functions.md
```diff
@@ -528,6 +528,10 @@ For floating-point operands,
 
 Calculates the division for each element `x1_i` of the input array `x1` with the respective element `x2_i` of the input array `x2`.
 
+```{note}
+If one or both of the input arrays have integer data types, the result is unspecified and thus implementation-dependent.
```
asmeurer, line 532: Don't know if this needs a note or not, but if we're doing this, we should make this say "floating-point" https://github.com/data-apis/array-api/pull/362/files#diff-c63388022ee724b37174b4191598444df80aacb366d1f63677da4d14784df654R566

asmeurer: Also for both divide and pow, if we do have a note would be clearer to have the note in the parameters section, so that it is right next to the part that says "numeric data types" or "floating-point data types".

rgommers (changes requested):
This should say why, otherwise it leaves the reader guessing. The `pow` note says:

_If `self` has an integer data type and `other` has a floating-point data type, behavior is implementation-dependent (type promotion between data type “kinds” (integer versus floating-point) is unspecified)._

I'd like to be a little more strict even, because there really are only two outcomes possible: a floating point array with the expected division result, or an exception. Doing Python 2 style division should be forbidden. So how about saying that explicitly in this note?

spec/API_specification/elementwise_functions.md
honno: > Also for both divide and pow, if we do have a note would be clearer to have the note in the parameters section, so that it is right next to the part that says "numeric data types" or "floating-point data types".

Ah I agree that would make sense. For divide functions this seems tricky though, because the note also talks about the result dtype. Splitting the notes kinda loses the connectedness of the input and output.

rgommers (approved):
This LGTM. Let's see if everyone else is happy too before merging.

spec/API_specification/elementwise_functions.md
```diff
@@ -528,6 +528,12 @@ For floating-point operands,
 
 Calculates the division for each element `x1_i` of the input array `x1` with the respective element `x2_i` of the input array `x2`.
 
+```{note}
+If one or both of the input arrays have integer data types, the result is implementation-dependent. This is because type promotion between data type “kinds” (i.e. integer versus floating-point) is unspecified.
+
+Specification-compliant libraries may choose to raise an error or return an array containing the element-wise results. If an array is returned, it must have a floating-point data type.
```
leofang, line 534: Ditto.

leofang (commented):
Leaving two comments below. I guess I find the fix suggested by @rgommers a bit confusing:

> I'd like to be a little more strict even, because there really are only two outcomes possible: a floating point array with the expected division result, or an exception. Doing Python 2 style division should be forbidden. So how about saying that explicitly in this note?

If the spec *requires* the returned array to be of floating-point type if not raising an error, then it is not fully "implementation defined". We should either 
1. leave it completely implementation defined 
2. or, remove the ambiguity by saying the implementation must pick a floating point type (but which to pick is up to the implementation)

rgommers:
> If the spec requires the returned array to be of floating-point type if not raising an error, then it is not fully "implementation defined". 
> ...
> remove the ambiguity by saying the implementation must pick a floating point type (but which to pick is up to the implementation)

+1 for this. Isn't that the same as what I said?

spec/API_specification/array_object.md
```diff
@@ -1249,6 +1249,12 @@ Element-wise results must equal the results returned by the equivalent element-w
 
 Evaluates `self_i / other_i` for each element of an array instance with the respective element of the array `other`.
 
+```{note}
+If one or both of `self` and `other` have integer data types, the result is implementation-dependent. This is because type promotion between data type “kinds” (i.e. integer versus floating-point) is unspecified.
+
+Specification-compliant libraries may choose to raise an error or return an array containing the element-wise results. If an array is returned, it must have a floating-point data type.
```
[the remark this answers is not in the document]
kgryte: Here the distinction of "result" is the actual values returned which is separate from the requirement that the output dtype must be floating-point. What this note is getting at is that, because implementations may vary in how they promote between different "kinds", what values are returned may also vary, but the output result dtype must be floating-point regardless.

spec/API_specification/array_object.md
```diff
@@ -1115,7 +1115,7 @@ Calculates an implementation-dependent approximation of exponentiation by raisin
 ```{note}
 If both `self` and `other` have integer data types, the result of `__pow__` when `other_i` is negative (i.e., less than zero) is unspecified and thus implementation-dependent.
 
-If `self` has an integer data type and `other` has a floating-point data type, behavior is implementation-dependent (type promotion between data type "kinds" (integer versus floating-point) is unspecified).
+If `self` has an integer data type and `other` has a floating-point data type, behavior is implementation-dependent. This is because type promotion between data type “kinds” (i.e. integer versus floating-point) is unspecified.
```
kgryte, line 1118:
```suggestion
If `self` has an integer data type and `other` has a floating-point data type, behavior is implementation-dependent, as type promotion between data type "kinds" (e.g., integer versus floating-point) is unspecified.
```

spec/API_specification/array_object.md
```diff
@@ -1249,6 +1249,12 @@ Element-wise results must equal the results returned by the equivalent element-w
 
 Evaluates `self_i / other_i` for each element of an array instance with the respective element of the array `other`.
 
+```{note}
+If one or both of `self` and `other` have integer data types, the result is implementation-dependent. This is because type promotion between data type “kinds” (i.e. integer versus floating-point) is unspecified.
```
kgryte, line 1253:
```suggestion
If one or both of `self` and `other` have integer data types, the result is implementation-dependent, as type promotion between data type "kinds" (e.g., integer versus floating-point) is unspecified.
```

spec/API_specification/array_object.md
kgryte, line 1255:
```suggestion
Specification-compliant libraries may choose to raise an error or return an array containing the element-wise results. If an array is returned, the array must have a floating-point data type.
```

spec/API_specification/elementwise_functions.md
```diff
@@ -528,6 +528,12 @@ For floating-point operands,
 
 Calculates the division for each element `x1_i` of the input array `x1` with the respective element `x2_i` of the input array `x2`.
 
+```{note}
+If one or both of the input arrays have integer data types, the result is implementation-dependent. This is because type promotion between data type “kinds” (i.e. integer versus floating-point) is unspecified.
```
kgryte, line 532:
```suggestion
If one or both of the input arrays have integer data types, the result is implementation-dependent, as type promotion between data type "kinds" (e.g., integer versus floating-point) is unspecified.
```

spec/API_specification/elementwise_functions.md
kgryte, line 534:
```suggestion
Specification-compliant libraries may choose to raise an error or return an array containing the element-wise results. If an array is returned, the array must have a floating-point data type.
```

spec/API_specification/elementwise_functions.md
```diff
@@ -1223,7 +1229,7 @@ Calculates an implementation-dependent approximation of exponentiation by raisin
 ```{note}
 If both `x1` and `x2` have integer data types, the result of `pow` when `x2_i` is negative (i.e., less than zero) is unspecified and thus implementation-dependent.
 
-If `x1` has an integer data type and `x2` has a floating-point data type, behavior is implementation-dependent (type promotion between data type "kinds" (integer versus floating-point) is unspecified).
+If `x1` has an integer data type and `x2` has a floating-point data type, behavior is implementation-dependent. This is because type promotion between data type “kinds” (i.e. integer versus floating-point) is unspecified.
```
kgryte, line 1232:
```suggestion
If `x1` has an integer data type and `x2` has a floating-point data type, behavior is implementation-dependent, as type promotion between data type "kinds" (e.g., integer versus floating-point) is unspecified.
```

kgryte (approved):
LGTM.

spec/API_specification/array_object.md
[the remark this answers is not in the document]
kgryte: I'll mark this conversation as resolved.

spec/API_specification/elementwise_functions.md
kgryte: Same as above.

kgryte:
Thanks, @honno!

PR #364, opened 2022-01-10
Prevent autosummary from using non-Python signature syntax in its tables

asmeurer:
As suggested here https://github.com/data-apis/array-api/pull/348#issuecomment-1004380206.


kgryte (approved):
LGTM.

PR #386, opened 2022-02-07
Add `copy` kwarg support to `reshape`

kgryte:
This PR

-   adds support for a `copy` kwarg to `reshape`. This has a similar specification as in `asarray`. The addition of a `copy` kwarg support helps address copy-view concerns and accommodates scenarios where returning a view is not possible. This change was discussed in the consortium meeting held on January 6, 2022.


honno (approved):
LGTM

kgryte:
Thanks, @honno!

kgryte:
As this change was agreed upon during the consortium meeting held on January 6, 2022, and the change matches the semantics of `asarray`, will merge. Any further changes can be addressed in follow-up PRs...

PR #387, opened 2022-02-07
Fix `axis` description in `expand_dims`

kgryte:
This PR

-   fixes the specification description of the `axis` kwarg in `expand_dims`. The current guidance indicates that `axis` should follow normal Python indexing rules, where a negative index counts back from the last dimension. This would indicate the `-1` would correspond to `N-1` where `N` is the rank of the input array. However, in order to ensure symmetry between positive and negative `axis` values for appending and prepending singleton dimensions, `-1` should resolve to appending a singleton dimension to the end of the array. In a sense, `axis` follows normal Python indexing rules, but for the output array which has an additional dimension and not the input array. This PR aims to clarify this ambiguity.

### DIFF
--- a/spec/API_specification/signatures/manipulation_functions.py
+++ b/spec/API_specification/signatures/manipulation_functions.py
@@ -29,7 +29,7 @@ def expand_dims(x: array, /, *, axis: int = 0) -> array:
     x: array
         input array.
     axis: int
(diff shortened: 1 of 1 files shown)


kgryte:
Thanks, @honno!

PR #388, opened 2022-02-07
Clarify expected results for `any` and `all` when operating on empty arrays

kgryte:
This PR

-   adds explicit guidance for `any` and `all` when operating on empty arrays. For the former, the test result should be `False` and, for the latter, the test result should be `True`. This matches both NumPy et al behavior (`np.any` and `np.all`), TF behavior (`tf.reduce_any` and `tf.reduce_all`), and Python behavior (`any` and `all`).
-   adds a note for both `any` and `all` that `+-inf` and `NaN` all evaluate to `True`.


honno (approved):
LGTM.

While you don't need to specifically note 0-sided dimensions (as the array would be empty anyway), I guess it probably is worth it.

kgryte:
Thanks, @honno!

PR #389, opened 2022-02-08
Disable the html_prettify setting

### LINKED ISSUE #139: Extra space after every inline code
In the rendered spec, there is a weird formatting issue. Every piece of inline code includes an extra space after it. For example, [here](https://data-apis.github.io/array-api/latest/API_specification/array_object.html#abs-x):

Screen Shot 2021-03-09 at 4 33 40 PM

It looks like:

- If `x_i `is `NaN `, the result is `NaN `.

instead of

- If `x_i` is `NaN`, the result is `NaN`.

If there is already a space after the code, it is rendered as code. If there isn't, one is added. The added spaces are also there if you copy-paste the text.

I'm not sure why this is happening, if it is some issue with Myst or our theme or something else.

asmeurer:
This fixes the issue where every code block would have an extra space inserted
after it.

Fixes #139.

### DIFF
--- a/spec/conf.py
+++ b/spec/conf.py
@@ -114,7 +114,7 @@ html_theme_options = {
     #'repo_name': 'Project',
 
     "html_minify": False,
-    "html_prettify": True,
+    "html_prettify": False,
     "css_minify": True,
     "logo_icon": "&#xe869",
     "repo_type": "github",


kgryte (approved):
LGTM.

PR #390, opened 2022-02-08
Make the Sphinx build error on warnings and fix all warnings

asmeurer:
Some warnings were legitimate cross-reference issues, and some are from autodoc wanting to create cross-references for every type hint. These have to be ignored using the `nitpick_ignore` variable in `conf.py`, since we don't actually have anything to reference for the given type hints.

### DIFF
--- a/spec/API_specification/array_object.rst
+++ b/spec/API_specification/array_object.rst
@@ -37,47 +37,47 @@ Arithmetic Operators
 
 A conforming implementation of the array API standard must provide and support an array object supporting the following Python arithmetic operators.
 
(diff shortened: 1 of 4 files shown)


kgryte (approved):
LGTM. Thanks, @asmeurer!

asmeurer:
By the way, ideally we would have `from __future__ import annotations` in all the files, but if you do this it triggers a bug in Sphinx, which is fixed by this unmerged PR https://github.com/sphinx-doc/sphinx/pull/9736. I need to double check but I think there is no difference in the rendering from doing this, so we can hold off on doing it for now.

PR #391, opened 2022-02-08
Use `Sequence` from `typing` instead of `collections.abc`

honno:
The `tensordot()` signatures in both the top-level and linalg extension signatures use `Sequence` for type hinting. They were previously imported from `collections.abc`, but they should be imported from `typing` as indeed we want the type hint as opposed to the abstract base class.

Correct typing here is necessary when importing the signatures directory as a Python module (otherwise you get `TypeError: 'ABCMeta' object is not subscriptable`), like we probably want to do in the test suite. cc @asmeurer


kgryte (approved):
LGTM. Thanks, @honno!

Issue #392, opened 2022-02-17
linspace behavior is underspecified for integer dtypes


kgryte:
The [specification](https://github.com/data-apis/array-api/blob/3c241a257ae4145f06b379da658817db51a852c1/spec/API_specification/signatures/creation_functions.py#L204) for `linspace` is currently underspecified for integer output array dtypes. Namely, the specification is silent concerning the following scenarios:

1.  what should happen `start` and `stop` are floating-point numbers with non-integral values and `dtype` is an integer data type?

    ```python
    linspace( 1.4, 100.75, num=50, dtype="int32" )
    ```

2.  what should happen when `start` and `stop` are Python `int`s and `dtype` is an integer data type, but the computed spacing between adjacent elements is not an integral value?

    ```python
    linspace( 0, 10, num=100, dtype="int32" )
    ```

For both scenarios, one option is to add a note to the specification stating that behavior in these instances is implementation-dependent--a conforming implementation would be free to round/truncate, raise an exception, or something else.

Otherwise, for (1), this could be explicitly disallowed in the specification; i.e., if `start` and `stop` are floating-point, then only floating-point `dtype`s are allowed.

For (2), this is somewhat trickier to resolve as the spacing between adjacent elements is an implicit variable which users may find difficult to reason about without explicitly computing the spacing and determining whether, say, a specific argument combination might trigger an exception. However, we could opt to be pedantic and only allow integer `dtype`s when `start`, `stop`, and the computed spacing are **all** integer values. Otherwise, an exception should be raised.

A final option is that `linspace` could be restricted to only floating-point data types. If a user wants to return an output array with an integer data type, they could use [`arange`](https://github.com/data-apis/array-api/blob/3c241a257ae4145f06b379da658817db51a852c1/spec/API_specification/signatures/creation_functions.py#L4) and explicitly specify the `step`.

## Prior Art

-   NumPy's [`linspace`](https://numpy.org/doc/stable/reference/generated/numpy.linspace.html) floors non-integral values when the output array data type is an integer `dtype`.

    ```python
    >>> np.linspace(1.0,10,21,dtype="int32")                                                            
    array([ 1,  1,  1,  2,  2,  3,  3,  4,  4,  5,  5,  5,  6,  6,  7,  7,  8,
        8,  9,  9, 10], dtype=int32)
    ```

-   TensorFlow's [`linspace`](https://tensorflow.google.cn/api_docs/python/tf/linspace) only supports floating-point data types.

[...]

PR #393, opened 2022-02-21
Restrict `linspace` support to floating-point data types

kgryte:
This PR

-   resolves https://github.com/data-apis/array-api/issues/392 in accordance with discussions during the 2022-02-17 consortium meeting. Namely, `linspace` should be limited to floating-point output data types, unless a conforming implementation has compelling reasons for supporting integer output data types (e.g., due to backward compatibility). However, integer output behavior is left implementation-defined and should not, in general, be considered portable.
-   adds a note concerning the conversion of `int` start and stop values to floating-point. Namely, that overflow behavior is implementation-defined. This follows similar guidance elsewhere in the specification.


kgryte:
cc @honno

honno (approved):
LGTM. Good shout on noting non-representable `start` and `stop` args.

kgryte:
As the changes in this PR are straightforward and have been discussed in consortium meetings, will go ahead and merge so that changes, if any, can be updated in the test suite.

PR #406, opened 2022-03-23
Make `logaddexp()` arguments pos-only

honno:


rgommers (approved):
LGTM, seems to have been an oversight. Thanks @honno.

Issue #360, opened 2022-01-06
Allowing `None` for indexing to add an axis


jakirkham:
In many array libraries today, one can use `None` when indexing to add a singleton dimension like `np.newaxis`. Would be good to support in the standard as well.

rgommers:
Thanks @jakirkham. I think this one fell through the cracks so far. `x[:, None, :]` is equivalent to `expand_dims(x, axis=1)`, the latter is in the standard so at least there's an easy alternative.

I just tested and it looks like every library supports indexing with `None` with consistent behavior. So we should just add this I think.

asmeurer:
We should probably also add `newaxis = None` to the namespace.

kgryte:
I've opened https://github.com/data-apis/array-api/pull/408 which adds support for expanding dimensions via `None`.

PR #408, opened 2022-03-24
Add support for expanding dimensions using `None`

kgryte:
This PR

-   resolves https://github.com/data-apis/array-api/issues/360 by adding support for expanding dimensions via `None`. The functional equivalent of the proposed changes is repeated invocation of `expand_dims`. Using `None` in a selection tuple is more convenient when needing to insert multiple size `1` dimensions.
-   as per [comment](https://github.com/data-apis/array-api/issues/360#issuecomment-1075608356), using `None` in selection tuples enjoys broad support across array libraries.

### DIFF
--- a/spec/API_specification/indexing.rst
+++ b/spec/API_specification/indexing.rst
@@ -137,7 +137,7 @@ Multi-dimensional arrays must extend the concept of single-axis indexing to mult
 
 - Providing a single negative integer ``i`` as a single-axis index must index the same elements as the slice ``n+i:n+i+1``, where ``n`` is the axis (dimension) size.
 
-- Providing a single integer as a single-axis index must reduce the number of array dimensions by ``1`` (i.e., the array rank should decrease by one; if ``A`` has rank ``2``, ``rank(A)-1 == rank(A[0, :])``). In particular, a selection tuple with the ``m``\th element an integer (and all other entries ``:``) indexes a sub-array with rank ``N-1``.
+- Providing a single integer as a single-axis index must reduce the number of array dimensions by ``1`` (i.e., the array rank must decrease by one; if ``A`` has rank ``2``, ``rank(A)-1 == rank(A[0, :])``). In particular, a selection tuple with the ``m``\th element an integer (and all other entries ``:``) indexes a sub-array with rank ``N-1``.
 
   .. note::
     When providing a single integer as a single-axis index to an array of rank ``1``, the result should be an array of rank ``0``, not a NumPy scalar. Note that this behavior differs from NumPy.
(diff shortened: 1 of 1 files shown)


rgommers (commented):
I think adding `None` makes sense. My main question is whether it's okay to combine it with other indexing semantics. E.g,:
```python
>>> import torch
>>> t = torch.arange(5)
>>> t[None, :3, None]
tensor([[[0],
         [1],
         [2]]])
```
That probably needs testing and specifying.

asmeurer:
> I think adding None makes sense. My main question is whether it's okay to combine it with other indexing semantics. E.g,:

There shouldn't be any issues with this I would imagine. The text from the [NumPy docs](https://numpy.org/doc/stable/user/basics.indexing.html) seems fine, although it could perhaps be more precise:

> Each [newaxis](https://numpy.org/doc/stable/reference/constants.html#numpy.newaxis) object in the selection tuple serves to expand the dimensions of the resulting selection by one unit-length dimension. The added dimension is the position of the [newaxis](https://numpy.org/doc/stable/reference/constants.html#numpy.newaxis) object in the selection tuple. [newaxis](https://numpy.org/doc/stable/reference/constants.html#numpy.newaxis) is an alias for None, and None can be used in place of this with the same result.

This should probably not apply to boolean indices where we should keep the "only index" rule.

spec/API_specification/indexing.rst
```diff
@@ -158,6 +158,11 @@ Multi-dimensional arrays must extend the concept of single-axis indexing to mult
 
     To perform flat indexing, use ``reshape(x, (-1,))[integer]``.
 
+- Each ``None`` in the selection tuple must expand the dimensions of the resulting selection by one dimension of size ``1``. The position of the added dimension must be the same as the position of ``None`` in the selection tuple.
```
asmeurer, line 161: Do we need to clarify that None is not a "single-axis" index for the purposes of the other rule (like the one below this one)?

kgryte: I've updated the guidance for the `IndexError` rule. This should now be resolved.

kgryte:
I've added an explicit note that the use of `None` in combination with boolean array indexing is not allowed. This should address @asmeurer's comment above.

kgryte:
Re: combining with other indexing semantics. As @asmeurer mentions, this should not be an issue, given that the use of `None` is syntactic sugar for first extracting a selection and then applying `expand_dims`. Meaning, an implementation could filter a selection tuple for non-`None` values, compute the selection, and then apply `expand_dims` to the result. As such, there should be no conflict with using `None` in conjunction with other single-axis indexing expressions.

asmeurer:
This looks good. Were we going to add `newaxis`?

kgryte:
As this PR has been open for 3 weeks, has received approvals here, and no objections were raised during consortium meetings, will go ahead and merge. Any changes can be addressed in follow-up pull requests.

PR #409, opened 2022-03-24
Clarify overflow behavior in `asarray`

kgryte:
This PR

-   adds a note to `asarray` concerning what happens when an input value exceeds the precision of the resolved output array data type. Namely, behavior is left unspecified and, thus, implementation-defined. This follows similar guidance as found in `full`.
-   clarifies in `full` and `full_like` that overflow behavior is implementation-defined even when `dtype` is not `None`.


honno (approved):
LGTM

rgommers (approved):
Makes sense, thanks @kgryte. I'll leave it open for a while to give people time to comment.

oleksandr-pavlyk (approved):

kgryte:
As this PR has received 3 approvals, will merge. Any further updates can be addressed in follow-up PRs...

PR #412, opened 2022-03-28
Fix typo in purpose and scope history section

cnpryer:
Hi! Small typo in the Purpose and Scope under the [History section](https://data-apis.org/array-api/latest/purpose_and_scope.html#history).


rgommers (approved):
Thanks @cnpryer

PR #413, opened 2022-04-05
`expm1` should having float inputs

honno:
Was `expm1()` accepting int dtypes an oversight? If so, here's a fix!

If not, I'm guessing the intention was to "allow" int dtypes here, but leave it as implementation-defined behaviour. In that case we'll need to add a note ala [`divide()`](https://data-apis.org/array-api/latest/API_specification/generated/signatures.elementwise_functions.divide.html).

### DIFF
--- a/spec/API_specification/signatures/elementwise_functions.py
+++ b/spec/API_specification/signatures/elementwise_functions.py
@@ -562,7 +562,7 @@ def expm1(x: array, /) -> array:
     Parameters
     ----------
     x: array
-        input array. Should have a numeric data type.
+        input array. Should have a floating-point data type.
 
     Returns
     -------


asmeurer:
Presumably this was just a typo. `expm1` should be no different from `exp`. And it looks like the test suite and numpy implementation are already correct here.

asmeurer:
All the floating-point functions can accept integer inputs if the library wants to. I don't think we need such a note. I think divide only gets a note because it's not obvious whether the spec requires integer division on integer inputs or not (compared to something like `exp` where it would be pointless to support integer inputs unless you did a float promotion).

kgryte:
This appears to have been an oversight. As such, will merge...

PR #414, opened 2022-04-07
Add `newaxis` constant for use in array indexing

kgryte:
This PR

-   adds the constant `newaxis` to the specification, as an alias for `None`. This change was discussed and supported in the March 24, 2022 consortium meeting.
-   addresses readability concerns when performing array indexing (e.g., `A[:, newaxis]` vs `A[:, None]`, where the former more explicitly conveys that the selection tuple results in an array with an additional dimension).

## Notes

`newaxis` as an alias for `None` enjoys broad support among array libraries. Only PyTorch and Dask lack explicit support.

-   NumPy: `np.newaxis`
-   PyTorch: no equivalent, but supports `None`
-   MXNet: `mx.np.newaxis`
-   TensorFlow: `tf.newaxis`
-   CuPy: `cupy.newaxis`
-   Dask: no equivalent, but supports `np.newaxis` (i.e., `None`)

### DIFF
--- a/spec/API_specification/constants.rst
+++ b/spec/API_specification/constants.rst
@@ -22,4 +22,5 @@ Objects in API
    e
    inf
    nan
+   newaxis
    pi
--- a/spec/API_specification/signatures/constants.py
+++ b/spec/API_specification/signatures/constants.py
@@ -15,6 +15,11 @@ nan = float('nan')
 IEEE 754 floating-point representation of Not a Number (``NaN``).
 """
 
+newaxis = None
+"""
+An alias for ``None`` which is useful for indexing arrays.
+"""
+
 pi = 3.141592653589793
 """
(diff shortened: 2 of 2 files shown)


asmeurer:
This looks good, and is absolutely trivial for libraries to implement. Obviously this needs to be merged alongside https://github.com/data-apis/array-api/pull/408. Is the plan to defer this until after the spec release?

kgryte:
@asmeurer As discussed during meetings, we'll merge this for inclusion in the current specification version given that indexing with `None` is already universally supported across array libraries.

Most libraries already have a `newaxis` constant, except Dask and Torch. Torch still needs to add all currently specified constants. Dask has plans to add all constants (https://github.com/dask/dask/issues/8931).

Accordingly, now that https://github.com/data-apis/array-api/pull/408 is merged, we can merge this PR, as well, for inclusion in the current specification version.

Issue #177, opened 2021-05-06
Proposal: add APIs for getting and setting elements via a list of indices (i.e., `take`, `put`, etc)


kgryte:
## Proposal

Add APIs for getting and setting elements via a list of indices.

## Motivation

Currently, the array API specification does not provide a direct means of extracting and setting a list of elements along an axis. Such operations are relatively common in NumPy usage either via "fancy indexing" or via explicit `take` and `put` APIs.

Two main arguments come to mind for supporting at least basic `take` and `put` APIs:

1.  Indexing does not currently support providing a list of indices to index into an array. The principal reason for not supporting fancy indexing stems from dynamic shapes and compatibility with accelerator libraries. However, use of fancy indexing is relatively common in NumPy and similar libraries where dynamically extracting rows/cols/values is possible and can be readily implemented.

2.  Currently, the output of a subset of APIs currently included in the standard cannot be readily consumed without manual workarounds if a specification-conforming library implemented only the APIs in the standard. For example,

    -   [`argsort`](https://github.com/data-apis/array-api/blob/2e759ef79fe52542044bf075e29a31f4eadbd8a5/spec/API_specification/sorting_functions.md#argsortx---axis-1-descendingfalse-stabletrue) returns an array of indices. In NumPy, the output of this function can be consumed by `put_along_axis` and `take_along_axis`.
    -   [`unique`](https://github.com/data-apis/array-api/blob/main/spec/API_specification/set_functions.md#uniquex---return_countsfalse-return_indexfalse-return_inversefalse) can return an array of indices if `return_index` is `True`.

## Background

The following table summarizes library implementations of such APIs:

| op | NumPy | CuPy | Dask | MXNet | Torch | TensorFlow |
| -- | -------- | ----- | ----- | ------- | ------ | ----------- |
| extracting elements along axis | [`take`][numpy-take] | [`take`][cupy-take] | [`take`][dask-take] | [`take`][mxnet-take] | [`take`][torch-take]/[`gather`][torch-gather] | [`gather`][tf-gather]/[`numpy.take`][tf-numpy-take] |
| setting elements along axis | [`put`][numpy-put] | [`put`][cupy-put] | -- | -- | [`scatter`][torch-scatter] | [`scatter_nd`][tf-scatter-nd]/[`tensor_scatter_nd_update`][tf-tensor-scatter-nd-update] |
[...]

PR #416, opened 2022-04-07
Add `take` specification for returning elements of an array along a specified axis

kgryte:
This PR

-   adds a specification for `take`, an API for returning elements of an array along a specified axis. This API was initially proposed in https://github.com/data-apis/array-api/issues/177.
-   supports only a single keyword argument: `axis`. This is derived from [comparison data](https://github.com/data-apis/array-api-comparison/blob/eaa75810fa287bd7a8813f6910fd81f7313f8fc4/signatures/indexing/take.md).
-   requires that the `axis` keyword argument always be provided, as suggested in [comment](https://github.com/data-apis/array-api/issues/177#issuecomment-1078481030), but only when the input array has more than one dimension.
-   does not currently support providing a multi-dimensional array of indices. Array libraries supporting multi-dimensional arrays for `indices` treat the input array as a flattened 1-dimensional array. If we were to follow similarly, we'd need to specify the flattened layout.
-   does not currently support lists/tuples/sequences for the `indices` argument. NumPy et al allow array-like. Torch requires `LongTensor`. Was not immediately clear to me how accepting we should be for non-arrays, so went with the most conservative choice.
-   does not support an integer for the `indices` argument. Instead, users should use normal slice/indexing operations.

## Questions

-   Should the API be extended to support multi-dimensional `indices`? **No, not yet.**
-   Should the `axis` argument be required? **Yes, but only for multi-dimensional `x`.**
-   Should the `indices` argument be more liberal in the types of allowed arguments? **No, should be restricted to only `array` values.**


honno:
Should this PR be blocked until an initial spec release?

asmeurer:
Regarding `axis` and flattening, it's worth noting `np.put` doesn't have `axis`, so requiring it here would either break the symmetry with a potential `put` function, or require `put` to be modified from NumPy (see https://github.com/data-apis/array-api/issues/177#issuecomment-1078482747).

shoyer:
Two thoughts:

1. We could make `axis` optional only in the case of 1D arrays
2.  I would not necessarily be limited by constraints on `np.put`, which in my opinion is much more niche functionality than `take`. I don't even know if we want `put` in the API standard.

kgryte:
@rgommers Before moving forward with this PR, we need to ensure buy-in from PyTorch. The current proposal deviates from supported PyTorch [behavior](https://pytorch.org/docs/stable/generated/torch.take.html) in the following ways:

1.  PyTorch does not currently support an `axis` kwarg.
2.  PyTorch always flattens the input array.

We'd need to ensure that PyTorch is (1) okay with adding an `axis` kwarg to support taking along a specified axis and (2) making an `axis` kwarg mandatory, thus breaking current behavior (i.e., based on the current proposal, no way to index into a flattened array using `take` which is the current PyTorch behavior).

For NumPy-inspired libraries, the `axis` kwarg is supported and, provided the spec-compliant API is in a sub-namespace, `xp.take(x, indices, axis=-1)` can simply wrap a call to `np.take`. Dask already effectively requires an `axis` kwarg, as the [default](https://github.com/data-apis/array-api-comparison/blob/02bffff439b766984ec86095827d3e6e51dd5847/signatures/indexing/take.md#daskarray) `axis` value is `0`.

Accordingly, PyTorch is the odd one out here, atm, so would be good to get a feel for potential specification constraints in order to ensure broad API compatibility, while also addressing fancy indexing concerns (indexing arrays via arrays) and the [proposal](https://github.com/data-apis/array-api/issues/177#issuecomment-1078481030) to require an `axis`.

rgommers:
> does not currently support lists/tuples/sequences for the `indices` argument. NumPy et al allow array-like. Torch requires `LongTensor`. Was not immediately clear to me how accepting we should be for non-arrays, so went with the most conservative choice.

As a rule we don't allow array-like inputs in the array API standard, because (a) it's ill-defined, and (b) it turns out not to be the best idea to allow it anyway. So it's fine to be consistent with that here.

> We could make `axis` optional only in the case of 1D arrays

This makes sense to me. And seems useful, based on my analysis of how scikit-learn uses `np.take`.

> The current proposal deviates from supported PyTorch [behavior](https://pytorch.org/docs/stable/generated/torch.take.html) in the following ways:
> 
>     1. PyTorch does not currently support an `axis` kwarg.
> 
>     2. PyTorch always flattens the input array.
> 
> 
> We'd need to ensure that PyTorch is (1) okay with adding an `axis` kwarg to support taking along a specified axis and (2) making an `axis` kwarg mandatory, thus breaking current behavior (i.e., based on the current proposal, no way to index into a flattened array using `take` which is the current PyTorch behavior).

I don't think there's an issue with progressing this PR:
- `torch.take` with two arguments is actually compatible with `numpy.take` with two arguments,
- adding an `axis` keyword is not controversial for PyTorch, and would be fully backwards compatible if it were to default to `None`,
- the introduction path would add `axis=None` (or better, `dim=None`) first, and then have the appropriate warnings raised before switching the default to `0`,
- the bc-breaking change would be to add `axis=0`; if as discussed above the `axis` keyword is mandatory for >=2-D array input, then there's still no risk of silent breakage, just some code churn. _And this would be the same for PyTorch as for NumPy_,
- I'll note that this will be low-prio for PyTorch to work on though, because there's already `take_along_dim` that does what we propose here (again similar to `np.take_along_axis`), so it's just some churn but doesn't add functionality.

spec/API_specification/signatures/indexing_functions.py
```diff
@@ -0,0 +1,28 @@
+from ._types import Union, array
+
+def take(x: array, indices: Union[int, array], /, *, axis: int) -> array:
+    """
+    Returns elements of an array along an axis.
+
+    .. note::
```
rgommers, line 7: This requires a blank line here to render correctly I think (can't check, preview expired). And some for `.. note::` comment further down.

spec/API_specification/signatures/indexing_functions.py
```diff
@@ -0,0 +1,28 @@
...
+       Conceptually, the call ``take(x, indices, axis=3)`` is equivalent to ``x[:,:,:,indices,...]``; however, explicit indexing via arrays of indices is not currently supported in this specification due to concerns regarding ``__setitem__`` and array mutation semantics.
+
+    Parameters
+    ----------
+    x: array
+        input array.
+    indices: Union[int, array]
```
rgommers, line 14: Allowing a scalar is unusual. It's probably worth doing here because this function is a direct mapping from indexing syntax, where it is allowed. I wanted to point it out though, since we don't allow array-like's almost anywhere else.

rgommers (changes requested):
Thanks @kgryte, overall looks good - a few comments.

One formatting comment: please wrap lines at 80 characters, that's standard in Python and far easier to read/review.

spec/API_specification/signatures/indexing_functions.py
```diff
@@ -0,0 +1,28 @@
...
+        array indices. If ``indices`` is an integer (or a zero-dimensional array satisfying ``operator.index``), the function must follow multi-axis indexing semantics when providing an integer for a single-axis index (see :ref:`indexing`). If ``indices`` is a non-zero-dimensional array, the array must be one-dimensional and have an integer data type.
```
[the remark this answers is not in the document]
kgryte: The reference to multi-axis indexing stems from the conceptual parity, as stated in the description, of `take` with `x[:,:,:,indices,...]` and behavior is already specified as described in the [indexing specification](https://github.com/data-apis/array-api/blob/3c7395c8b681762454f154b7556d75a0fb60676e/spec/API_specification/indexing.rst#multi-axis-indexing) (e.g., dimension reduction, positive/negative integer behavior, etc).

rgommers:
> Seems OK, since it is compatible with `None` as a default. But, is there much reason for it as in that case `arr[index]` is practically identical (unless you need to ensure a copy and `index` is a scalar)?

Except integer array indexing is not in the standard (yet at least), which is why there was so much interest in adding `take`.

> `take_along_axis` is a very different functionality,

Good point, I overlooked that.

spec/API_specification/signatures/indexing_functions.py
kgryte: Seems to render correctly without extra linebreaks. We don't include additional linebreaks elsewhere in the spec atm (e.g., see https://github.com/data-apis/array-api/blob/733f27c91b963216cf14159435c04d4a3dcce294/spec/API_specification/array_api/array_object.py#L167 and https://data-apis.org/array-api/latest/API_specification/generated/signatures.array_object.array.__add__.html#signatures.array_object.array.__add__). Some time ago, we had a mix of notes with and without linebreaks. These were standardized to without linebreaks.

spec/API_specification/signatures/indexing_functions.py
kgryte: I'd also be okay with limiting `indices` to `array` input. This would simplify `take` behavior as spec'd. As currently spec'd, `take` has distinct behavior depending on whether `indices` is an `array` or an `integer`. Accordingly, given

```python
xp.take(x, 1)
```

you need to remember that this will result in a dimension reduction; whereas,

```python
xp.take(x, xp.array([1]))
```

does not result in a dimension reduction.

Furthermore, limiting to only array `indices` would also mean that there is only one way to do things. If a user wants to access a single axis, they should use explicit indexing

```python
x[:, :, 1, :]
```

spec/API_specification/signatures/indexing_functions.py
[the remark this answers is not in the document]
kgryte: Rephrased.

kgryte:
As discussed in the Consortium meeting on 22 September 2022, consensus was to limit the `indices` argument to one-dimensional arrays.

For now, the API will not support multi-dimensional arrays for the `indices` kwarg. In the interim, array API consumers should be able to use `reshape` to achieve something similar. If, based on feedback and usage, multi-dimensional `indices` support is desired, we can revisit the specification and update accordingly.

This PR has been updated to reflect current consensus.

spec/API_specification/signatures/indexing_functions.py
kgryte: The proposed spec has been updated accordingly.

kgryte:
As discussed in the 3 November 2022 meeting, no objections were raised. This PR should be ready for merge.

@rgommers would you mind taking a last look and then approving?

rgommers (approved):

rgommers:
I went through the whole discussion here again, and it looks like we're all happy and there aren't any remaining blockers. This wasn't an easy one! In it goes, thanks all!

Issue #415, opened 2022-04-07
Broadcast behavior in `linalg.cross`


kgryte:
The purpose of this issue is to determine how, or if, broadcasting behavior should be specified for `linalg.cross`.

## Overview

At the moment, `linalg.cross` is specified to only compute the cross product between two arrays `x` and `y` having the **same** shape. Meaning, the current `linalg.cross` behavior is specified to **not** support broadcasting.

The origin of the specified behavior comes from [PyTorch](https://pytorch.org/docs/1.10/generated/torch.cross.html?highlight=cross#torch.cross) (<= v1.10.x) and [TensorFlow](https://www.tensorflow.org/api_docs/python/tf/linalg/cross) which required (at the time of specification) input arrays of the same shape. Not supporting broadcasting was thus the lowest common denominator across array libraries.

TensorFlow has not changed its behavior; PyTorch has.

NumPy, in its `np.cross` API, supports a limited form of broadcasting, as well as `2` element vectors. The broadcasting is limited as the computation dimension must have `2` or `3` elements and is not allowed to broadcast. E.g.,

```python
In [1]: x = np.random.randn(4,3)                                                                                       

In [2]: y = np.random.randn(1,3)                                                                                      

In [3]: z = np.random.randn(4,1)                                                                                       

In [3]: np.cross(x,y)                                                                                                  
Out[3]: 
array([[ 0.04541174,  0.07097804, -0.53846464],
       [ 0.01457143, -0.74247471, -0.1795357 ],
       [ 2.4163107 , -0.72177202, -5.98228811],
       [ 1.14862403, -1.86202792,  0.32601926]])

In [4]: np.cross(x,z)
ValueError: incompatible dimensions for cross product
(dimension must be 2 or 3)

In [5]: np.add(x,z)                                                                                                    
Out[5]: 
array([[ 1.66159175,  0.92220278,  0.93708491],
       [ 0.26326781, -0.37688777,  1.26610177],
       [ 1.34535177,  1.13112439,  0.38157179],
       [-1.78861678, -1.34595513, -0.08110636]])
```

Even though `x` and `z` are broadcast compatible, `np.cross` raises an exception.

Starting in PyTorch v1.11.0, `linalg.cross` supports broadcasting. From the [docs](https://pytorch.org/docs/1.11/generated/torch.linalg.cross.html#torch.linalg.cross),

[...]

PR #417, opened 2022-04-18
Add support for broadcasting to `linalg.cross`

kgryte:
This PR

-   addresses https://github.com/data-apis/array-api/issues/415 by adding support for broadcasting to `linalg.cross`. Broadcasting follows NumPy behavior in only broadcasting the non-compute dimensions. With the exception of TensorFlow, this behavior is consistent (or will be after PyTorch addresses its broadcasting behavior) among array libraries.
-   updates `vecdot` to match `linalg.cross` in being more explicit that broadcasting only applies to non-broadcast dimensions.

### DIFF
--- a/spec/API_specification/signatures/linalg.py
+++ b/spec/API_specification/signatures/linalg.py
@@ -26,16 +26,16 @@ def cholesky(x: array, /, *, upper: bool = False) -> array:
 
 def cross(x1: array, x2: array, /, *, axis: int = -1) -> array:
     """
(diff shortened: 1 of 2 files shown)


kgryte:
As this PR has been discussed (see https://github.com/data-apis/array-api/issues/415#issuecomment-1101130081) and met with approval in consortium meetings and now that PyTorch has now added support for broadcasting, will merge. We can submit follow-up PRs if the specification needs further changes/refinement.

Issue #373, opened 2022-01-20
RFC: Adding complex number support to the specification


kgryte:
# Complex Number Support

> Plan for complex number support in the array API specification.

What follows is a plan for adding complex number support to the [2022 array API specification](https://github.com/data-apis/array-api/issues/343). This RFC is comprised of the following sections:

-   Prior discussions
-   High-level summary concerns and questions
-   Individual API changes

## Prior Discussions

-   Issue: <https://github.com/data-apis/array-api/issues/102>
-   Issue: <https://github.com/data-apis/array-api/issues/153>

### New APIs

```
real
conj (complex conjugate)
imag
arg/angle/phase (phase angle)
linalg.eig
linalg.eigvals
```

### Decisions

-   initial data types: `complex64` and `complex128`?

    -   **Update**: only `complex64` and `complex128` to be added, thus mirroring `float32` and `float64`.

-   rounding complex numbers (component-wise)

    -   NumPy supports round, but not ceil, floor, and trunc.
    -   **Update**: ceiling, flooring, and truncating complex numbers is not well-defined. The only rounding operation which is commonly supported is `round`.
    
-   ordering (lexicographic?)

    -   necessary for sorting, maximum, minimum (e.g., `argsort`, `argmax`, `argmin`, `min`, `max`, `sort`)
    -   issue: <https://github.com/data-apis/array-api/issues/102#issuecomment-748328170>
    -   choices are to define a complex number ordering or leave out of the specification
    -   **Update**: ordering complex numbers is left out of the specification.

-   branch cut policy?

    -   branch cut reference: <https://en.wikipedia.org/wiki/Branch_point#Branch_cuts>
    -   should we strictly follow NumPy/C99, or should we allow for implementation flexibility?
    -   if we strictly follow something other than NumPy/C99, obvious backward compat concern
    -   **Update**: branch cuts will be included in the 2022 revision under a provisional status. Provided no issues arise during adoption and implementation, the provisional status may be lifted in a future revision of the standard.

-   casting from complex to real

    -   discard imaginary component (as in C, NumPy)
    -   other choice would to require user to explicitly invoke `real` before casting
    -   **Update**: decision was to require that array API consumers be explicit in specifying which component they want to cast when casting from complex to real.

-   complex numbers with components which are infinity and/or NaN

    -   in Python, a complex number can be both infinite and NaN (according to `cmath.isinf`, `cmath.isnan`)

        ```
        In [1]: z = complex(float('inf'), float('nan'))

[...]

PR #418, opened 2022-04-18
Add complex floating-point data types

kgryte:
This PR

-   add `complex64` and `complex128` data types to the specification, as outlined in https://github.com/data-apis/array-api/issues/373.
-   defines `complex64` as a single-precision (64-bit) complex floating-point number whose real and imaginary components must be IEEE 754 single-precision (32-bit) binary floating-point numbers.
-   defines `complex128` as a double-precision (128-bit) complex floating-point number whose real and imaginary components must be IEEE 754 double-precision (64-bit) binary floating-point numbers.
-   requires that the default complex floating-point data type match the default floating-point data type (i.e., if `float32`, the default complex floating-point data type must be `complex64`).
-   adds `complex64` and `complex128` to the list of "numeric" data types.
-   adds a new data type category: "complex floating-point data types" for use throughout the specification.
-   adds a new data type category: "real-valued floating-point data types". This category is equivalent to the previous "floating-point data types". The "floating-point data types" category is updated to include both real-valued and complex floating-point data types.
-   adds a new data type category: "real-valued data types". This category is equivalent to the previous "numeric" data type category.
-   updates all current use of "numeric data type" to "real-valued data type" with the aim to add support for complex number data types to APIs on a case-by-case basis via subsequent PRs.
-   updates all current use of "floating-point data type" to "real-valued floating-point data type" with the aim to add support for complex number data types to APIs on a case-by-case basis via subsequent PRs.
-   updates to type promotion tables to follow in a subsequent PR.

### DIFF
--- a/spec/API_specification/array_object.rst
+++ b/spec/API_specification/array_object.rst
@@ -82,7 +82,7 @@ A conforming implementation of the array API standard must provide and support a
     -   `operator.pow(x1, x2) <https://docs.python.org/3/library/operator.html#operator.pow>`_
     -   `operator.__pow__(x1, x2) <https://docs.python.org/3/library/operator.html#operator.__pow__>`_
 
-Arithmetic operators should be defined for arrays having numeric data types.
+Arithmetic operators should be defined for arrays having real-valued data types.
 
 Array Operators
 ~~~~~~~~~~~~~~~
@@ -94,7 +94,7 @@ A conforming implementation of the array API standard must provide and support a
     -   `operator.matmul(x1, x2) <https://docs.python.org/3/library/operator.html#operator.matmul>`_
     -   `operator.__matmul__(x1, x2) <https://docs.python.org/3/library/operator.html#operator.__matmul__>`_
 
-The matmul ``@`` operator should be defined for arrays having numeric data types.
+The matmul ``@`` operator should be defined for arrays having real-valued data types.
 
 Bitwise Operators
 ~~~~~~~~~~~~~~~~~
--- a/spec/API_specification/data_types.rst
+++ b/spec/API_specification/data_types.rst
@@ -117,14 +117,14 @@ Default Data Types
 
 A conforming implementation of the array API standard must define the following default data types.
 
--   a default floating-point data type (either ``float32`` or ``float64``).
+-   a default real-valued floating-point data type (either ``float32`` or ``float64``).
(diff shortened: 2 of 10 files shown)


oleksandr-pavlyk:
Why aren't complex types considered numeric data types? Is this is because they lack inherent total order, than the "numeric data types" is a misnomer and should be renamed to "data types with canonical ordering" or some such

kgryte:
@oleksandr-pavlyk Updated the PR. Had originally shied away from changing the dtypes belonging to "numeric" as this implied a change to the semantic meaning for purposes of the spec. However, after discussion, this was not seen as a problem. Added a "real-valued data types" category and will update the spec accordingly.

leofang:
> Added a "real-valued data types" category and have updated the spec accordingly.

This is still confusing... A lot of dunder methods like `__abs__`, `__mul__`, ..., are well defined even for complex-valued arrays. I think we should only exclude a few like those involving ordering (`__ge__`, `__le__`, ...), modulo (`__mod__`), etc.

kgryte:
@leofang For many of the APIs, "real-valued data types" is a placeholder and each API will be updated on an individual basis in subsequent PRs. The purpose of this PR is to solely update the list of data types. Because of the change to the data type categories (e.g., `numeric`), I updated the APIs to use the "real-valued data type" category to preserve the status quo.

leofang:
Ah ok, I missed the updated PR description:

> * updates all current use of "numeric data type" to "real-valued data type" with the aim to add support for complex number data types to APIs on a case-by-case basis via subsequent PRs.

Thanks, Athan!

leofang (approved):
LGTM!

kgryte:
Thanks, @leofang!

I'll leave it open for another day or so before merging. And then we'll get the ball moving on the rest of the complex number support updates. 😅

leofang:
Question: Is the complex-number support counted as extension or required? If it's extension, then does it mean all APIs supporting complex numbers become extension as well?

kgryte:
@leofang Complex number support would be **required**. Because of how deeply intertwined dtype support is with all APIs in the spec, allowing complex number to be an extension does not seem to make sense (e.g., how would I ascertain whether a given array API supports complex numbers in a portable fashion?).

kgryte:
As this PR has been approved and up for review for 4 weeks, will merge. Any further revisions may be addressed in follow-up PRs...

PR #421, opened 2022-04-21
Fix erroneous updates to arange intended for linspace

kgryte:
This PR

-   fixes erroneous updates to `arange` intended for `linspace` in https://github.com/data-apis/array-api/pull/393. Namely, `linspace` should be limited to floating-point data types, not `arange`. Accordingly, this PR reverts the changes to `arange` and adds the intended changes to `linspace`.


spec/API_specification/signatures/creation_functions.py
```diff
@@ -237,6 +231,12 @@ def linspace(start: Union[int, float], stop: Union[int, float], /, num: int, *,
     -------
     out: array
         a one-dimensional array containing evenly spaced values.
+
```
honno, line 234: Locally this doesn't render correctly:

Screenshot from 2022-04-22 11-43-29

But another newline seems to does the trick:

```suggestion


```

honno (changes requested):
Looks good except for one comment. Sorry I didn't catch this last time :sweat_smile:

kgryte:
Thanks, @honno! Will merge...

PR #429, opened 2022-05-10
Clarify the required semantics for broadcasting with `__setitem__`

### LINKED ISSUE #424: Be more clear about broadcasting behavior in setitem
NumPy allows this behavior in setitem

```py
>>> import numpy as np
>>> x = np.empty((2, 3, 4))
>>> a = np.empty((1, 3, 4))
>>> x[1, ...].shape
(3, 4)
>>> x[1, ...] = a
```

This is sort of a "reverse broadcasting". The rhs can't be broadcast to the shape of the lhs, but it can be made into that shape by adding a size 1 dimension.

We should more explicitly disallow this in the spec. Right now the spec [just says](https://data-apis.org/array-api/latest/API_specification/broadcasting.html#in-place-semantics):

> As implied by the broadcasting algorithm, in-place element-wise operations must not change the shape of the in-place array as a result of broadcasting.

I think it should also say something like "the shape of the right-hand side of an operation must be broadcastable into the shape of the left-hand side, after applying any indices". It's also not completely clear that this statement applies to plain `__setitem__` as an "in-place operation". 

Note that the NumPy developers have expressed interest in deprecating this behavior in NumPy https://github.com/numpy/numpy/pull/10615#issuecomment-920987394 (@seberg). It's also worth noting that NumPy's [documentation](https://numpy.org/doc/stable/user/basics.indexing.html) says

> You may use slicing to set values in the array, but (unlike lists) you can never grow the array. The size of the value to be set in x[obj] = value must be (broadcastable) to the same shape as x[obj]."

which itself doesn't quite capture this "reverse broadcasting" semantics.

Currently the numpy.array_api implementation allows the above to work. I haven't checked any other array libraries.

asmeurer:
Fixes #424.

I have used SHOULD here as suggested in the consortium meeting to not disallow NumPy-like "reverse broadcasting" semantics.

### DIFF
--- a/spec/API_specification/broadcasting.rst
+++ b/spec/API_specification/broadcasting.rst
@@ -113,3 +113,16 @@ In-place Semantics
 ------------------
 
 As implied by the broadcasting algorithm, in-place element-wise operations (including ``__setitem__``) must not change the shape of the in-place array as a result of broadcasting. Such operations should only be supported in the case where the right-hand operand can broadcast to the shape of the left-hand operand, after any indexing operations are performed.
+
+For example:
+
+::
+
+   x = empty((2, 3, 4))
+   a = empty((1, 3, 4))
+
+   # This is OK. The shape of a, (1, 3, 4), can broadcast to the shape of x[...], (2, 3, 4)
+   x[...] = a
+
+   # This is not allowed. The shape of a, (1, 3, 4), can NOT broadcast to the shape of x[1, ...], (3, 4)
+   x[1, ...] = a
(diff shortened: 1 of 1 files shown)


leofang:
Can we add two simple examples to illustrate what's allowed and what's not?

kgryte:
@asmeurer Would you be willing to update this PR with two examples, one illustrating what is allowed and another illustrating what is disallowed?

asmeurer:
Added.

rgommers (approved):
LGTM, the examples are clear. Thanks @asmeurer, all.

PR #434, opened 2022-05-23
Add complex number support to `asarray`

kgryte:
This PR

-   adds complex number support to `asarray`. Importantly, if one or more values are `complex` numbers, the output data type must be the default complex floating-point data type.

### DIFF
--- a/spec/API_specification/array_api/creation_functions.py
+++ b/spec/API_specification/array_api/creation_functions.py
@@ -46,7 +46,7 @@ def asarray(obj: Union[array, bool, int, float, complex, NestedSequence, Support
         output array data type. If ``dtype`` is ``None``, the output array data type must be inferred from the data type(s) in ``obj``. If all input values are Python scalars, then, in order of precedence,
 
         -   if all values are of type ``bool``, the output data type must be ``bool``.
-        -   if the values are a mixture of ``bool``\s and ``int``, the output data type must be the default integer data type.
+        -   if all values are of type ``int`` or are a mixture of ``bool`` and ``int``, the output data type must be the default integer data type.
         -   if one or more values are ``complex`` numbers, the output data type must be the default complex floating-point data type.
         -   if one or more values are ``float``\s, the output data type must be the default real-valued floating-point data type.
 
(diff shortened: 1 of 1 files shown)


spec/API_specification/array_api/creation_functions.py
```diff
@@ -43,10 +43,11 @@ def asarray(obj: Union[array, bool, int, float, NestedSequence, SupportsBufferPr
            An object supporting the buffer protocol can be turned into a memoryview through ``memoryview(obj)``.
 
     dtype: Optional[dtype]
-        output array data type. If ``dtype`` is ``None``, the output array data type must be inferred from the data type(s) in ``obj``. If all input values are Python scalars, then
+        output array data type. If ``dtype`` is ``None``, the output array data type must be inferred from the data type(s) in ``obj``. If all input values are Python scalars, then, in order of precedence,
```
rgommers, line 46: The precedence doesn't seem needed, there's only ever one condition that can be true.

spec/API_specification/array_api/creation_functions.py
```diff
@@ -43,10 +43,11 @@ def asarray(obj: Union[array, bool, int, float, NestedSequence, SupportsBufferPr
...
 
         -   if all values are of type ``bool``, the output data type must be ``bool``.
         -   if the values are a mixture of ``bool``\s and ``int``, the output data type must be the default integer data type.
```
rgommers, line 49: This one is a little odd, you can read it as explicitly disallowing `int`-only values.

rgommers (commented):
LGTM modulo a few small textual things.

spec/API_specification/array_api/creation_functions.py
asmeurer: I think that was the case before but now the last two bullet points can both be true.

spec/API_specification/array_api/creation_functions.py
kgryte: Will rephrase.

kgryte: This has now been addressed.

leofang:
I fail to see why complex should appear before (or after) float with the "precedence" clause, I must be missing something...

kgryte:
@leofang Because if at least one value is `complex`, then the returned array must have a complex data type. Hence, if provided a mixture of `complex` and `float`, `complex` takes precedence. This follows type promotion semantics.

leofang (approved):

kgryte:
Thanks, @leofang! As this now has two approvals, will merge and any further changes can be addressed through subsequent PRs...

PR #436, opened 2022-05-23
Clarify expected values in `eye` when `dtype` is a complex number data type

kgryte:
This PR

-   adds a note to `eye` concerning expected output values when `dtype` is either `complex64` or `complex128`. Namely, only real components along the kth diagonal are ones. Imaginary components are always zero for all elements.


rgommers (approved):
LGTM

kgryte:
As this PR has received approval and the changes straightforward, will go ahead and merge...

PR #438, opened 2022-05-23
Clarify expected values in `ones` and `ones_like` when `dtype` is a complex number data type

kgryte:
This PR

-   adds a note to both `ones` and `ones_like` concerning expected output values when `dtype` is either `complex64` or `complex128`. Namely, only real components are ones. Imaginary components are zero for all elements.


kgryte:
As this PR has received approval and the changes are straightforward, will go ahead and merge...

PR #439, opened 2022-05-23
Update `iinfo` description to be consistent with `finfo`

kgryte:
This PR

-   updates the return value description for `iinfo` to be consistent with `finfo`. The current description seems to have meant to say "encapsulates". The changes introduced in this PR use a simpler sentence construction which mirrors that of `finfo`.


honno (approved):
LGTM

kgryte:
Thanks, @honno! As this is typo fix, will go ahead and merge.

PR #441, opened 2022-05-23
Clarify expected behavior when providing a complex number array to `nonzero`

kgryte:
This PR

-   clarifies expected behavior when providing a complex number array to `nonzero`. Namely, nonzero elements are those elements have either a real or imaginary component (or both) which is non-zero.
-   adds a note clarifying that, for boolean arrays, `True` is non-zero.


kgryte:
Thanks, @leofang! As this PR now has two approvals and the changes are straightforward, will merge...

PR #442, opened 2022-05-23
Clarify expected behavior when providing a complex number array to `all` and `any`

kgryte:
This PR

-   clarifies expected behavior when providing a complex number array to `all` and `any`. Namely, complex number values having a non-zero component (real or imaginary) must evaluate to ``True``.

### DIFF
--- a/spec/API_specification/array_api/utility_functions.py
+++ b/spec/API_specification/array_api/utility_functions.py
@@ -8,6 +8,9 @@ def all(x: array, /, *, axis: Optional[Union[int, Tuple[int, ...]]] = None, keep
        Positive infinity, negative infinity, and NaN must evaluate to ``True``.
 
     .. note::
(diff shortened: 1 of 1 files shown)


rgommers (approved):
LGTM

asmeurer:
Do any and all already support numeric data types? I would have expected them to only support boolean data types.

kgryte:
@asmeurer Yes, they already support all supported data types; see https://github.com/data-apis/array-api/blob/34aa9251bec8e53d8e7f4330f0b2b6221b3f6dcb/spec/API_specification/array_api/utility_functions.py#L15-L16

This PR adds clarification, similar to floating-point infinities and NaN, for how complex numbers should evaluate to `True`.

kgryte:
As this PR has received approval and the changes are straightforward, will go ahead and merge. Any further changes can be addressed in subsequent PRs...

PR #446, opened 2022-05-30
Add `conj` specification for computing the complex conjugate

kgryte:
This PR

-   adds support for computing the complex conjugate via `conj`, which is widely [implemented](https://github.com/data-apis/array-api-comparison/blob/7ecfa9867ec43a28e914b35868c041443d31d772/signatures/mathematical-functions-elementwise/conj.md) by array libraries.
-   follows the conventions set forth for element-wise mathematical functions.
-   restricts input array types to complex floating-point arrays, following the precedent in https://github.com/data-apis/array-api/pull/427.

## Notes

-   Re: restricting to complex dtypes. We should note that this, similar to `real`, runs contrary to array libraries, more generally, which allow real-valued arrays as input and define `conj` as a no-op for those inputs (see [PyTorch v1.11](https://pytorch.org/docs/1.11/generated/torch.conj.html), [TensorFlow](https://www.tensorflow.org/api_docs/python/tf/math/conj), and NumPy). However, this PR chooses to be consistent with `real` in requiring users to provide a complex number array for the primary reason that, in most cases, when a user is calling `conj(z)`, the expectation is that `z` is complex-valued, and, if real-valued, this is likely a bug.

### DIFF
--- a/spec/API_specification/array_api/elementwise_functions.py
+++ b/spec/API_specification/array_api/elementwise_functions.py
@@ -401,6 +401,34 @@ def ceil(x: array, /) -> array:
         an array containing the rounded result for each element in ``x``. The returned array must have the same data type as ``x``.
     """
 
+def conj(x: array, /) -> array:
+    """
+    Returns the complex conjugate for each element ``x_i`` of the input array ``x``.
+
+    For complex numbers of the form
+
+    .. math::
+       a + bj
+
+    the complex conjugate is defined as
+
+    .. math::
+       a - bj
+
+    Hence, the returned complex conjugate must be computed by negating the imaginary component of each element ``x_i``.
+
+    Parameters
+    ----------
+    x: array
+        input array. Should have a complex-floating point data type.
+
+    Returns
+    -------
+    out: array
+        an array containing the element-wise results. The returned array must have the same data type as ``x``.
+    """
+
+
 def cos(x: array, /) -> array:
     """
     Calculates an implementation-dependent approximation to the cosine, having domain ``(-infinity, +infinity)`` and codomain ``[-1, +1]``, for each element ``x_i`` of the input array ``x``. Each element ``x_i`` is assumed to be expressed in radians.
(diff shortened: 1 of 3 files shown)


spec/API_specification/array_api/elementwise_functions.py
```diff
@@ -401,6 +401,34 @@ def ceil(x: array, /) -> array:
...
+
+    .. math::
+       a - bj
+
+    Hence, the returned complex conjugates must be computed by negating the imaginary component of each element ``x_i``.
+
+    Parameters
+    ----------
+    x: array
+        input array. Should have a complex-floating point data type.
```
leofang, line 423: @kgryte Could you remind me what's our recommendation for the exception? `ValueError` or `TypeError` (or neither)?

kgryte: Typically, we don't specify the exact form an exception should take and defer to implementing libraries to make that decision.

kgryte: Additionally, the spec here uses SHOULD to allow implementations to support other data types due to, e.g., backward compat concerns. So requiring an exception is not desirable.

kgryte: IMO, I think the restriction is something we still need to decide (both here and for `real` and `imag`; was on the agenda for the most recent consortium meeting, but we ran out of time). Personally, I'm sympathetic to both allowing all numeric dtypes and to limiting to only complex dtypes.

leofang: +1 no strong opinion here.

jakirkham (approved):

spec/API_specification/array_api/elementwise_functions.py
shoyer: I would prefer to allow both complex and real inputs for `real`, `imag` and `conj`. This is useful for writing generic routines that handle real or complex numbers.

leofang (approved):

leofang:
Would be nice if we can discuss #425 again this week. It seems this and #427 are blocked.

seberg:
My feeling would be that we should probably allow it, since Python itself also always implemntes `.real` and `.conjugate()` (althought not `.imag` I believe).
But of course for now that just means not explicitly saying that it must be forbidden (vs. should not be allowed or does not have to be allowed).

spec/API_specification/array_api/elementwise_functions.py
kgryte: In the 9/22/2022 meeting, the plan is to go ahead and merge the PR as is (i.e., restricted to complex dtypes). We can revisit if implementors receive user feedback that the API should be extended to real-valued dtypes.

kgryte:
As discussed in the Sept 22, 2022, consortium meeting, the plan is to merge this PR as is. We can revisit the decision to only support complex dtypes based on downstream library implementations and feedback should this prove problematic/undesirable.

The spec, as written, uses **should**, so implementors can extend to real-valued dtypes for reasons of backward compatibility. However, users should only except that complex dtypes will work across all spec compliant array API implementations.

PR #447, opened 2022-05-30
Add complex number support to `positive`

kgryte:
This PR

-   adds complex number support to `positive` (and `__pos__`). No special considerations need to be made for complex numbers, as positive is effectively an identity function.


leofang (approved):

kgryte:
Thanks, @leofang! Will merge, as the changes are straightforward and align with existing array implementations.

PR #454, opened 2022-06-13
Add complex number support to `cos`

kgryte:
This PR

-   adds complex number support to `cos` by documenting special cases. The cosine is an entire function in the complex plane. Thus, the function has no branch cuts.
-   derives special cases from [C99](https://en.cppreference.com/w/c/numeric/complex/ccos).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ node_modules/
 __pycache__/
 *.pyc
 spec/**/generated
+tmp/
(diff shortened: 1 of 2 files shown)


kgryte:
As `cos` does not involve branch cuts and the special cases introduced in this PR rely on `cosh` and follow prior art, will merge. Further revisions can be addressed in subsequent PRs...

PR #458, opened 2022-06-21
Add complex number support to `tanh`

kgryte:
This PR

-   adds complex number support to `tanh` by documenting special cases. The hyperbolic tangent is an analytical function in the complex plane and has no branch cuts.
-   updates the input and output array data types to be any floating-point data type, not just real-valued floating-point data types.
-   derives special cases from [C99](https://en.cppreference.com/w/c/numeric/complex/ctanh) and tested against NumPy (script found below).
-   adds a warning concerning two special cases where results may differ depending on which C version an array library compiles against (for those libraries which are C compiled).

<details>

```python
import numpy as np
import math

def is_equal_float(x, y):
    """Test whether two floating-point numbers are equal with special consideration for zeros and NaNs.

    Parameters
    ----------
    x : float
        First input number.
    y : float
        Second input number.

    Returns
    -------
    bool
        Boolean indicating whether two floating-point numbers are equal.

    Examples
    --------
    >>> is_equal_float(0.0, -0.0)
    False
    >>> is_equal_float(-0.0, -0.0)
    True
    """
    # Handle +-0:
    if x == 0.0 and y == 0.0:
        return math.copysign(1.0, x) == math.copysign(1.0, y)

    # Handle NaNs:
    if x != x:
        return y != y

    # Everything else, including infinities:
    return x == y


def is_equal(x, y):
    """Test whether two complex numbers are equal with special consideration for zeros and NaNs.

    Parameters
    ----------
    x : complex
        First input number.
    y : complex
        Second input number.

    Returns
    -------
    bool
        Boolean indicating whether two complex numbers are equal.

    Examples
    --------
    >>> import numpy as np
    >>> is_equal(complex(np.nan, np.nan), complex(np.nan, np.nan))
    True
    """
    return is_equal_float(x.real, y.real) and is_equal_float(x.imag, y.imag)


# Strided array consisting of input values and expected values:
values = [
    complex(0.0, 0.0),        # 0
    complex(0.0, 0.0),        # 0
    
    complex(-0.0, -0.0),      # 1
    complex(-0.0, -0.0),      # 1
    
    complex(1.0, np.inf),     # 2
    complex(np.nan, np.nan),  # 2
    
    complex(0.0, np.inf),     # 3
    complex(0.0, np.nan),     # 3, seems to be a bug in NumPy, which returns `NaN + NaN j`; however, this does match old C99 behavior (see https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1892.htm#dr_471)
    
[...]

PR #459, opened 2022-06-21
Add complex number support to `tan`

kgryte:
This PR

-   adds complex number support to `tan` by documenting special cases. The tangent is an analytical function in the complex plane. Thus, the function has no branch cuts.
-   derives special cases from [C99](https://en.cppreference.com/w/c/numeric/complex/ctan).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ node_modules/
 __pycache__/
 *.pyc
 spec/**/generated
+tmp/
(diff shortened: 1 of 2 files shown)


kgryte:
As https://github.com/data-apis/array-api/pull/458 has been merged, this PR can also be merged. Further modifications can be addressed in subsequent PRs...

PR #464, opened 2022-07-18
Add `self` param to array object properties

IsaacBreen:


leofang:
I am not typing expert but I think we need `TypeVar(T, bound='array')` now (PEP 484), and in the future `Self` (PEP 673). The current approach would not work for subclasses. My vote would go for the status quo (no type hint for self) but let's wait for others to chime in.

asmeurer:
`array` is what the other methods use so keeping that here is fine. We can open another issue about using a different type hint for `self`.

rgommers:
CI failure is unrelated, let's retry that.

leofang:
Sorry, so this PR does two things:
1. It fixes the missing `self` to the array methods (which I overlooked)
2. It adds a type hint for `self`

>  We can open another issue about using a different type hint for self.

Let's do that.

rgommers:
Okay, CI is unrelated, I'll open a separate issue for that too.

rgommers:
The type hint for `self` is one instance of a general pattern; same as for scalars we use `int` and not anything that ducktypes with `int` (which can be type-able by `SupportsInt` or some such thing).

rgommers (approved):
So it looks like we're happy here, in it goes.

Issue #471, opened 2022-08-04
Should vecdot broadcast along the contracted axis?


asmeurer:
The spec implies that it should https://data-apis.org/array-api/latest/API_specification/generated/signatures.linear_algebra_functions.vecdot.html#signatures.linear_algebra_functions.vecdot. However, other functions like matmul do not broadcast along the contracted axis. 

This came up in the numpy implementation, which is currently wrong, but is being fixed at https://github.com/numpy/numpy/pull/21928.

leofang:
tensordot does not allow broadcasting in the contracted axes, so I think vecdot shouldn't, either.

PR #473, opened 2022-08-29
Clarify broadcasting behavior in `vecdot`

kgryte:
This PR resolves https://github.com/data-apis/array-api/issues/471.

The existing spec provides conflicting guidance saying both that the axes over which to compute the dot product must be equal, while also saying that input arrays must be broadcast compatible without qualification, thus implying that the contracted axis could also broadcast.

This PR explicitly defines broadcast behavior for only the contracted axes, thus bringing `vecdot` inline with broadcasting behavior in `tensordot`.


rgommers (approved):
This LGTM and aligns with what @leofang and @asmeurer said in the issue. So in it goes, thanks @kgryte.

PR #480, opened 2022-09-21
Include `__array_api_version__` in API specification section

honno:
`xp.__array_api_version__` has been specified in our [Versioning](https://data-apis.org/array-api/latest/future_API_evolution.html#versioning) section for a while now (~2yrs), but no such mention of it is found in the API specification section, which is probably why this wasn't implemented in numpy/numpy/pull/18585 and all subsequent adoption efforts across the ecosystem (happy to submit respective PRs after we get this in).

So this PR simply creates a `spec/API_specification/version.rst` file which includes a documented `__array_api_version__` attribute, which also will show up in our [`array-api/latest/API_specification/`](https://data-apis.org/array-api/latest/API_specification/index.html) page (as well as the sidebar).

### DIFF
--- a/spec/API_specification/array_api/__init__.py
+++ b/spec/API_specification/array_api/__init__.py
@@ -12,3 +12,9 @@ from .sorting_functions import *
 from .statistical_functions import *
 from .utility_functions import *
 from . import linalg
+
+
+__array_api_version__: str = "2021.12"
+"""
+String representing the version of the array API specification which the conforming implementation adheres to.
+"""
--- a/spec/API_specification/array_api/__init__.py
+++ b/spec/API_specification/array_api/__init__.py
@@ -14,7 +14,7 @@ from .utility_functions import *
 from . import linalg
 
 
-__array_api_version__: str = "2021.12"
+__array_api_version__: str = "YYYY.MM"
 """
 String representing the version of the array API specification which the conforming implementation adheres to.
 """
--- a/spec/API_specification/index.rst
+++ b/spec/API_specification/index.rst
@@ -24,3 +24,4 @@ API specification
    statistical_functions
    type_promotion
    utility_functions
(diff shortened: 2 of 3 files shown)


Zac-HD:
After this is merged, let's also add it to the conformance test suite.

spec/API_specification/array_api/__init__.py
```diff
@@ -12,3 +12,9 @@
 from .statistical_functions import *
 from .utility_functions import *
 from . import linalg
+
+
+__array_api_version__: str = "2021.12"
```
leofang, line 17: Q: Would this appear in all versions (2021.12 and draft)?

honno: Oh yeah good shout. I would like to keep this variable in the stubs to keep in line with how we document all over non-function objects - we can bump it when we have a release, and just hide the value in the RST docs.

honno: I've assigned this var to `"YYYY.MM"`, which I think works well. And it means we don't have to update this variable, which we'll inevitably forget to do (without tooling at least).

---

Ideally I thought having no value would help, i.e.

```python
__array_api_version__: str
```

but this means we don't actually have a Python object `__array_api_version__` to import.

We could also play around with Sphinx `:autosummary:` to just hide the render value, but couldn't find an obvious and simple solution.

leofang: Thanks, @honno, that's fine, I can live with it 🙂 I must have done some kind of Sphinx hacks to do what you described, but this is the kind of hacks that you forget once it's working... We can revisit if needed.

oleksandr-pavlyk (approved):

leofang:
Merging as this got two green checks and is what we should have had.

PR #484, opened 2022-09-23
Complex dtype support in `finfo()`

honno:
Per discussion in #433 and yesterdays meeting, this PR makes `finfo()` support complex dtypes by returning information of its real-valued component dtype, e.g.

```python
>>> xp.finfo(xp.complex64)
finfo_object(
    bits=32,
    eps=1.1920928955078125e-07,
    max=3.4028234663852886e+38,
    min=-3.4028234663852886e+38,
    smallest_normal=1.1754943508222875e-38
)
>>> xp.finfo(xp.float32)
... # should contain the same information as above
```

(see #485 RE `.dtype`)

PR #485, opened 2022-09-23
`iinfo()` and `finfo()` result holds `dtype` attribute

honno:
In yesterdays meeting @oleksandr-pavlyk bought up that NumPy's `iinfo()`/`finfo()` returns info objects with a `dtype` attribute. ~There seemed to be no objections, but it turns out `dtype` wasn't actually document for NumPy...~ Resolved in https://github.com/numpy/numpy/pull/22334

~I've left this PR separate from #484 to keep things atomic, incase things become finicky. It depends on that PR because of the subtle distinction I made in `finfo()` to clarify attributes relate to real-valued floats... just want that to resolve (i.e. be merged) first.~ That was merged, so happy with this PR on my end!

### DIFF
--- a/spec/API_specification/array_api/data_type_functions.py
+++ b/spec/API_specification/array_api/data_type_functions.py
@@ -80,6 +80,10 @@ def finfo(type: Union[dtype, array], /) -> finfo_object:
         - **smallest_normal**: *float*
 
           smallest positive real-valued floating-point number with full precision.
+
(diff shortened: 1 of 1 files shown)


rgommers:
@honno after fixing the merge conflict, I think this can be moved out of draft status now that gh-484 has been merged?

leofang (approved):

PR #496, opened 2022-10-17
Add `imag` specification for returning the imaginary component of a complex number

kgryte:
This PR

-   adds a specification for `imag`, an API for returning the imaginary component of a complex number for each element ``x_i`` of an input array ``x``.

-   follows https://github.com/data-apis/array-api/pull/427 and https://github.com/data-apis/array-api/pull/446 in restricting the API to only arrays having complex floating-point dtypes. This is as discussed in the Sept 22, 2022, Consortium workgroup meeting.


leofang (approved):

kgryte:
As this is a straightforward addition and has already been discussed in consortium meetings, will merge. Any further revisions can be addressed in a subsequent PR.

spec/API_specification/array_api/elementwise_functions.py
```diff
@@ -785,6 +785,21 @@ def greater_equal(x1: array, x2: array, /) -> array:
...
+
+    Parameters
+    ----------
+    x: array
+        input array. Should have a complex floating-point data type.
+
+    Returns
+    -------
+    out: array
+        an array containing the element-wise results. The returned array must have a floating-point data type with the same floating-point precision as ``x`` (e.g., if ``x`` is ``complex64``, the returned array must have the floating-point data type ``float32``).
```
asmeurer, line 800: Should this be "must" or "should"?

kgryte: In what scenarios should/would this not be the case? And why?

asmeurer: Oh my bad I was reading this as if it were the input array, not the output. I see the input does say "should".

kgryte: For the input dtype, allowing array libraries to support real-valued dtypes makes sense in order to allow for backward compat; however, all array libraries should return an array having a dtype with the same precision.

I suppose you may be thinking if an array library supports real-valued int dtypes, would requiring a float dtype make sense.

PR #498, opened 2022-10-20
API: Specify that DLPack should use BufferError

seberg:
I would actually like MUST, but since most implementations currently don't actually give a ``BufferError``, writing it as SHOULD instead.

First part of closing https://github.com/numpy/numpy/issues/20742

### DIFF
--- a/spec/API_specification/array_api/array_object.py
+++ b/spec/API_specification/array_api/array_object.py
@@ -295,9 +295,9 @@ class _array():
         ------
         BufferError
             Implementations should raise ``BufferError`` when the data cannot
-            cannot be exported as DLPack (e.g. incompatible dtype or strides).
-            Other errors are raised when export fails for other reasons
-            (e.g. incorrect arguments passed or out of memory).
+            be exported as DLPack (e.g., incompatible dtype or strides). Other
+            errors are raised when export fails for other reasons (e.g., incorrect
+            arguments passed or out of memory).
 
         """
 
--- a/spec/API_specification/array_api/array_object.py
+++ b/spec/API_specification/array_api/array_object.py
@@ -290,6 +290,15 @@ class _array():
         -------
         capsule: PyCapsule
             a DLPack capsule for the array. See :ref:`data-interchange` for details.
+
+        Raises
+        ------
+        BufferError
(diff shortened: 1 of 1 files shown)


oleksandr-pavlyk:
@seberg I assume that this recommendation does not fix the namespace for the actual exception object, right?

oleksandr-pavlyk:
While DLPack is a buffer conceptually, it is not the buffer as in `Py_buffer`. Why not introduce dedicated `DLPackCreationError` in array namespace? 

Here is what I had in my POC:

    Raises:
        DLPackCreationError: when array can be represented as
            DLPack tensor. This may happen when array was allocated
            on a partitioned sycl device, or its USM allocation is
            not bound to the platform default SYCL context.
        MemoryError: when host allocation to needed for `DLManagedTensor`
            did not succeed.
        ValueError: when array elements data type could not be represented
            in `DLManagedTensor`.

kgryte:
Based on 11/3/2022 meeting, we are good moving forward with this PR.

leofang (approved):

spec/API_specification/array_api/array_object.py
```diff
@@ -290,6 +290,15 @@ def __dlpack__(self: array, /, *, stream: Optional[Union[int, Any]] = None) -> P
...
         capsule: PyCapsule
             a DLPack capsule for the array. See :ref:`data-interchange` for details.
+
+        Raises
+        ------
+        BufferError
+            Implementations should raise ``BufferError`` when the data cannot
+            cannot be exported as DLPack (e.g. incompatible dtype or strides).
+            Other errors are raised when export fails for other reasons
+            (e.g. incorrect arguments passed or out of memory).
```
kgryte, line 300:
```suggestion
            Implementations should raise ``BufferError`` when the data cannot
            be exported as DLPack (e.g., incompatible dtype or strides). Other
            errors are raised when export fails for other reasons (e.g., incorrect
            arguments passed or out of memory).
```

kgryte (approved):
@seberg Removed a duplicated word. Otherwise, LGTM.

kgryte:
This has received two approvals and was discussed in workgroup meetings. Will merge.

PR #502, opened 2022-11-03
Add support for specifying the output array dtype in `linalg.trace`

### LINKED ISSUE #493: Use similar promotion rules in trace() as sum() and prod()
`sum()` and `prod()` use special promotion rules (see https://github.com/data-apis/array-api/issues/202 and https://data-apis.org/array-api/latest/API_specification/generated/signatures.statistical_functions.sum.html#signatures.statistical_functions.sum). `trace()` is effectively a sum over the diagonal, so it should use the same rules. `np.trace()` currently already acts the same as `np.sum()` in this regard. 

Presumably this also means we should add a dtype argument to `trace()`

kgryte:
This PR

- resolves #493 by adding support for a `dtype` kwarg and specifying promotion rules for the output array dtype.
- documents special cases for when `N = 0` and when an element is `NaN`, as done for `sum`.

### DIFF
--- a/spec/API_specification/array_api/linalg.py
+++ b/spec/API_specification/array_api/linalg.py
@@ -1,4 +1,4 @@
-from ._types import Literal, Optional, Tuple, Union, Sequence, array
+from ._types import Literal, Optional, Tuple, Union, Sequence, array, dtype
 from .constants import inf
 
 def cholesky(x: array, /, *, upper: bool = False) -> array:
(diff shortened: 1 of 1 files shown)


kgryte:
Based on the 3 November 2022 meeting, no objections were raised to the changes introduced in this PR. Will keep open for another week to allow further comment.

kgryte:
As the changes introduced in this PR follow the same guidance as `sum`, will merge. Any further modifications can be performed in follow-up PRs.

PR #503, opened 2022-11-03
Add specification for `isdtype`

### LINKED ISSUE #425: RFC: add data type inspection utilities to the array API specification

kgryte:
This PR

- resolves #425 by adding an API for testing whether a provided `dtype` is of a specified data type kind.
- ~~specifies that the function name is `is_type` (rather than `is_dtype`) in order to match `result_type`. Another possibility is `istype` in order to match `astype`.~~ **Update**: based on feedback, the function name is specified as `isdtype`. The functions `astype` and `result_type` are now considered ill-named (e.g., `as_dtype`/`asdtype` and `result_dtype`/`common_dtype` are considered better based on present naming conventions, but are named as such due to historical reasons). The choice of `isdtype` is considered preferable over `is_dtype` in order to be consistent with `isfinite`, `isinf`, and `isnan`, which already exist in the specification. One should note that this does not follow strict PEP8 guidance; however, consistency within specification was considered preferable over PEP8 purity.
- follows the API design proposed and voted on in https://github.com/data-apis/array-api/issues/425#issuecomment-1271413533 and supports providing a `dtype`, a `str`, and a `tuple` of `dtype` and `str`.
- supports the following dtype kinds: `'bool'`, `'signed integer'`, `'unsigned integer'`, `'integral'`, `'real floating'`,`'complex floating'`, and `'numeric'`.
- does not include the kind `"floating"` (as described in the specification data type categories). This kind would be achieved via a tuple: `('real floating', 'complex floating')`.


kgryte:
Based on the 3 November 2022 meeting, a few changes should be made to this PR:

1. change `integer` to `integral` (in order to avoid ambiguity with `integer` referring to signed integer dtypes).
2. change `real` to `real floating` (due to ambiguity of what is meant by "real").
3. change `complex` to `complex floating` (in order to mirror `real floating`).
4. add a note explaining that array API implementors are not limited to including only dtypes as described in the spec in particular categories. E.g., `float16` and `bfloat16` can be included in `real floating`.

spec/API_specification/array_api/data_type_functions.py
```diff
@@ -117,6 +117,36 @@ def iinfo(type: Union[dtype, array], /) -> iinfo_object:
           integer data type.
     """
 
+def is_type(dtype: dtype, kind: Union[dtype, str, Tuple[Union[dtype, str], ...]]) -> bool:
```
rgommers, line 120: should be `is_dtype`?

kgryte: Discussed in the OP. In short, if we use `is_dtype`, we have three different naming conventions:

```
astype
result_type
is_dtype
```

Chose `is_type` to be consistent with `result_type`. Discussed in consortium meeting. If consistency is valued, another option is `istype`.

seberg: Yeah, this is annoying.  I am considering bikeshedding it also on NumPy, maybe someone has a good idea.
`type` is weird, and `result_type` and `astype` should be more reasonably `asdtype` and especially `result_dtype` (for `astype`, somehow I find it a bit less weird).

So I much prefer `dtype`, but I am a bit on the hedge of whether aiming for consistency is better here.  There was also discussion about `isdtype/istype` vs. with underscore.  Without underscore does make some sense also w.r.t. `isinstance` or `isclass` maybe @honno's rule of is and as don't need underscore is right.

rgommers: @seberg I share your sentiments, discussing in NumPy sounds like a good idea.

Regarding consistency: I'd argue that `astype` and `result_type` are the outliers; we already have `.dtype` and consistently use "dtype" in language everywhere. "type" has a very different meaning in Python.

seberg: There are many similar things in Python which are exceptions, like `isinstance`, `hasattr`, etc.  I am not sure if Python considers these badly named.  That said, I don't have a strong opinion either way.
I do agree that `dtype` is better and `result_type` is badly named (I really prefer `common_dtype`, numpy has `common_type`, but that needs to go anyway).

rgommers: Re PEP8 and last comments I agree with @seberg. Sebastian, was it discussed in NumPy (sorry, I missed most of the last community calls)?

rgommers:
> Based on the 3 November 2022 meeting, a few changes should be made to this PR:

(1) and (4) are more clear cut than (2) and (3), which make things more verbose. But I'm fine with all four of those.

I do really want to see the name change to either `is_dtype` or `isdtype`. Then, with those five changes made, I think this is good to go.

spec/API_specification/array_api/data_type_functions.py
kgryte: Renamed to `isdtype`. In general, unless a function is a contracted name (e.g., `vecdot` or `matmul`, etc), the specification prefers using underscores to separate distinct words. The only functions in the specification which do not include underscores separating two words are:

- `asarray`
- `astype`
- `isfinite`
- `isinf`
- `isnan`

Given the precedent of `isfinite`, `isinf`, and `isnan`, I've updated this PR to use `isdtype`, preferring consistency over PEP8 purity.

kgryte:
I've updated the PR to incorporate the proposed changes 1-4, and I've renamed the API to `isdtype` to be consistent with `isfinite`, `isinf`, and `isnan`. Both `isdtype` and `is_dtype` were considered acceptable based on feedback here and in consortium meetings; the former was chosen for consistency with existing specification APIs.

This PR should be ready for another round of review.

Issue #501, opened 2022-10-31
Clarify extensions should live in specified namespaces


honno:
I noticed our newly introduced FFT stubs live in `fourier_transform_functions.py`, which might implicitly communicate those extensions should live in the `xp.fourier_transform_functions` namespace (`xp` being the top-level Array API namespace), where presumably we want adopters to use `xp.fft` like say NumPy has it.

So that can be renamed, but this brings up that there's no explicit clarification in the spec that the `linalg` and `fft` extensions should indeed live in those respective namespaces. I think a note could live in all extension pages (currently just [Linear Algebra](https://data-apis.org/array-api/draft/extensions/linear_algebra_functions.html) and [Fourier transform Function](https://data-apis.org/array-api/draft/extensions/fourier_transform_functions.html)) explaining that the respective functions should be present in these specific namespaces.

honno:
In todays meeting we were in agreement for both points (`fourier_transform_functions.py -> fft.py` and the notes), so I'll submit a PR.

honno:
I completely missed #470, which aims to do the latter point (note that these extensions must live in specified namespaces).

rgommers:
Fixed by gh-470

PR #504, opened 2022-11-04
Rename `fourier_transform_functions.py` to `fft.py`, update respective uses

honno:
#501 related. In the consortium meeting yesterday we discussed and agreed `fourier_transform_functions.py` should be renamed to `fft.py`, implicitly implying the desired extension namespace (`xp.fft` as opposed to `xp.fourier_transform_functions`), hence this PR. cc @leofang 

I've not tackled noting explicitly where extensions live, as @rgommers is tackling that in #470.


leofang (approved):
Thanks, Matthew, LGTM.

rgommers (approved):
LGTM too, thanks Matthew.

Issue #488, opened 2022-09-29
Keep older versions of the spec in the repo's `HEAD`


honno:
Currently versioning in our docs works building pages for released versions (i.e. just `2021.12` for now) a final time, to be linked to later. So in the [`gh-pages` branch](https://github.com/data-apis/array-api/tree/gh-pages), you have the `2021.12/` folder unchanged after we tagged this repo with [`2021.12`](https://github.com/data-apis/array-api/tree/2021.12), and a `draft/` folder which now gets updated along this repo. It makes sense for projects to do this and seems an ubiquitous practice, as you rarely want to care about modifying older docs.

I was wondering however if we'd want to take a different approach: docs for every released version of the spec always live in the repo's `HEAD`, alongside the draft. Such an approach would certainly have us have a separate [stubs](https://github.com/data-apis/array-api/tree/main/spec/API_specification/array_api) folder/sub-module, e.g.

```
.
  spec/
    ...
    API_specification/
      ...
      array_api/
        draft/
          __init__.py
          _types.py
          _array_object.py
          ...
        2021.12/
          __init__.py
          _types.py
          _array_object.py
          ...
```

and probably will require separate folders for the rest of the RST in [`./spec/API_specification`](https://github.com/data-apis/array-api/tree/main/spec/API_specification), depending on how you go about things.

The advantage to this comes back to "caring about modifying older docs"—a spec standard probably does care more about "backports" than your typical software package. Currently we can't backport non-API changes to our `2021.12` docs, where it'd be nice to include clarifications and other niceties such as:

* https://github.com/data-apis/array-api/pull/464
* https://github.com/data-apis/array-api/pull/480
* https://github.com/data-apis/array-api/pull/430 (i.e. it would be nice to keep urls and intersphinx-refs consistent accross versions)

Another advantage is if we want to release the stubs as a package (https://github.com/data-apis/array-api/issues/411), where @asmeurer bought up the idea of keeping older versioned stubs in the repo so we could just publish the spec as-is on PyPI https://github.com/data-apis/array-api/pull/472#issuecomment-1219868893. Package or not, having the stubs of past versions live on `HEAD` is useful for the [compliance suite](https://github.com/data-apis/array-api-tests/) too.

All in all, this might not be a big deal? And I do see value if the docs for a released version are immutable.

[...]

PR #505, opened 2022-11-08
Keep documentation for previous spec versions alongside `draft`

honno:
WIP PR to resolve https://github.com/data-apis/array-api/issues/488. This PR keeps the source of the `2021.12` spec in the `HEAD` commit of the repo, which is important for backporting features and cross-version tooling (i.e. [`array-api-tests`](https://github.com/data-apis/array-api-tests/)).

**Note on the diff:** it's bloated by having soft-duplicated the docs/stubs so we can have `2021.12` and `draft` versions. Importantly files for both versions preserve their history ([solution used](https://stackoverflow.com/questions/74365771/copy-an-older-version-of-an-existing-file-into-a-seperate-file-which-preserves-h)), so you can `git blame` the `2021.12` files like they were on release day. The bloated diff otherwise seems inevitable, as `git` (or at least GitHub) can't mark multiple files as renamed, even if there are no changes to either file :confused: 

It won't be up-to-date but you can check https://honno.dev/array-api/ for rendered docs. I've also updated live previews, but that won't work for this PR as it requires the default branch to be updated too (does [work for `honno/array-api`](https://app.circleci.com/pipelines/github/honno/array-api)).

## What this PR changes

This introduces a module `array_api_stubs` which holds the signature stubs of *all* versions, which lives in `src/` and can be installed via `pip install .`. Currently you can see in the `main` branch only draft stubs are available, living in `spec/API_specification/array_api` (this PR moves them to `src/array_api_stubs/_draft/`).

To hold the non-docstring documentation of all versions, `spec/` now contains technically separate Sphinx docs for each version in respective folders (i.e. `spec/2021.12/` and `spec/draft/`). Sphinx configuration is however mostly shared, as each `conf.py` imports the same `_array_api_conf` module, i.e.

```python
# spec/draft/conf.py
import sys

from array_api_stubs import _draft as stubs_mod
from _array_api_conf import *

release = "DRAFT"
sys.modules["array_api"] = stubs_mod
# EOF
```

The main drive in having technically separate docs is to keep utilising Sphinx's autodoc, which is incredibly useful but ill-equipped of having multiple spec versions in the repo. The ability to monkey-patch `sys.modules` in the `conf.py` of every version-doc gives us consistency, basically just so we don't have to worry about modifying the `.. currentmodule` targets for every version.

Building the spec website is now fully reproducible locally via the introduced `make build` command (I wrote a quickstart for building the spec in `README.md`). Before we would just update the current version folder in `gh-pages`, which is sensible for most docs, but feels a bit loosey-goosey if we want the ability to backport things and generally it's nice to build the whole website locally. The pages workflow now works with `pip install .[doc]` and `make build`, but I have yet to update the preview workflow. `sphinx-(auto)build` will still serve us well when working on just the draft version.

## TODOs

* [x] Preserve git history for `2021.12` spec/stub files
* [x] Working site previews again
* [x] Remove temporary commits (namely used to bypass CI workflow restraints)
* [x] Document procedure on spec releases

(So beware, there's going to be a fair few force-pushes here.)

<details>
<summary><b>Follow-up PRs</b></summary>

First point of order is known backport candidates. On the top of my head:

* https://github.com/data-apis/array-api/pull/430 had to be backported for this PR to work, so follow-up should be redirecting `signatures.<file>.*` to respective `array_api.*` urls, just for the `2021.12` spec so links don't break.
* https://github.com/data-apis/array-api/pull/464
* https://github.com/data-apis/array-api/pull/480

And then I would want the package to actually be useful by tooling, notably `array-api-tests`—see https://github.com/data-apis/array-api/pull/472 for what I've already worked on in this regard. Another PR could automate releases (probably CalVer'd) anytime a workflow detects the package has changed (inspired by Hypothesis here, and for our purposes should be simple). I've already upload the package to PyPI as [`array-api-stubs`](https://pypi.org/project/array-api-stubs/) (just to park it), so any consortium member who's interested, feel free to ask me to make you a collaborator on PyPI.

Also, CircleCI doesn't handle that `index.html` magic that happens when you go to "folder" URL, i.e. `https://output.circle-artifacts.com/output/job/kfc/artifacts/0/_site/latest/` won't redirect to `.../index.html` and so give us a seamless preview when checking non-draft versions. Generally site previews seems to be an unintended but popular use of artifacts ([related discussion](https://discuss.circleci.com/t/circle-artifacts-com-not-using-index-html/320)). A way to handle this, probably by making the version jumping explicitly suffix hrefs with `index.html`, would be nice.

</details>

---

So that's broad strokes. Obviously lots of little details but I didn't want to make the PR body an automatic TL;DR heh... hopefully those details speak to themselves, so if not I have/should/will be writing documentation (ofc still working on this PR). Both excited and terrified to hear what folks think :sweat_smile:


README.md
```diff
@@ -15,6 +15,35 @@ These are relevant documents related to the content in this repository:
...
+
+## Building docs locally
+
+The spec website comprises of multiple Sphinx docs (one for each spec version),
+all of which exist in `spec/` and rely on the modules found in `src/` (most
+notably `array_api_stubs`). To install these modules and the additional
+dependencies of the Sphinx docs, you can use
+
+```sh
+$ pip install -e .[doc]  # ensure you install the dependencies extra "doc"
```
asmeurer, line 27: Is it necessary to install the stubs? I would expect the respective conf.py files to add the stubs to the PYTHONPATH automatically.

spec/_gitignore.txt
```diff
@@ -0,0 +1,33 @@
+#/
```
asmeurer, line 1: Why is this _gitignore.txt instead of .gitignore?

spec/index.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
```
asmeurer, line 1: Why is this file here? This seems like something that should only be on gh-pages. Also, redirects like this can be generated automatically with sphinx-reredirects.

asmeurer:
I would add a README to the spec directory explaining how the layout works and what steps need to be taken to make a spec release (if there are a lot of them, maybe it could be put into a script). Also, make sure we do test making a release to make sure everything works.

src/_array_api_conf.py
```diff
@@ -1,30 +1,25 @@
-# Configuration file for the Sphinx documentation builder.
```
asmeurer, line 1: Why is this file in `src/`? Shouldn't it be in `spec/`?

honno: It gets installed with `pip install .`. Seemed nicer than adding it to `spec/` and making relative imports, given you then you're in the territory of having to treat `spec/` and its sub-folders as Python modules.

Makefile
```diff
@@ -0,0 +1,23 @@
...
+	-rm -rf $(BUILDDIR)
+	-rm -rf $(SOURCEDIR)/**/generated
+
+build:
+	-mkdir -p $(BUILDDIR)
+	-cp "$(SOURCEDIR)/_gitignore.txt" "$(BUILDDIR)/.gitignore"
+	-cp "$(SOURCEDIR)/versions.json" "$(BUILDDIR)/versions.json"
+	-cp "$(SOURCEDIR)/index.html" "$(BUILDDIR)/index.html"
+	-touch "$(BUILDDIR)/.nojekyll"
+	-sphinx-build "$(SOURCEDIR)/2021.12" "$(BUILDDIR)/2021.12" $(SPHINXOPTS)
```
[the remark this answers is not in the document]
honno: Lower Makefiles already removed. The ugliness here isn't from building the docs so much as copying all the other stuff required for site building IMO.

spec/_gitignore.txt
honno: It's to be copied in the `gh-pages` branch. As you suggested in https://github.com/data-apis/array-api/pull/505#issuecomment-1314506373, something to add to a `spec/README.md`, or just document in `_gitignore.txt`.

spec/index.html
honno: To be copied to `gh-pages` from the `make build` command. As mentioned in OP, I want to keep the build process fully reproducible. Need to explore `sphinx-reredirects` generally, but not applicable for this file as it exists outside a Sphinx doc (i.e. `data-apis.org/array-api/index.html`, not `data-apis.org/array-api/{2021.12/latest/draft}/index.html`. Again, something for a `spec/README.me` as you say.

README.md
asmeurer: Personally I'd rather have a little more complexity in conf.py if it means the build process is easier (i.e., doesn't require installing the package first).

src/_array_api_conf.py
asmeurer: I find it somewhat confusing to do it this way. You generally don't install Sphinx conf files.

README.md
asmeurer: Plus adding stuff to sys.path in conf.py is not all that uncommon. The sys.modules hacks are just a necessary evil due to autosummary being very limited in its configurability. 

Aside: I've spend an inordinate amount of time in the past trying to figure out how to make autosummary not show the full module path for functions, see https://github.com/Chilipp/autodocsumm/issues/66 and https://github.com/sympy/sympy/pull/22589. Messing around with sys.modules is an idea I hadn't considered before. If it works, we should stick with it even if it feels hacky, because it's unlikely we'd find any other solution.

src/_array_api_conf.py
honno: Right, but you also generally don't keep contents of multiple doc versions in a repo. Something probably has to give here :sweat_smile:

README.md
honno: > Personally I'd rather have a little more complexity in conf.py if it means the build process is easier (i.e., doesn't require installing the package first).

I think actually we don't benefit much from magic here, as users still need to install the doc dependencies. Documenting in the README to use a single command `pip install -e .[doc]` doesn't seem too bad, even if just having to run `pip install -r requirements.txt` is more intuitive and what we're used to.

Having `sys.path.append("../../array_api_stubs/YYYY_MM/"); import YYYY_MM; sys.modules["array_api"] = YYYY_MM` for all `conf.py` files isn't much a problem and I see not uncommon, but it's still something in terms of maintenance burden and environment hacks. Promoting just a single way to work on both the `array_api_stubs` package and the spec documentation feels a bit nicer IMO as it's more true-to-form for a Sphinx doc and doesn't confuse what the stubs actually are.

honno: Also I like the modern `pyproject.toml` process of the `src/` folder which can contain multiple modules. Right now that's just `_array_api_conf`, which gets imported to each `conf.py` and saves us from path hacks per `conf.py`.

I'd also like to utilise `src/` for a small Python script which can be run to build redirect pages in the future (unfortunately `sphinx-reredirects` isn't workable for what we need). Even if we don't do that specifically, it's nice to have the flexibility (i.e. the local development instructions won't change if we introduce a module).

rgommers:
Let's hold off on this refactor until the changes to the standard that modify content (new functions, keywords, etc.) are done for `v2022`. Then we'll have a period of doing all these larger refactor/reshuffle PRs before finalizing `v2022`.

honno:
Alright, I'll need to sleep on it but I think I'm happy with this PR. It was kind of a nightmare getting this PR updated post the `2022.12` release—namely because typical rebase/merge workflows completely broke down with an already magic'd git history—but I *think* it's now up-to-date. I'm more confident now on these git things, but I'd appreciate if there was a freeze for merging other PRs until this one gets through, assuming we're close enough to merging :sweat_smile:

PACKAGE.md
```diff
@@ -0,0 +1,9 @@
+# Stubs for the array API standard
+
+Documentation specific to singular Python objects in the spec (i.e. functions,
```
kgryte, line 3:
```suggestion
Documentation specific to singular Python objects in the spec (i.e., functions,
```

PACKAGE.md
```diff
@@ -0,0 +1,9 @@
...
+methods and attributes) are infact represented by stub objects in the package
```
kgryte, line 4:
```suggestion
methods and attributes) are in fact represented by stub objects in the package
```

README.md
```diff
@@ -15,6 +15,122 @@ These are relevant documents related to the content in this repository:
 See [CONTRIBUTING.md](CONTRIBUTING.md) for how to go about contributing to
 this array API standard.
 
+
+## Building docs locally
+
+The spec website comprises of multiple Sphinx docs (one for each spec version),
```
kgryte, line 21:
```suggestion
The spec website is comprised of multiple Sphinx docs (one for each spec version),
```

README.md
```diff
@@ -15,6 +15,122 @@ These are relevant documents related to the content in this repository:
...
+
+To build specific versions of the spec, run `sphinx-build` on the respective
+folder in `spec/`, e.g.
+
+```sh
+$ sphinx-build spec/draft/ _site/draft/
+```
+
+To build the whole website, which includes every version of
+the spec, you can utilize the `make` commands defined in `spec/Makefile`, e.g.
```
kgryte, line 38:
```suggestion
the spec, you can utilize the `make` commands defined in `spec/Makefile`; e.g.,
```

src/_array_api_conf.py
```diff
@@ -1,30 +1,25 @@
...
-import os
-import sys
 import sphinx_material
-sys.path.insert(0, os.path.abspath('./API_specification'))
 
 # -- Project information -----------------------------------------------------
 
 project = 'Python array API standard'
-copyright = '2020-2022, Consortium for Python Data API Standards'
+copyright = '2020, Consortium for Python Data API Standards'
```
kgryte, line 20: Why was this changed?

kgryte (changes requested):
Overall, looks good. I reviewed each file. A few minor nitpicks/questions.

I'll leave it to you, @honno, to address, as may depend on what sort of Git magic is required. Additionally, would be good to know whether merging this PR requires any special consideration (e.g., whether we can squash commit or not).

src/array_api_stubs/__init__.py
```diff
@@ -0,0 +1,2 @@
+from . import _2021_12, _2022_12, _draft
```
asmeurer, line 1: If the idea is that these are public submodules, I don't think we should start them with underscores.

honno: Ah so I've left stubs packaging to a follow-up PR (see collapsed Follow-up PRs section in the submission post). I wouldn't want these modules public anywho, instead opting for an API more aligned with #472.

src/_array_api_conf.py
honno: Missed this when going through merge conflicts likely, so reverted this change.

honno:
> as may depend on what sort of Git magic is required. 

In case there's anything else, it should be fine to push commits on this PR, only the history shouldn't be changed without consideration.

> Additionally, would be good to know whether merging this PR requires any special consideration (e.g., whether we can squash commit or not).

Yeah this can't be squashed in as it would break the preserved blame history. Either  "Create merge commit" (i.e. the default for repos unless modified in settings) or "Rebase and merge" will be fine.

src/array_api_stubs/__init__.py
kgryte: Why don't we want the stubs modules public?

kgryte (approved):

PR #511, opened 2022-11-17
Fix description for `trunc`

kgryte:
This PR

- fixes the description for `trunc`. Currently, the description is incorrect for negative numbers (e.g., according to the description, `-1.5` should round to `-2.0` rather than `-1.0`).
- this was an oversight during review of [gh-12](https://github.com/data-apis/array-api/pull/12/files#diff-c63388022ee724b37174b4191598444df80aacb366d1f63677da4d14784df654R602).


honno (approved):
Nice catch, LGTM.

PR #513, opened 2022-11-21
Update guidance concerning mixing arrays with Python scalars to handle `complex`

kgryte:
This PR

- updates the guidance concerning mixing arrays with Python scalars. The current specification accommodates `bool` and real-valued data types. This PR updates the guidance to accommodate mixing a `complex` Python scalar with arrays. Guidance is intentionally limited to when a scalar is `complex` and an array has a complex floating-point array data type.
- does **not** specify what happens when a scalar is `complex` and an array has a real-valued array data type. Behavior is left implementation-specific (e.g., promotion, exception, etc). Given that current guidance requires converting a scalar to the array data type, to accommodate `complex` scalars and real-valued arrays, we'd need to specify casting rules for `complex` to real, which is something the specification has thus far avoided (e.g., `astype`, `real`, `imag`, `conj`, et al). This PR aims to be consistent with those prior decisions.

### DIFF
--- a/spec/API_specification/type_promotion.rst
+++ b/spec/API_specification/type_promotion.rst
@@ -123,7 +123,7 @@ Notes
 Mixing arrays with Python scalars
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Using Python scalars (i.e., instances of ``bool``, ``int``, ``float``) together with arrays must be supported for:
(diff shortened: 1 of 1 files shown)


rgommers (approved):
This looks logical and straightforward to me, so I'll merge it. Thanks @kgryte

PR #514, opened 2022-11-21
Add complex number support to `log`

kgryte:
This PR

-   adds complex number support to `log` by documenting special cases. By convention, the natural logarithm has a single branch cut, which is defined as the real interval `(-infinity, 0)`.
-   derives special cases from [C99](https://en.cppreference.com/w/c/numeric/complex/clog)

### DIFF
--- a/spec/API_specification/array_api/elementwise_functions.py
+++ b/spec/API_specification/array_api/elementwise_functions.py
@@ -933,7 +933,7 @@ def log(x: array, /) -> array:
     - If ``a`` is ``NaN`` and ``b`` is ``NaN``, the result is ``NaN + NaN j``.
 
     .. note::
-       The natural logarithm of a complex number :math:`z` with polar coordinates :math:`(r,\theta)` equals :math:`\ln r + (θ+ 2n\pi)j` with principal value :math:`\ln r + \theta j`.
+       The natural logarithm of a complex number :math:`z` with polar coordinates :math:`(r,\theta)` equals :math:`\ln r + (\theta + 2n\pi)j` with principal value :math:`\ln r + \theta j`.
 
     .. note::
        For complex floating-point operands, ``log(conj(x))`` must equal ``conj(log(x))``.
--- a/spec/API_specification/array_api/elementwise_functions.py
+++ b/spec/API_specification/array_api/elementwise_functions.py
@@ -904,18 +904,48 @@ def less_equal(x1: array, x2: array, /) -> array:
     """
 
 def log(x: array, /) -> array:
-    """
-    Calculates an implementation-dependent approximation to the natural (base ``e``) logarithm, having domain ``[0, +infinity]`` and codomain ``[-infinity, +infinity]``, for each element ``x_i`` of the input array ``x``.
+    r"""
+    Calculates an implementation-dependent approximation to the natural (base ``e``) logarithm for each element ``x_i`` of the input array ``x``.
 
     **Special cases**
 
-    For floating-point operands,
+    For real-valued floating-point operands,
 
     - If ``x_i`` is ``NaN``, the result is ``NaN``.
     - If ``x_i`` is less than ``0``, the result is ``NaN``.
     - If ``x_i`` is either ``+0`` or ``-0``, the result is ``-infinity``.
     - If ``x_i`` is ``1``, the result is ``+0``.
     - If ``x_i`` is ``+infinity``, the result is ``+infinity``.
+
+    For complex floating-point operands, let ``a = real(x_i)``, ``b = imag(x_i)``, and
+
+    - If ``a`` is ``-0`` and ``b`` is ``+0``, the result is ``-infinity + πj``.
+    - If ``a`` is ``+0`` and ``b`` is ``+0``, the result is ``-infinity + 0j``.
+    - If ``a`` is a finite number and ``b`` is ``+infinity``, the result is ``+infinity + πj/2``.
+    - If ``a`` is a finite number and ``b`` is ``NaN``, the result is ``NaN + NaN j``.
+    - If ``a`` is ``-infinity`` and ``b`` is a positive (i.e., greater than ``0``) finite number, the result is ``+infinity + πj``.
+    - If ``a`` is ``+infinity`` and ``b`` is a positive (i.e., greater than ``0``) finite number, the result is ``+infinity + 0j``.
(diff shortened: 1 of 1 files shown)


rgommers:
Sphinx is unhappy with the branch cuts label:
```
/home/circleci/repo/spec/API_specification/array_api/elementwise_functions.py:docstring of
array_api.elementwise_functions.log:41: WARNING: undefined label: branch-cuts
```

kgryte:
@rgommers This is because we need to merge https://github.com/data-apis/array-api/pull/461 first. The other branch cut PRs are dependent on that PR due to the added/referenced design doc.

kgryte:
Now that #461 is merged, all checks for this PR now pass.

spec/API_specification/array_api/elementwise_functions.py
```diff
@@ -904,28 +904,58 @@ def less_equal(x1: array, x2: array, /) -> array:
     """
 
 def log(x: array, /) -> array:
-    """
-    Calculates an implementation-dependent approximation to the natural (base ``e``) logarithm, having domain ``[0, +infinity]`` and codomain ``[-infinity, +infinity]``, for each element ``x_i`` of the input array ``x``.
```
rgommers, line 908: It looks like the domain/codomain part here is still relevant for real-valued complex numbers, but it seems to be deleted rather than moved. Is that on purpose?

rgommers (commented):
Thanks @kgryte. One minor comment on the diff. My main comment is about the branch cuts - it would be very helpful to document that they are consistent with (and based on) those from C99. It is noted in the branch cuts design document, however that is probably easy to miss (and it requires the user to know which functions are in C99 and which aren't). Maybe in a couple of words, "_branch cuts for `log` match those in C99"_?.

asmeurer:
For functions like this with a branch cut, should we add special cases along the cut, i.e., `a + 0j` and `a - 0j` where `a < 0`?

spec/API_specification/array_api/elementwise_functions.py
```diff
@@ -904,28 +904,58 @@ def less_equal(x1: array, x2: array, /) -> array:
...
+    - If ``a`` is ``+infinity`` and ``b`` is a positive (i.e., greater than ``0``) finite number, the result is ``+infinity + 0j``.
+    - If ``a`` is ``-infinity`` and ``b`` is ``+infinity``, the result is ``+infinity + 3πj/4``.
+    - If ``a`` is ``+infinity`` and ``b`` is ``+infinity``, the result is ``+infinity + πj/4``.
+    - If ``a`` is either ``+infinity`` or ``-infinity`` and ``b`` is ``NaN``, the result is ``+infinity + NaN j``.
+    - If ``a`` is ``NaN`` and ``b`` is a finite number, the result is ``NaN + NaN j``.
+    - If ``a`` is ``NaN`` and ``b`` is ``+infinity``, the result is ``+infinity + NaN j``.
+    - If ``a`` is ``NaN`` and ``b`` is ``NaN``, the result is ``NaN + NaN j``.
+
+    .. note::
+       The natural logarithm of a complex number :math:`z` with polar coordinates :math:`(r,\theta)` equals :math:`\ln r + (θ+ 2n\pi)j` with principal value :math:`\ln r + \theta j`.
```
asmeurer, line 936:
```suggestion
       The natural logarithm of a complex number :math:`z` with polar coordinates :math:`(r,\theta)` equals :math:`\ln r + (\theta + 2n\pi)j` with principal value :math:`\ln r + \theta j`.
```

kgryte: Thanks for catching!

kgryte:
@asmeurer That should already be covered by https://github.com/data-apis/array-api/pull/514/files#diff-6a52ec5810cf6d5389f09ea49af4311f7dc38da86d9c091e2a0234dfc5f662cbR939.

asmeurer:
Is it worth specially noting that functions like log, sqrt, etc. are not expected to promote floating-point arguments outside of their real domain to complex (and indeed, such value-based casting is explicitly recommended against)?

asmeurer:
> @asmeurer That should already be covered by [#514 (files)](https://github.com/data-apis/array-api/pull/514/files#diff-6a52ec5810cf6d5389f09ea49af4311f7dc38da86d9c091e2a0234dfc5f662cbR939).

Did you mean to link to a specific line? I don't see where this is mentioned.

kgryte:
> Did you mean to link to a specific line?

From your [comment](https://github.com/data-apis/array-api/pull/514#issuecomment-1333407069), seemed like you were referring to how complex conjugates should be specified.

kgryte:
> Is it worth specially noting that functions like log, sqrt, etc. are not expected to promote floating-point arguments outside of their real domain to complex (and indeed, such value-based casting is explicitly recommended against)?

Yes, this is likely a good idea. Probably better to include as part of the general complex number design doc.

asmeurer:
I mean specifically special cases along the cut, for +0 and -0. Something like "If `a` is negative and `b` is `+0.0`, `log(a + bj)` should equal `log(-a) + π`" and "If `a` is negative and `b` is `-0.0`, `log(a + bj)` should equal `log(-a) - π`". That's why I asked in the last meeting if all the cuts where either on the real or imaginary axes, because when they are you can always pick a side of the cut with positive or negative 0.

Also, a weird thing I just discovered: `1-0.0j` doesn't create a complex number with `-0.0` imaginary part. You have to do `-(-1+0j)` or `complex(1, -0.0)`.

rgommers:
> My main comment is about the branch cuts - it would be very helpful to document that they are consistent with (and based on) those from C99. It is noted in the branch cuts design document, however that is probably easy to miss (and it requires the user to know which functions are in C99 and which aren't). Maybe in a couple of words, "_branch cuts for `log` match those in C99"_?.

This comment applies to a large number of open PRs, so I will add it as a separate task in gh-533.

asmeurer:
Hmm, OK. What about functions whose branch cuts are along the imaginary axis?

Anyway, let's make sure this is tested properly when we add it to the test suite (specifically, with +/- 0 being tested) @honno

asmeurer:
Just to be clear, I'm OK with this PR. My question applies to some other functions, but it's not something that should block things, since it's just about a special case which could easily be added later.

rgommers:
Thanks @asmeurer for the clarification and the new issue. Looks like we're all good here then.

PR #517, opened 2022-11-27
Add complex number support to `acos`

kgryte:
This PR

-   adds complex number support to `acos` by documenting special cases. By convention, the inverse cosine has a branch cut placed at the line segments `(-infinity, -1)` and `(1, infinity)` of the real axis.
-   derives special cases from [C99](https://en.cppreference.com/w/c/numeric/complex/cacos)


rgommers (approved):
LGTM, thanks @kgryte

PR #520, opened 2022-11-27
Add complex support to `acosh`

kgryte:
This PR

-   adds complex number support to `acosh` by documenting special cases. By convention, the inverse hyperbolic cosine has a branch cut placed at the line segment `(-infinity, 1)` of the real axis.
-   derives special cases from [C99](https://en.cppreference.com/w/c/numeric/complex/cacosh)


rgommers (approved):
LGTM, thanks @kgryte

PR #521, opened 2022-11-27
Add complex support to `asin`

kgryte:
This PR

-   adds complex number support to `asin` by documenting special cases. By convention, the inverse sine has a branch cut placed at the line segments `(-infinity, -1)` and `(1, infinity)` of the real axis.
-   derives special cases from [C99](https://en.cppreference.com/w/c/numeric/complex/casin)


rgommers (approved):
LGTM, thanks @kgryte

PR #522, opened 2022-11-27
Add complex support to `asinh`

kgryte:
This PR

-   adds complex number support to `asinh` by documenting special cases. By convention, the inverse hyperbolic sine has a branch cut placed at the line segments `(-infinity j, -j)` and `(j, infinity j)` of the imaginary axis.
-   derives special cases from [C99](https://en.cppreference.com/w/c/numeric/complex/casinh)


rgommers (approved):
LGTM, thanks @kgryte

PR #528, opened 2022-11-28
Add complex number support to `equal`

kgryte:
This PR

-   adds complex number support to `equal` by documenting special cases. By convention, equality is determined by independently comparing real and imaginary components respectively and then performing a logical AND (i.e., if `x = a + bj` and `y = c + dj`, then `x == y` iff `a == c AND b == d`.
-   follows precedent in Python, Julia, NumPy, and elsewhere.
-   does not embrace the one-infinity model in which all infinities in the complex plane map to a single infinity (see Riemann sphere), as doing so would lead to unintuitive results when casting real-valued infinities to complex infinities.
-   adds a design document discussing complex number equality comparison and general considerations for specifying complex number behavior within the specification.

## References

- Issue posted to Julia issue tracker concerning closed infinities: <https://github.com/JuliaLang/julia/issues/9790>
- Domain errors involving complex numbers: <https://github.com/JuliaLang/julia/issues/5234>
- Critique of C99 complex number behavior: <https://medialibrary.uantwerpen.be/oldcontent/container2652/files/papers/Ba_Cu_con_05.pdf>
- ISO/IEC 9899:1999: <https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf>
- Closeness of complex numbers: <https://github.com/numpy/numpy/issues/15959>


spec/API_specification/array_api/array_object.py
```diff
@@ -328,9 +328,31 @@ def __dlpack_device__(self: array, /) -> Tuple[Enum, int]:
...
+
+        For real-valued floating-point operands,
+
+        - If ``x1_i`` is ``NaN`` or ``x2_i`` is ``NaN``, the result is ``False``.
+        - If ``x1_i`` is ``+infinity`` and ``x2_i`` is ``+infinity``, the result is ``True``.
+        - If ``x1_i`` is ``-infinity`` and ``x2_i`` is ``-infinity``, the result is ``True``.
+        - If ``x1_i`` is ``-0`` and ``x2_i`` is either ``+0`` or ``-0``, the result is ``True``.
+        - If ``x1_i`` is ``+0`` and ``x2_i`` is either ``+0`` or ``-0``, the result is ``True``.
+        - If ``x1_i`` is a finite number, ``x2_i`` is a finite number, and ``x1_i`` equals ``x2_i``, the result is ``True``.
+        - In the remaining cases, the result is ``False``.
```
rgommers, line 346: This is 100% duplicate content, please refer to `equal` instead.

rgommers (changes requested):
The discussion on complex number equality looks good to me (I have added a note on refactoring the content regarding complex numbers into a single doc to gh-533). The specified behavior matches current array library implementations, so we should be all good there.

It'd be good to fix the duplicate content issue here before merging, easiest to do it now.

spec/API_specification/array_api/array_object.py
kgryte: Yes, we chose to duplicate content, where applicable, early on during spec writing, so this is the case for all relevant APIs. This should be fine to do as a single pass for all applicable array object APIs.

rgommers: This strategy creates a bunch of extra work for ourselves for no real gain, so not ideal - but I'll ignore it in the open PRs. I'll go ahead and merge this.

rgommers (approved):

PR #529, opened 2022-11-28
Add complex number support to `not_equal`

kgryte:
This PR

-   adds complex number support to `not_equal` by documenting special cases. By convention, inequality is determined by independently comparing real and imaginary components respectively and then performing a logical OR (i.e., if `x = a + bj` and `y = c + dj`, then `x != y` iff `a != c OR b != d`.
-   follows precedent in Python, Julia, NumPy, and elsewhere.
-   does not embrace the one-infinity model in which all infinities in the complex plane map to a single infinity (see Riemann sphere), as doing so would lead to unintuitive results when casting real-valued infinities to complex infinities.
-   this PR depends on #528 which includes a design document which this PR references.

### DIFF
--- a/spec/API_specification/array_api/array_object.py
+++ b/spec/API_specification/array_api/array_object.py
@@ -731,6 +731,26 @@ class _array():
         """
         Computes the truth value of ``self_i != other_i`` for each element of an array instance with the respective element of the array ``other``.
 
+        **Special Cases**
+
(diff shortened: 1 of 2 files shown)


rgommers (approved):
Content seems fine, duplication will be cleaned up later. LGTM, in it goes. Thanks @kgryte

PR #534, opened 2022-12-01
Add complex number support to `log1p`

kgryte:
This PR

-   adds complex number support to `log1p` by documenting special cases. By convention, the natural logarithm has a single branch cut, which is defined as the real interval `(-infinity, 0)`.
-   derives special cases from those described in <https://github.com/data-apis/array-api/pull/514>.


rgommers (approved):
This LGTM, thanks @kgryte

PR #535, opened 2022-12-01
Add complex number support to `log2`

kgryte:
This PR

-   adds complex number support to `log2` by documenting special cases.
-   Specifies that special cases should be derived from those described in <https://github.com/data-apis/array-api/pull/514> according to the textbook change of base formula.


rgommers (approved):
This is consistent with `log` (reviewed in gh-514), so LGTM. In it goes. Thanks @kgryte

PR #537, opened 2022-12-01
Add complex number support for `pow`

kgryte:
This PR

-   adds complex number support to `pow` by documenting special cases. This PR indicates that implementations should handle special cases as if `pow` is implemented `exp(x2*log(x1))`, following C99; however, implementations are allowed to handle special cases more carefully. Accordingly, `pow` inherits a branch cut from `log`. In general, `pow` is underspecified, as handling all special cases is difficult given the number of argument permutations.
-   derives special cases from [C99](https://en.cppreference.com/w/c/numeric/complex/cpow)

## References

- CPython issue 15996: <https://bugs.python.org/issue15996> (and GitHub: <https://github.com/python/cpython/issues/60200>)
- CPython issue 44970: <https://bugs.python.org/issue44970> (and GitHub: <https://github.com/python/cpython/issues/89133>)
- ISO/IEC 10967-3:2006: _Language independent arithmetic — Part 3: Complex integer and floating point arithmetic and complex elementary numerical functions_

### DIFF
--- a/spec/API_specification/array_api/array_object.py
+++ b/spec/API_specification/array_api/array_object.py
@@ -814,17 +814,19 @@ class _array():
         """
 
     def __pow__(self: array, other: Union[int, float, array], /) -> array:
-        """
+        r"""
         Calculates an implementation-dependent approximation of exponentiation by raising each element (the base) of an array instance to the power of ``other_i`` (the exponent), where ``other_i`` is the corresponding element of the array ``other``.
 
         .. note::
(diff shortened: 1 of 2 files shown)


seberg:
Btw. do you happen to have a clear answer to what the branch-cuts are with respect to returning `-0` for the case of `0**something`?

kgryte:
@seberg For the case of `0**something`, I believe this should just follow IEEE 754, so long as `something` is real-valued. If `something` is a complex number, the branch cut from `log` would apply and the result would be according to `exp`.

rgommers (approved):
LGTM, thanks @kgryte

PR #540, opened 2022-12-01
Add complex number support to APIs for returning unique elements

kgryte:
This PR

-   adds complex number support to `unique_all`, `unique_counts`, `unique_inverse`, and `unique_values`. This PR builds on <https://github.com/data-apis/array-api/pull/528>. Namely, by convention, equality is determined by independently comparing real and imaginary components respectively and then performing a logical AND (i.e., if `x = a + bj` and `y = c + dj`, then `x == y` iff `a == c AND b == d`.
-   requires that complex numbers having at least one component which is `NaN` be considered unique.
-   as this PR builds on <https://github.com/data-apis/array-api/pull/528>, this PR does not support the one-infinity model (see C99) where complex infinites are treated as equal.


rgommers (approved):
LGTM, thanks @kgryte

PR #541, opened 2022-12-01
Add complex number support to `linalg.trace`

kgryte:
This PR

-   adds support for complex numbers to `linalg.trace` and follows the guidance set forth in <https://github.com/data-apis/array-api/pull/538>.
-   specifies that special cases be derived **as if** the operation was implemented as a reduction over `add`.
-   requires that, if `x` is a complex floating-point data type and `dtype` is `None`, the function must return an array having the default complex floating-point data type.


rgommers (approved):
This one is pretty straightforward, and looks consistent with `sum`. In it goes, thanks @kgryte

PR #542, opened 2022-12-01
Add complex number support to `linalg.det`

kgryte:
This PR

-   adds support for complex numbers to `linalg.det`.
-   updates the input and output array data types to be any floating-point data type, not just real-valued floating-point data types. In certain circumstances, the determinant of a complex matrix can be real-valued; however, this is not generally true. In which case, the guidance remains that the output data type be required to have the same data type as the input array `x`.


rgommers (approved):
LGTM, thanks @kgryte

Issue #403, opened 2022-03-12
Use case study: Intheon


chkothe:
We're very happy to see progress in getting the array API standardized -- obviously a monumental undertaking when all is considered.

We've been using backend-agnostic numpy-style code for over a year now in research and now production, and are gradually rolling it out across a multi-megabyte codebase (supporting mainly numpy, cupy, pytorch, tensorflow, jax (also dask, but not using that at the moment), so I thought I'd share our user story in case it's helpful.

Our codebase has an entrypoint equivalent to get `get_namespace`, although we call them backends, and a typical use case looks like:
```
def sqrt_sum(arr1, arr2):
    be = backend_for(arr1, arr2)
    return be.sqrt(be.asarray(arr1) + be.asarray(arr2))
```
For each of the backends we have a (usually thin) compatibility layer that adds any missing functions or fixes up issues with the function signature. In our case, `backend_for` looks at `__array_priority__` to return the namespace for the highest-priority array, although we rarely use it with more than one array (but it results in the above function accepting multiple array types and promoting according to np < dask < {jax, tf, cupy} < torch). Getting a namespace this way is about as fast as doing `.T` on a 1x1 numpy array (thanks to some caching) so we use it on even the smallest subroutine.

We use a lot of the API surface of these backends, and typically the most compute-intensive subroutines have an option to choose a preferred backend (via a function `backend_get(shorthand)`) out of the subset that supports the necessary ops, or to keep the same backend. We've been extremely impressed with the compatibility of cupy, the performance of torch (notably also its cpu arrays), and we have a few places where we prefer tf or jax because they might have a faster implementation of a critical op or parallelize better (e.g. jax). We find that even over a half-year time span, things evolved rapidly enough that the ideal backend pick for a function may change from one to another (one of the reasons we aim for that much flexibility).

We traced our API usage and found that perhaps 90% of our backend-agnostic call sites would be covered by the array API as it exists now. There are a few places that use different aliases for the same functionality (due to some forms being more popular with current backends than others) the most frequent issues being `absolute(x)` and `concatenate(x)`, other examples being `arccos(x)` (all our backends) vs `acos(x)` (no backend). We also frequently use convenience shorthands like `hstack(x)`, `vstack(x)`, `ravel(x)` or `x.flatten()`, but those could be substituted easily (or provided in a wrapper). 

[...]

PR #543, opened 2022-12-01
Add complex number support to `eigh` and `eigvalsh`

kgryte:
This PR

-   adds complex number support to `eigh` and `eigvalsh`.
-   follows PyTorch and NumPy in requiring that eigenvalues be returned as real-valued floating-point arrays.
-   requires that complex input arrays be Hermitian.
-   the precision of the eigenvalue array must match that of the input array.

### DIFF
--- a/spec/API_specification/array_api/linalg.py
+++ b/spec/API_specification/array_api/linalg.py
@@ -110,59 +110,80 @@ def diagonal(x: array, /, *, offset: int = 0) -> array:
     """
 
 def eigh(x: array, /) -> Tuple[array]:
-    """
-    Returns an eigendecomposition x = QLQᵀ of a symmetric matrix (or a stack of symmetric matrices) ``x``, where ``Q`` is an orthogonal matrix (or a stack of matrices) and ``L`` is a vector (or a stack of vectors).
+    r"""
+    Returns an eigenvalue decomposition of a complex Hermitian or real symmetric matrix (or a stack of matrices) ``x``.
+
+    If ``x`` is real-valued, let :math:`\mathbb{K}` be the set of real numbers :math:`\mathbb{R}`, and, if ``x`` is complex-valued, let :math:`\mathbb{K}` be the set of complex numbers :math:`\mathbb{C}`.
+
+    The **eigenvalue decomposition** of a complex Hermitian or real symmetric matrix :math:`x \in\ \mathbb{K}^{n \times n}` is defined as
+
+    .. math::
+       x = Q \Lambda Q^H
+
+    where :math:`Q^H` is the conjugate transpose when :math:`Q` is complex and the transpose when :math:`Q` is real-valued and :math:`\Lambda` is a diagonal matrix whose diagonal elements are the corresponding eigenvalues. When ``x`` is real-valued, :math:`Q` is orthogonal, and, when ``x`` is complex, :math:`Q` is unitary.
 
     .. note::
-       The function ``eig`` will be added in a future version of the specification, as it requires complex number support.
+       The eigenvalues of a complex Hermitian or real symmetric matrix are always real.
 
-    ..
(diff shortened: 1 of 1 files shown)


spec/API_specification/array_api/linalg.py
```diff
@@ -110,59 +110,82 @@ def diagonal(x: array, /, *, offset: int = 0) -> array:
...
+    Returns an eigenvalue decomposition of a complex Hermitian or real symmetric matrix (or a stack of matrices) ``x``.
+
+    If ``x`` is real-valued, let :math:`\mathbb{K}` be the set of real numbers :math:`\mathbb{R}`, and, if ``x`` is complex-valued, let :math:`\mathbb{K}` be the set of complex numbers :math:`\mathbb{C}`.
+
+    The **eigenvalue decomposition** of a complex Hermitian or real symmetric matrix :math:`x \in\ \mathbb{K}^{n \times n}` is defined as
+
+    .. math::
+       x = Q \Lambda Q^H
+
+    with :math:`Q \in \mathbb{K}^{n \times n}` and :math:`\Lambda \in \mathbb{R}^n` and where :math:`Q^H` is the conjugate transpose when :math:`Q` is complex and the transpose when :math:`Q` is real-valued and :math:`\Lambda` is a diagonal matrix whose diagonal elements are the corresponding eigenvalues. When ``x`` is real-valued, :math:`Q` is orthogonal, and, when ``x`` is complex, :math:`Q` is unitary.
```
rgommers, line 123: @kgryte regarding _"When ``x`` is real-valued, :math:`Q` is orthogonal, and, when ``x`` is complex, :math:`Q` is unitary."_, I'm wondering why the distinction. I tested with numpy and pytorch, and behavior seems to be the same. In this case the phrasing implies that eigenvectors for complex inputs are not orthogonal. For random inputs I do get orthogonal vectors. Not sure when that would not be the case.

spec/API_specification/array_api/linalg.py
```diff
@@ -110,59 +110,82 @@ def diagonal(x: array, /, *, offset: int = 0) -> array:
...
+       The eigenvectors of a symmetric matrix are not unique and are not continuous with respect to ``x``. Because eigenvectors are not unique, different hardware and software may compute different eigenvectors.
+
+       Non-uniqueness stems from the fact that multiplying an eigenvector by :math:`-1` when ``x`` is real-valued and by :math:`e^{\phi j}` (:math:`\phi \in \mathbb{R}`) when ``x`` is complex produces another set of valid eigenvectors.
 
     .. note::
-       Whether an array library explicitly checks whether an input array is a symmetric matrix (or a stack of symmetric matrices) is implementation-defined.
+       Whether an array library explicitly checks whether an input array is Hermitian or a symmetric matrix (or a stack of matrices) is implementation-defined.
+
+    .. note::
+       The function ``eig`` will be added in a future version of the specification.
```
rgommers, line 137: I refreshed my memory on this point. There is gh-113 and https://github.com/data-apis/array-api/issues/403#issuecomment-1074069254, so yes there seems to be demand and hence this note seems fine.

spec/API_specification/array_api/linalg.py
rgommers: This does match the PyTorch docs, so I think it did have some review/discussion.

rgommers (approved):
LGTM modulo my one question on (non-)orthogonality of eigenvectors.

spec/API_specification/array_api/linalg.py
kgryte: For clarity, an orthogonal matrix is the real analogue of a unitary matrix.

rgommers: Ah right, linalg lingo:)

Issue #544, opened 2022-12-01
Add support for computing the hypotenuse via `hypot`


kgryte:
This RFC requests to include a new API in the array API specification for the purpose of computing the hypotenuse.

Based on array comparison [data](https://github.com/data-apis/array-api-comparison/blob/7ecfa9867ec43a28e914b35868c041443d31d772/data/common_apis.csv#L93), the API is available across all major array libraries in the PyData ecosystem (note: TensorFlow implements `hypot` in its experimental `numpy` namespace).

While `hypot` can be implemented in userland using existing array API functions (e.g., `sqrt(x1*x1 + x2*x2)`), this function is typically implemented in a specialized manner to help prevent over/underflow.

Furthermore, need for this API arose when adding complex number support to the array API specification, particularly with regard to adding complex number support to `abs`, where the absolute value (magnitude) of a complex number is computed as `hypot(real(z), imag(z))`.

As this API already exists in the ecosystem and satisfies needs (e.g., numerical stability and complex number support), this API seems a good candidate for inclusion in the standard.

rgommers:
This seems like a reasonable addition, given it's already commonly implemented (the linked data shows it is named differently in MXNet though). Are the signatures consistent among libraries?

> Furthermore, need for this API arose when adding complex number support to the array API specification

That does not seem directly relevant here. For `abs` to use a hypot calculation internally, `hypot` does not have to be exposed as public API.

kgryte:
> That does not seem directly relevant here.

It does in order for the specification to be able to define special case behavior.

kgryte:
Concerning special case behavior, according to IEEE 754, `hypot` handles `NaN` differently than would be the case for the naive implementation. Namely, if only one of `x` or `y` in `hypot(x,y)` is `NaN` and the other is `infinity`, the result is `infinity`. This also applies to `abs` when computing the absolute value of a complex number (see C99).

Wrt complex number support, while we could explicitly spell out complex number special case behavior for `abs`, the end result is that all libraries would effectively have to implement IEEE 754 `hypot`.

kgryte:
In terms of API, all supporting libraries implement as a standard element-wise binary function: `hypot(x1, x2)`, where `x1` and `x2` follow conventional broadcast rules.

rgommers:
> It does in order for the specification to be able to define special case behavior.

As discussed last Thursday, the special cases for `abs` can be put in the `abs` description. This is implemented in gh-546. 

[...]

PR #546, opened 2022-12-05
Add complex number support to `abs`

kgryte:
This PR

-   adds complex number support to `abs` by documenting special cases. Special cases follow special cases for C99 `hypot`.
-   includes a TODO to update the specification once [hypot](https://github.com/data-apis/array-api/issues/544) is added to the specification. Ideally, implementations would implement the complex absolute value in terms of `hypot` in order to avoid undue overflow/underflow during intermediate stages of computation. For now, the specification includes a note instructing implementations to avoid such overflow/underflow.
-   requires that the return data type always be a real-valued data type.

### DIFF
--- a/spec/API_specification/array_api/array_object.py
+++ b/spec/API_specification/array_api/array_object.py
@@ -110,28 +110,48 @@ class _array():
 
     def __abs__(self: array, /) -> array:
         """
-        Calculates the absolute value for each element of an array instance (i.e., the element-wise result has the same magnitude as the respective element but has positive sign).
+        Calculates the absolute value for each element of an array instance.
+
+        For real-valued input arrays, the element-wise result has the same magnitude as the respective element in ``x`` but has positive sign.
 
         .. note::
            For signed integer data types, the absolute value of the minimum representable integer is implementation-dependent.
 
         **Special cases**
 
-        For floating-point operands, let ``self`` equal ``x``.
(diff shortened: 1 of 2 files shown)


spec/API_specification/array_api/array_object.py
```diff
@@ -110,28 +110,48 @@ def T(self: array) -> array:
...
+        - If ``b`` is either ``+0`` or ``-0``, the result is equal to ``abs(a)``.
+        - If ``a`` is ``NaN`` and ``b`` is a finite number, the result is ``NaN``.
+        - If ``a`` is a finite number and ``b`` is ``NaN``, the result is ``NaN``.
+        - If ``a`` is ``NaN`` and ``b`` is ``NaN``, the result is ``NaN``.
+
+        .. note::
+           For complex floating-point operands, conforming implementations should take care to avoid undue overflow or underflow during intermediate stages of computation.
+
+        ..
+           TODO: once ``hypot`` is added to the specification, remove the special cases for complex floating-point operands and the note concerning guarding against undue overflow/underflow, and state that special cases must be handled as if implemented as ``hypot(real(x), imag(x))``.
```
rgommers, line 144: Checked in the rendered html that this isn't visible, so should be fine to leave that in here.

rgommers (approved):
LGTM, thanks @kgryte

PR #547, opened 2022-12-05
Add complex number support for `linalg.inv`

kgryte:
This PR

-   adds complex number support to `linalg.inv`. The main substantive change is updating the input and output data types.
-   makes explicit that, for matrix stacks, the multiplicative inverse must be computed for each matrix in the stack.


rgommers (approved):
The change for complex number support is straightforward. The added math notation and explanations look helpful. LGTM, thanks @kgryte

PR #548, opened 2022-12-05
Add complex number support to `linalg.qr`

kgryte:
This PR

-   adds complex number support to `linalg.qr`. The main substantive change is updating the input and output data types.
-   makes explicit that, for matrix stacks, the QR decomposition must be computed for each matrix in the stack.

### DIFF
--- a/spec/API_specification/array_api/linalg.py
+++ b/spec/API_specification/array_api/linalg.py
@@ -315,16 +315,40 @@ def pinv(x: array, /, *, rtol: Optional[Union[float, array]] = None) -> array:
     """
 
 def qr(x: array, /, *, mode: Literal['reduced', 'complete'] = 'reduced') -> Tuple[array, array]:
-    """
-    Returns the qr decomposition x = QR of a full column rank matrix (or a stack of matrices), where ``Q`` is an orthonormal matrix (or a stack of matrices) and ``R`` is an upper-triangular matrix (or a stack of matrices).
+    r"""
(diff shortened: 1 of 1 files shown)


spec/API_specification/array_api/linalg.py
```diff
@@ -315,16 +315,42 @@ def pinv(x: array, /, *, rtol: Optional[Union[float, array]] = None) -> array:
...
+    where :math:`Q \in\ \mathbb{K}^{m \times n}` and :math:`R \in\ \mathbb{K}^{n \times n}`.
+
+    The reduced QR decomposition equals with the complete QR decomposition when :math:`n \qeq m` (wide matrix).
+
+    When ``x`` is a stack of matrices, the function must compute the QR decomposition for each matrix in the stack.
 
     .. note::
        Whether an array library explicitly checks whether an input array is a full column rank matrix (or a stack of full column rank matrices) is implementation-defined.
 
+    .. warning::
```
rgommers, line 344: This could be a note instead of a warning, in order to look less scary. That's a minor thing though, and perhaps a question of taste. So I'll leave it alone for now, just noting it in case someone revisits this PR in the future.

rgommers (approved):
Added content LGTM, thanks @kgryte

spec/API_specification/array_api/linalg.py
kgryte: Followed PyTorch in using a warning, instead of a note.

PR #549, opened 2022-12-05
Add complex number support to `linalg.matrix_power`

kgryte:
This PR

-   adds complex number support to `linalg.matrix_power`. The main substantive change is updating the input and output data types.
-   depends on <https://github.com/data-apis/array-api/pull/547>, which adds complex number support for matrix inversion.


rgommers (approved):
This is nicely straightforward, in it goes.

PR #550, opened 2022-12-05
Add complex number support to `linalg.vector_norm`

kgryte:
This PR

-   adds complex number support to `linalg.vector_norm`. The main substantive change is updating the input and output data types. For both real- and complex-valued inputs, norms (apart from `ord=0)` involve computing the absolute value, so no changes are necessary in norm definitions to accommodate complex numbers.
-   requires that the output data type always be a real-valued floating-point data type.


rgommers (approved):
This looks pretty straightforward, in it goes. Thanks @kgryte

PR #551, opened 2022-12-05
Add complex number support for multiplication

kgryte:
This PR

-   adds support for complex number multiplication. For finite components, complex number multiplication is well-defined, where `(a + bj) x (c + dj) = (ac - bd) + (bc + ad)j`. However, for complex numbers having infinite or `NaN` components, results may diverge depending on how one chooses to model complex infinities. In C99 with its one-infinity model, a complex value is infinite, even if the other component is `NaN`. In order to maintain a single point at infinity, C99 suspends the normal rules of arithmetic floating-point operations and requires branching logic within implementations in order to accurately handle special cases. Such branching logic is undesirable when performance is paramount (e.g., in certain accelerator libraries), and, thus, special cases involving `NaNs` and infinities is specified as implementation-dependent.
-   Specifies that, in general, the special cases described for real-valued floating-point operands apply to the component operations separately. Meaning, operations involving real-valued components (i.e., `a`, `b`, `c`, `d`) must abide by real-valued floating-point special cases. 

### DIFF
--- a/spec/API_specification/array_api/array_object.py
+++ b/spec/API_specification/array_api/array_object.py
@@ -717,7 +717,7 @@ class _array():
         +------------+----------------+-----------------+--------------------------+
         | **bj**     | (b*c)j         | -(b*d)          | -(b*d) + (b*c)j          |
         +------------+----------------+-----------------+--------------------------+
-        | **a + bj** | (a*c) + (b*c)j | -(b*d) + (a*d)j | implementation-dependent |
+        | **a + bj** | (a*c) + (b*c)j | -(b*d) + (a*d)j | special rules            |
         +------------+----------------+-----------------+--------------------------+
 
         In general, for complex floating-point operands, real-valued floating-point special cases must independently apply to the real and imaginary component operations involving real numbers as described in the above table.
--- a/spec/API_specification/array_api/array_object.py
+++ b/spec/API_specification/array_api/array_object.py
@@ -689,12 +689,14 @@ class _array():
         """
 
     def __mul__(self: array, other: Union[int, float, array], /) -> array:
-        """
+        r"""
         Calculates the product for each element of an array instance with the respective element of the array ``other``.
 
         **Special cases**
 
-        For floating-point operands, let ``self`` equal ``x1`` and ``other`` equal ``x2``.
+        Let ``self`` equal ``x1`` and ``other`` equal ``x2``.
+
+        For real-valued floating-point operands,
 
         -   If either ``x1_i`` or ``x2_i`` is ``NaN``, the result is ``NaN``.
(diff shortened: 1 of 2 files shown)


spec/API_specification/array_api/array_object.py
```diff
@@ -706,16 +708,42 @@ def __mul__(self: array, other: Union[int, float, array], /) -> array:
...
+        For complex floating-point operands, multiplication is defined according to the following table. For real components ``a`` and ``c`` and imaginary components ``b`` and ``d``,
+
+        +------------+----------------+-----------------+--------------------------+
+        |            | c              | dj              | c + dj                   |
+        +============+================+=================+==========================+
+        | **a**      | a * c          | (a*d)j          | (a*c) + (a*d)j           |
+        +------------+----------------+-----------------+--------------------------+
+        | **bj**     | (b*c)j         | -(b*d)          | -(b*d) + (b*c)j          |
+        +------------+----------------+-----------------+--------------------------+
+        | **a + bj** | (a*c) + (b*c)j | -(b*d) + (a*d)j | implementation-dependent |
```
rgommers, line 720: The ` implementation-dependent` here is confusing, and inconsistent with the PR description. Why not add `(ac - bd) + (bc + ad)j` here? The "if a component is nan/inf" is already handled by special-case text below, now it looks like the whole operation isn't well-defined.

rgommers (changes requested):
Thanks @kgryte. I have one comment on the table here that would be good to address.

spec/API_specification/array_api/array_object.py
rgommers: The same comment applies to division (gh-554)

kgryte: I updated the entry from `implementation-dependent` to `special rules`. The main purpose of the table is to describe when the normal rules for real-valued floating-point case special cases apply. As the normal rules don't apply for multiplication of two complex numbers, IMO adding the textbook formula for cmplx-cmplx would also be misleading given the content which immediately follows the table.

My concern is that if the table shows only the textbook formula without indication that cmplx-cmplx may vary between implementations, readers won't continue reading on to find out that special case handling doesn't always apply.

kgryte: Wrt complex division, I should note that the guidance following the table is that, for finite components, division "should" (not "must") be computed as if by the textbook formula, as there are algorithms, similar to `hypot`, for avoiding underflow/overflow. And for complex multiplication, one can use [Karatsuba's algorithm](https://en.wikipedia.org/wiki/Karatsuba%27s_algorithm) to reduce the number of multiplications. In short, I am hesitant to be too prescriptive in what is allowed/disallowed.

rgommers: "special rules" seems better, thanks. After that section is moved to notes, so it's clearer that the whole thing is special case handling related, it will be even better. Let's roll with this

rgommers (approved):
Thanks @kgryte

PR #552, opened 2022-12-05
Add complex number support to `square`

kgryte:
This PR

-   adds complex number support to `square`. As `square` is shorthand for `multiply(x, x)`, special cases should match that of `multiply` for both real- and complex-valued floating-point operands. 
-   depends on <https://github.com/data-apis/array-api/pull/551>.


rgommers (approved):
The PR this depended on is merged now, and this LGTM - so in it goes. Thanks @kgryte

PR #553, opened 2022-12-05
Add complex number support to `prod`

kgryte:
This PR

-   adds support for computing the product of complex numbers. As described in <https://github.com/data-apis/array-api/pull/551>, for complex numbers having finite components, complex number multiplication is well-defined; however, special cases involving `NaN` and infinite components may diverge depending on how one chooses to model complex infinity.
-   specifies that special cases be derived **as if** the operation was implemented as a reduction over `multiply`.
-   requires that, if `x` is a complex floating-point data type and `dtype` is `None`, the function must return an array having the default complex floating-point data type.
-   depends on <https://github.com/data-apis/array-api/pull/551>.


rgommers (approved):
The PR this depended on is merged now, and this LGTM - so in it goes. Thanks @kgryte

PR #554, opened 2022-12-05
Add support for complex number division

kgryte:
This PR

-   adds support for complex number division. For finite components, complex number division is well-defined and can be readily computed according to the textbook formula for complex number division. However, for complex numbers having infinite or `NaN` components, results may diverge depending on how one chooses to model complex infinities. In C99 with its one-infinity model, a complex value is infinite, even if the other component is `NaN`. In order to maintain a single point at infinity, C99 suspends the normal rules of arithmetic floating-point operations and requires branching logic within implementations in order to accurately handle special cases. Such branching logic is undesirable when performance is paramount (e.g., in certain accelerator libraries), and, thus, special cases involving `NaNs` and infinities is specified as implementation-dependent.
-   Specifies that, in general, the special cases described for real-valued floating-point operands apply to the component operations separately. Meaning, operations involving real-valued components (i.e., `a`, `b`, `c`, `d`) must abide by real-valued floating-point special cases. 

### DIFF
--- a/spec/API_specification/array_api/array_object.py
+++ b/spec/API_specification/array_api/array_object.py
@@ -946,7 +946,7 @@ class _array():
         """
 
     def __truediv__(self: array, other: Union[int, float, array], /) -> array:
-        """
+        r"""
         Evaluates ``self_i / other_i`` for each element of an array instance with the respective element of the array ``other``.
 
         .. note::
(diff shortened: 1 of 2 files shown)


rgommers (approved):
LGTM, thanks @kgryte

Issue #545, opened 2022-12-01
Computing the sign of a complex number


kgryte:
Originally discussed in <https://github.com/data-apis/array-api/issues/153#issuecomment-869059671>.

In order to add complex support to `sign`, we need to agree on the definition for complex numbers.

NumPy uses `sqrt(z^2)/z` while an arguably more common convention is `x/|x|`.

For paths forward we have 3 options:

1. Use NumPy's definition.
2. Use the more common convention used elsewhere (MATLAB, Julia, etc).
3. Not support complex numbers in `sign`.

## Questions

- If we choose (2), given NumPy's potential v2 release, would NumPy be willing to introduce this breaking change?

## See Also

- <https://github.com/numpy/numpy/issues/3621>
- <https://github.com/numpy/numpy/issues/13179>
- MATLAB: <https://www.mathworks.com/help/matlab/ref/sign.html>
- Julia: <https://docs.julialang.org/en/v1/base/math/#Base.sign>

rgommers:
As discussed in a couple of places, the `z/|z|` definition seems clearly better and preferred, also for numpy. Not sure of a timeline for introducing that (breaking changes are always tricky), but I think in this case it's not like they're two equal alternatives. In the standard I think we should go with `z/|z|`; numpy may not be compliant on that point for a while.

PR #556, opened 2022-12-12
Add complex number support to `sign`

kgryte:
This PR

-   adds complex number support to `sign`.
-   defines the sign to be `x/|x|` ensuring a consistent definition for real and complex data types (as discussed in <https://github.com/data-apis/array-api/issues/545>).
-   closes <https://github.com/data-apis/array-api/issues/545>.
-   requires that special cases follow the rules of complex number division. In this case, the numerator is a complex number and the denominator a real value and, thus, complex number division is well-defined.
-   updates the input array data type to be any numeric data type, not just real-valued data types.
-   adds a special case for real-valued inputs in that `NaN` should be returned if `x_i` is `NaN`.
-   depends on <https://github.com/data-apis/array-api/pull/554>.

### DIFF
--- a/spec/API_specification/array_api/elementwise_functions.py
+++ b/spec/API_specification/array_api/elementwise_functions.py
@@ -1356,14 +1356,20 @@ def sign(x: array, /) -> array:
 
     **Special cases**
 
+    For real-valued operands,
+
     - If ``x_i`` is less than ``0``, the result is ``-1``.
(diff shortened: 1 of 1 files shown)


rgommers (approved):
This looks consistent with what was discussed in the linked issue. LGTM, in it goes. Thanks @kgryte

PR #557, opened 2022-12-12
Add complex number support to `matmul`

kgryte:
This PR

-   adds complex number support to `matmul`.
-   requires that input arrays **not** be transposed or conjugated when arguments have complex floating-point data types. If `matmul` were to automatically compute the conjugate transpose of the second argument, this would allow returning the complex dot product for the 1D case. However, following NumPy et, this PR pushes conjugate-transposition to userland by requiring conjugation and/or transposition be explicitly performed prior to computing to the matrix product.


rgommers (approved):
That note looks correct; `vdot` (not in the array API standard), is the matrix multiplication with conjugation. In it goes, thanks @kgryte

PR #558, opened 2022-12-12
Add complex number support to `tensordot`

kgryte:
This PR

-   adds complex number support to `tensordot` (i.e., the generalized matrix product).
-   similar to `matmul` (see <https://github.com/data-apis/array-api/pull/557>), requires that input arrays **not** be transposed or conjugated when arguments have complex floating-point data types. If `tensordot` were to automatically compute the conjugate transpose of the second argument, this would allow returning the complex dot product for the 1D case when `N=1`. However, following NumPy et, this PR pushes conjugate-transposition to userland by requiring conjugation and/or transposition be explicitly performed prior to computing to the generalized matrix product.


rgommers (approved):
Looks correct, note and behavior are consistent with `matmul`. In it goes, thanks @kgryte

PR #559, opened 2022-12-12
Add complex number support to `linalg.cross`

kgryte:
This PR

-   adds complex number support to `linalg.cross`. No special considerations are required to accommodate complex numbers.
-   updates guidance to **require** a numeric data type. Previously, it said "should", but this was likely an oversight, as computing the cross product for non-numeric data types does not make sense.


rgommers (approved):
LGTM, thanks @kgryte

PR #560, opened 2022-12-12
Add complex number support to `linalg.outer`

kgryte:
This PR

-   adds complex number support to `linalg.outer`. No special considerations are required to accommodate complex numbers.
-   updates guidance to **require** a numeric data type. If left as "should", would imply that non-numeric data types would be permitted if a library had overriding reasons for doing so. However, for computing the outer product, non-numeric data types are not applicable.


rgommers (approved):
This is as straightforward as it can get - in it goes

PR #561, opened 2022-12-12
Add complex number support to `linalg.svd`

kgryte:
This PR

-   adds complex number support to `linalg.svd`.
-   requires that the output arrays `U` and `Vh` must have the same data type as the input array and that `S` must be unconditionally real.
-   makes explicit that, for matrix stacks, the SVD must be computed for each matrix in the stack.

### DIFF
--- a/spec/API_specification/array_api/linalg.py
+++ b/spec/API_specification/array_api/linalg.py
@@ -393,28 +393,51 @@ def solve(x1: array, x2: array, /) -> array:
     """
 
 def svd(x: array, /, *, full_matrices: bool = True) -> Union[array, Tuple[array, ...]]:
-    """
-    Returns a singular value decomposition A = USVh of a matrix (or a stack of matrices) ``x``, where ``U`` is a matrix (or a stack of matrices) with orthonormal columns, ``S`` is a vector of non-negative numbers (or stack of vectors), and ``Vh`` is a matrix (or a stack of matrices) with orthonormal rows.
+    r"""
+    Returns a singular value decomposition (SVD) of a matrix (or a stack of matrices) ``x``.
+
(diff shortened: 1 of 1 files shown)


spec/API_specification/array_api/linalg.py
```diff
@@ -393,28 +393,51 @@ def solve(x1: array, x2: array, /) -> array:
...
     x: array
-        input array having shape ``(..., M, N)`` and whose innermost two dimensions form matrices on which to perform singular value decomposition. Should have a real-valued floating-point data type.
+        input array having shape ``(..., M, N)`` and whose innermost two dimensions form matrices on which to perform singular value decomposition. Should have a floating-point data type.
     full_matrices: bool
         If ``True``, compute full-sized ``U`` and ``Vh``, such that ``U`` has shape ``(..., M, M)`` and ``Vh`` has shape ``(..., N, N)``. If ``False``, compute on the leading ``K`` singular vectors, such that ``U`` has shape ``(..., M, K)`` and ``Vh`` has shape ``(..., K, N)`` and where ``K = min(M, N)``. Default: ``True``.
 
     Returns
     -------
-    ..
-      NOTE: once complex numbers are supported, each square matrix must be Hermitian.
```
rgommers, line 409: This code comment got removed. I believe that is correct - the comment is wrong. My guess is that it was added when looking at the `hermitian` boolean keyword for [numpy.linalg.svd](https://numpy.org/devdocs/reference/generated/numpy.linalg.svd.html#numpy.linalg.svd)

rgommers (approved):
Additions LGTM, nice to have a bit more detail on what the function does. Thanks @kgryte

PR #562, opened 2022-12-12
Add complex number support to `linalg.svdvals`

kgryte:
This PR

-   adds complex number support to `linalg.svdvals`.
-   requires that the output array have a real-valued floating-point data type having the same precision as the input array.
-   makes explicit that, for matrix stacks, singular values must be computed for each matrix in the stack.


rgommers (approved):
LGTM, in it goes. The interesting bits are in the `svd` specification. Thanks @kgryte

PR #563, opened 2022-12-12
Add complex number support to `linalg.matrix_rank`

kgryte:
This PR

-   adds complex number support to `linalg.matrix_rank`.
-   makes explicit that, for matrix stacks, the number of non-zero singular values must be computed for each matrix in the stack.


rgommers (approved):
LGTM, thanks @kgryte

PR #564, opened 2022-12-12
Add complex number support to `linalg.pinv`

kgryte:
This PR

-   adds complex number support to `linalg.pinv`.
-   makes explicit that, for matrix stacks, the pseudo-inverse must be computed for each matrix in the stack.
-   adds an extended description helping contextualize the use `rtol`.
-   (in practice) depends on <https://github.com/data-apis/array-api/pull/561>.


rgommers (approved):
This LGTM, thanks @kgryte

PR #565, opened 2022-12-12
Add complex number support to `linalg.matrix_norm`

kgryte:
This PR

-   adds complex number support to `linalg.matrix_norm`. The main substantive change is updating the input and output data types. For both real- and complex-valued inputs, matrix norms involve computing absolute values or singular values, so no changes are necessary in norm definitions to accommodate complex numbers.
-   requires that the output data type always be a real-valued floating-point data type.


rgommers (approved):
LGTM, thanks @kgryte

PR #566, opened 2022-12-12
Add complex number support to `linalg.solve`

kgryte:
This PR

-   adds complex number support to `linalg.solve`.
-   makes explicit that, for matrix stacks, a solution must be computed for each matrix in the stack.


rgommers (approved):
Added math explanations and their rendered version LGTM. In it goes, thanks @kgryte

PR #567, opened 2022-12-13
Add complex number support to `linalg.slogdet`

kgryte:
This PR

-   adds complex number support to `linalg.slogdet`.
-   defines the sign of the determinant in agreement with <https://github.com/data-apis/array-api/pull/556>. In contrast to its `sign` function, NumPy uses the definition $\operatorname{sign}(\det x) = \frac{\det x}{|\det x|}$ to compute the sign of the determinant. At the time of this PR, [PyTorch](https://github.com/pytorch/pytorch/pull/56265/files#diff-316ce439a56491298e2d98deeca82606c52e5bde2f1ceb16c534ec03386c817eR224) refers to the "sign" as the angle (which I take to mean the argument of the determinant); however, I don't believe this is correct, as, via polar decomposition, $\det A = |\det A| e^{\theta j}$. Given the expected equality that $\det A = \operatorname{sign}(\det A) \cdot e^{\ln |\det A|} =  |\det A| \operatorname{sign}(\det A)$, we see that what should be returned is $e^{\theta j}$, not simply the argument $\theta$. While the PyTorch docs state that the angle is returned, what is actually returned is $\frac{\det x}{|\det x|}$.
-   requires that the `sign` array be the same data type as the input array and `logabsdet` be unconditionally real-valued.
-   makes explicit that, for matrix stacks, a solution must be computed for each matrix in the stack.

### DIFF
--- a/spec/API_specification/array_api/linalg.py
+++ b/spec/API_specification/array_api/linalg.py
@@ -361,6 +361,8 @@ def slogdet(x: array, /) -> Tuple[array, array]:
 
     where :math:`|\det x|` is the absolute value of :math:`\det x`.
 
+    When ``x`` is a stack of matrices, the function must compute the sign and natural logarithm of the absolute value of the determinant for each matrix in the stack.
+
     **Special Cases**
 
     For real-valued floating-point operands,
--- a/spec/API_specification/array_api/linalg.py
+++ b/spec/API_specification/array_api/linalg.py
@@ -367,11 +367,11 @@ def slogdet(x: array, /) -> Tuple[array, array]:
 
     For real-valued floating-point operands,
 
-    - If the determinant is zero, the corresponding ``sign`` should be ``0`` and ``logabsdet`` should be ``-infinity``.
+    - If the determinant is zero, the ``sign`` should be ``0`` and ``logabsdet`` should be ``-infinity``.
 
     For complex floating-point operands,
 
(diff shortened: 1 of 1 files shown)


rgommers:
> ...  we see that what should be returned is $e^{\theta j}$, not simply the argument $\theta$. While the PyTorch docs state that the angle is returned, what is actually returned is $\frac{\det x}{|\det x|}$.

I think that that's mostly a terminology thing? The implementations are consistent:

```python
>>> x = np.array([[0.1+0.9j, 0.5-0.5j], [1, 2j]])
>>> np.linalg.slogdet(x)
((-0.9566738804288585+0.29116161578269606j), 0.877201841342143)
>>> torch.linalg.slogdet(torch.as_tensor(x))
torch.return_types.linalg_slogdet(
sign=tensor(-0.9567+0.2912j, dtype=torch.complex128),
logabsdet=tensor(0.8772, dtype=torch.float64))
```

If there's something to clarify or improve in the PyTorch wording in the `slogdet` description, that's of course useful. @lezcano since you wrote those docs, you may want to have a look at the comment in the PR description here.

rgommers (approved):
Changes here LGTM, thanks @kgryte

lezcano:
I agree that the term angle is not correct. The quantity that's returned in PyTorch is indeed the sign, and the rest of the docs and the implementation are consistent with this. I'll fix that in the PyTorch docs soon.

PR #568, opened 2022-12-13
Add complex number support to `linspace`

kgryte:
This PR

- adds complex number support to `linspace`.
- allows `start` and `stop` to be `complex` numbers.
- allows `dtype` to a complex floating-point data type, not just a real-valued floating-point data type.
- specifies that spacing between complex numbers be computed by independently computing the spacing for real and imaginary components, respectively. This follows existing practice in NumPy, PyTorch, Julia, MATLAB, and elsewhere.
- specifies that if `dtype` is `None`, if either `start` or `stop` is a complex number, then the output dtype must be the default complex floating-point dtype; otherwise, the output dtype must be the default real-valued floating-point dtype.
- adds a warning that conversions of `start` and `stop` should follow type promotion rules; conversions not following type promotion rules are implementation-dependent (e.g., NumPy allows specifying a real-valued dtype when `start` and/or `stop` is complex by discarding the imaginary component; as such a conversion is not specified in this standard, such behavior is implementation-defined).
- removes a requirement that `linspace` raise an exception if `num` is a not a nonnegative integer, as the standard typically avoids specifying validation behavior, leaving it up to conforming libraries to determine how to best handle invalid inputs (e.g., raise an exception, clip inputs, etc).

### DIFF
--- a/spec/API_specification/array_api/creation_functions.py
+++ b/spec/API_specification/array_api/creation_functions.py
@@ -213,24 +213,45 @@ def full_like(x: array, /, fill_value: Union[bool, int, float, complex], *, dtyp
         an array having the same shape as ``x`` and where every element is equal to ``fill_value``.
     """
 
-def linspace(start: Union[int, float], stop: Union[int, float], /, num: int, *, dtype: Optional[dtype] = None, device: Optional[device] = None, endpoint: bool = True) -> array:
-    """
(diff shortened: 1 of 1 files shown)


rgommers (approved):
This all looks correct, in it goes. Thanks @kgryte

PR #572, opened 2022-12-14
Update dtype guidance for `eigh` and `eigvalsh`

kgryte:
This PR

- updates the data type guidance for `eigh` and `eigvalsh` to be aligned with other `linalg` APIs. Currently, the specification says the input dtypes for these two APIs "must" be floating-point; however, all other similar `linalg` APIs state that the input dtypes "should" be floating-point. This PR corrects this discrepancy.


rgommers (approved):
LGTM, thanks @kgryte

Issue #539, opened 2022-12-01
Disallow promoting real-valued inputs to complex dtypes when values are outside of their real domain


kgryte:
Initially discussed in <https://github.com/data-apis/array-api/issues/514#issuecomment-1333421314>.

In summary, when a user provides a real-valued input to a function such as `log` or `sqrt` and one or more values are outside of their real domain, implementations should not automatically promote to a complex floating-point dtype.

The specification does not support value based promotion, but we should reiterate this with particular regard to real-complex. Responsibility for type promotion should be punted to userland.

Issue #533, opened 2022-12-01
Tracking issue for specification maintenance tasks


kgryte:
The purpose of this issue is to identify and track various maintenance tasks which have accumulated concerning specification organization and development.

- [ ] Move special cases and other explanatory content to a dedicated "notes" section: <https://github.com/data-apis/array-api/issues/518>
- [ ] Setup EditorConfig to ensure consistent line endings and spacing.
- [ ] Use black for automated code formatting and line wrapping (possibly using Git hooks).
- [ ] Setup linting of Python files.
- [ ] Move universal preamble content (e.g., positional parameters, optional parameters, etc) to a dedicated design document.
- [ ] Migrate special cases to use math directives: <https://github.com/data-apis/array-api/issues/519>

rgommers:
Thanks for summarizing this @kgryte. A few thoughts:
- When doing this, let's try to ensure that large content refactors do not make it harder to use `git blame`, by using https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view
- Some of this may be a little time-consuming. To me the higher-prio ones, good to get into the v2022 spect, are "move special cases", "move universal preamble", and "use `black`"
- After that, the "use math directives one", because it's also still a big diff, so makes later maintenance easier if it's done before branching
- Linting in CI and EditorConfig are lower prio, they can be fiddly and don't impact the 2022 content or organization of it

rgommers:
The v2022 tasks are done. I reorganized the description so it's clear what is left, and removed the milestone.

kgryte:
As commented in https://github.com/data-apis/array-api/pull/689#discussion_r1449622314, we should investigate whether we can minimize type annotation duplication. Currently, we copy-paste the type annotation from the signature into the docstring. This increases the risk of drift, and preferably, we'd only need to write these annotations once.

rgommers:
Sphinx has functionality for that ([`autodoc_typehints`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_typehints)), but like anything `autodoc` related it had some issues, especially with dealing with type aliases. Not sure if that has been improved recently, but I did try this when initially setting up the docs.

rgommers:
If I had to deal with this now, I'd dump Sphinx complete in favor of MkDocs / mkdocs-material.

kgryte:
> Sphinx has functionality for that...I'd dump Sphinx completely...

[...]

PR #573, opened 2022-12-14
Consolidate complex number design topics and add guidance on value-based promotion

kgryte:
This PR

- consolidates the three complex number design documents ("complex numbers", "complex number ordering", and "branch cuts") into a single design document.
- add guidance on value-based promotion when the results of operations on real-valued input arrays are outside of their real domain. This closes <https://github.com/data-apis/array-api/issues/539>.
- addresses one of the tasks described in <https://github.com/data-apis/array-api/issues/533>.


rgommers (approved):
This LGTM, thanks @kgryte. The one added section on avoiding value-based promotion is clear, and matches the rest of the spec on type promotion.

Issue #518, opened 2022-11-27
Reorganize documentation to display parameters and return values more prominently


kgryte:
As discussed in a [comment](https://github.com/data-apis/array-api/pull/461#pullrequestreview-1188833926), we should reorganize docstring content such that parameters and return values are documented immediately after the interface definition and initial description and such that explanatory notes, including special cases, are documented in a separate section immediately following documentation of return values and/or raised exceptions.

This reorganization will ensure that parameters and return values are more prominently displayed and are closer in proximity to the API definition. Currently, for those APIs with many floating-point special cases, a definition and its associated parameter descriptions can be spaced far apart, thus requiring a user to scroll to view desired content and preventing the user from viewing the definition and associated descriptions within the same view at the same time.

PR #578, opened 2022-12-19
Move special cases to notes sections

kgryte:
This PR

- resolves https://github.com/data-apis/array-api/issues/518 (and as discussed in https://github.com/data-apis/array-api/issues/533) whereby special cases should be moved to "Notes" sections.
- only moves notes directives which explicitly discuss special case behavior; otherwise, other notes concerning other aspects of API behavior were left in place (e.g., when complex conjugation should result in the same output whether conjugating the input or the output, etc).
- removes special case documentation in the `array_object` which is documented in existing function API documentation. This addresses duplication and maintainability concerns.
- adds two commits to the `.git-blame-ignore-revs` file in order to retain Git blame for documented special cases.
- a follow-up PR based on the changes introduced in this PR will be submitted to add C99 branch cut notes.

## Notes

- Only 2 of the 3 substantive commits were added to the `.git-blame-ignore-revs` file. The omitted commit should not be included in the list of ignored commits as the changes are not simply reorganization and should be tracked via Git blame.
- This PR should be merged with a merge commit, rather than squash-merged, in order for the commit hashes in the `.git-blame-ignore-revs` file to be correct.

### DIFF
--- a/spec/API_specification/array_api/elementwise_functions.py
+++ b/spec/API_specification/array_api/elementwise_functions.py
@@ -16,6 +16,25 @@ def abs(x: array, /) -> array:
        .. math::
           \operatorname{abs}(z) = \sqrt{a^2 + b^2}
 
+    .. note::
+       For complex floating-point operands, conforming implementations should take care to avoid undue overflow or underflow during intermediate stages of computation.
+
+    ..
+       TODO: once ``hypot`` is added to the specification, remove the special cases for complex floating-point operands and the note concerning guarding against undue overflow/underflow, and state that special cases must be handled as if implemented as ``hypot(real(x), imag(x))``.
+
+    Parameters
+    ----------
+    x: array
+        input array. Should have a numeric data type.
+
+    Returns
+    -------
+    out: array
+        an array containing the absolute value of each element in ``x``. If ``x`` has a real-valued data type, the returned array must have the same data type as ``x``. If ``x`` has a complex floating-point data type, the returned arrayed must have a real-valued floating-point data type whose precision matches the precision of ``x`` (e.g., if ``x`` is ``complex128``, then the returned array must have a ``float64`` data type).
+
+    Notes
+    -----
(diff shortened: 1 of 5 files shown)


spec/API_specification/array_api/elementwise_functions.py
```diff
@@ -398,6 +404,9 @@ def atan(x: array, /) -> array:
     -------
     out: array
         an array containing the inverse tangent of each element in ``x``. The returned array must have a floating-point data type determined by :ref:`type-promotion`.
+
+    Notes
+    -----
```
rgommers, line 409: The special cases got lost here somehow

spec/API_specification/array_api/elementwise_functions.py
```diff
@@ -1014,25 +1062,66 @@ def floor(x: array, /) -> array:
     - If ``x_i`` is ``+0``, the result is ``+0``.
     - If ``x_i`` is ``-0``, the result is ``-0``.
     - If ``x_i`` is ``NaN``, the result is ``NaN``.
+    """
+
+
+def floor_divide(x1: array, x2: array, /) -> array:
```
rgommers, line 1068: `floor_divide` ended up with duplicate content

rgommers (changes requested):
Thanks @kgryte. I went through the rendered version of everything (I think), and found two issues. There's more to do on this topic, but let's leave that until a later time (after tagging `v2022.12`).

spec/API_specification/array_api/elementwise_functions.py
kgryte: Thanks for catching. Fixed.

kgryte:
@rgommers Thanks for reviewing and catching `atan` and `floor_divide`. Those issues should now be addressed.

rgommers (approved):
LGTM now, in it goes.

Issue #584, opened 2023-01-14
Run-time version of API


nstarman:
Hi, first time contributor, long-time fan. I've been following the development of the Array API and having discussions with my fellow Astropy maintainers about how Astropy's Quantity class can support the API and how this might impact our users. Overall we feel that the adoption of the API will be beneficial (preaching to the choir here) but we, or at least I, have a point of concern.

Right now NumPy serves as the primary source of mathematical functions: ``np.cos``, etc. Through their interoperability protocols, like ``__array_function__`` we can vendor custom array classes like Quantity and for users most of numpy *just works*. This API standard has a somewhat different paradigm: where ``astropy.units`` will need to have a Quantity namespace that defines ``cos`, etc. In some ways this is good: for Astropy, Quantity can more easily support non-numpy array objects by acting as a wrapper and forwarding actual computations to the wrapped array object. For example,

```python
def cos(x: Quantity, /) -> Quantity:
    unit = ...
    return Quantity(x.__array_namespace__().cos(x.value), unit)
```

However, we worry the loss of a central dispatching library is a regression in ease of user experience. Yes numpy will continue to ``__array_(u)func(tion)__`` forever/ a long time and so can still serve as a central dispatching library, but base numpy does *not* conform to the array API standard. So there will not be an array-API-conformant central dispatching library and users will need to uses non-conformant libraries, which seems at odds to the purpose of this standard.

The situation becomes more difficult for Astropy and users when we consider Astropy's number of other array-like objects  for which we might want to implement the array API: [table Columns](https://docs.astropy.org/en/stable/api/astropy.table.Column.html), [coordinate representations](https://docs.astropy.org/en/stable/api/astropy.coordinates.BaseRepresentationOrDifferential.html#astropy.coordinates.BaseRepresentationOrDifferential), [Distributions](https://docs.astropy.org/en/stable/api/astropy.uncertainty.Distribution.html). Users will need to import a new array namespace for each — e.g. ``from astropy.table import column_api as cp``.

I'm sure this issue has come up in discussion before. To add to that discussion, I'd like to propose what came up in ours. We felt the best solution might be if array-api published a package, not offering a new Array implementation which I know is out of scope, just of the supported functions. These functions would be thin wrappers, dispatching the actual computations to the array namespace of the argument. For example:

```python
def cos(x: ArrayAPIConformant, /) -> ArrayAPIConformant:
    return get_namespace(x).cos(x)
```

Now the following is possible:

```python
>>> import array_api as ap
[...]

PR #589, opened 2023-02-05
Array Protocol

nstarman:
Continuing from https://github.com/data-apis/array-api/issues/584, I've done a quick refactor of the array (and dtype) objects to make them typing Protocols.
@rgommers, I agree that making Array generic wrt the dtype is good for a followup.


src/array_api_stubs/_draft/_types.py
```diff
@@ -60,6 +68,949 @@ def __len__(self, /) -> int:
...
+        x: array,
+        /,
+        *,
+        axis: int | tuple[int, ...] | None = None,
+        keepdims: bool = False,
+    ) -> array:
+        ...
+
+
+class ArrayAPILinAlgNamespace(Protocol[array]):
```
honno, line 878: I guess making these namespace protocols generic allows downstream users to subscript their own type hints for array class. Excuse my ignorance here—if one were not to subscript the type hint, would say `mypy` type check with our array `Protocol`?

src/array_api_stubs/_draft/array_object.py
```diff
@@ -52,9 +58,10 @@ def mT(self: array) -> array:
         out: array
             array whose last two dimensions (axes) are permuted in reverse order relative to original array (i.e., for an array instance having shape ``(..., M, N)``, the returned array must have shape ``(..., N, M)``). The returned array must have the same data type as the original array.
         """
+        ...
 
     @property
-    def ndim(self: array) -> int:
+    def ndim(self) -> int:
```
honno, line 64: So I see methods/properties which don't return arrays don't get any type hint for their `self` arg—could these be hinted with `Array`?

honno (commented):
Thanks for your work on this :)

The `Array` protocl stuff looks grand. And personally I like having a `DType` class just for consistency, even if practically type hinting an object has `__eq__` isn't useful.

On namespace type hints (e.g. `_CreatoinFuncs`, `ArrayAPINamespace`): _IMO_ I'm not sure on this yet, as they make contributing to the spec less ergonomic given we repeat the signatures, although they are pretty nifty to have. <sup>Is there a world where these could be dynamically construct them from the stubs instead?</sup> We might want to put these in a follow-up PR if just to ship the concretely non-controversial stuff first—interested in what others think.

nstarman:
> We might want to put these in a follow-up PR if just to ship the concretely non-controversial stuff first—interested in what others think.

I was thinking the same thing last evening about moving this to a followup PR.

> On namespace type hints (e.g. `_CreatoinFuncs`, `ArrayAPINamespace`): _IMO_ I'm not sure on this yet, as they make contributing to the spec less ergonomic given we repeat the signatures, although they are pretty nifty to have. Is there a world where these could be dynamically construct them from the stubs instead? 

Maybe? I haven't experimented yet but perhaps mypy will be satisfied with:

```python
class ArrayAPINamespace(Protocol):

    zeros_like = staticmethod(creation_funcs.zeros_like)
```

src/array_api_stubs/_draft/_types.py
nstarman: The advantage of making ``ArrayAPINamespace`` (and related) generic wrt the array type is for 

```python
class Array:
    def __array_namespace__() -> ArrayAPINamespace[Self]: ...
```

Now a static type checker knows that for an array-api implementing library *X*, the namespace returned by ``<X>.__array_namespace__`` work on *X* arrays.

src/array_api_stubs/_draft/array_object.py
nstarman: Static type checkers don't like this because the ``TypeVar`` is not properly bound — see https://peps.python.org/pep-0484/#scoping-rules-for-type-variables ("Unbound type variables should not appear in the bodies of generic functions, or in the class bodies apart from method definitions:").
In general, there's no need to type hint the self arg unless the return type depends on the type of self — e.g. see https://peps.python.org/pep-0484/#user-defined-generic-types.

nstarman: A nice thing about using Protocol is that ``self`` is understood by static type checkers to be the ``Protocol``.

rgommers (commented):
Thanks @nstarman, and apologies for the delay in reviewing (your PR came in right at the start of my 2 week holiday). This is looking quite good. I pushed a couple of small commits to resolve some Mypy and Sphinx complaints.

This is close to ready I'd say. One thing that isn't nice is that `Self` now shows up instead of `array` in the signatures of the html docs. I tried to fiddle with `autodoc_type_aliases` in `src/_array_api_conf.py`, but was not successful in mapping that back to `array`. Using "self" will be pretty confusing in docs for regular methods that return arrays. I think we'll have to figure that one out.

The `DType` protocol is fine with me - it won't add much to type-checking, but it should be fine to have it here for consistency.

The `Array` protocol looks good.

rgommers:
@BvB93 if you have any feedback on this PR, that would be great to see.

rgommers:
@nstarman there are 3 errors left when running `mypy array_api_stubs/_draft/array_object.py` (from lots more on `main`). There are more errors in other files - but if we are going to maintain static typing support in this repo, adding a CI job running Mypy seems feasible. We could start with this file, and expand it to all files later. WDYT?

nstarman:
> One thing that isn't nice is that `Self` now shows up instead of `array` in the signatures of the html docs.

If the `array` type is just text in the html docs and not 'clickable', then we can rename `Self` to `array`.

> adding a CI job running Mypy seems feasible. We could start with this file, and expand it to all files later. WDYT?

That sounds like a good idea. I can try adding to the CI and create an ignore file that can later be whittled down as the package is made more type friendly.

rgommers:
> If the `array` type is just text in the html docs and not 'clickable', then we can rename `Self` to `array`.

That works. It's not clickable - there is no array object in the standard on purpose, because in the end we don't care whether it's named `ndarray`, `Tensor`, or something else.

src/array_api_stubs/_draft/array_object.py
```diff
@@ -145,8 +154,9 @@ def __abs__(self: array, /) -> array:
         .. versionchanged:: 2022.12
             Added complex data type support.
         """
+        ...
 
-    def __add__(self: array, other: Union[int, float, array], /) -> array:
+    def __add__(self: Self, other: int | float | Self, /) -> Self:
```
kgryte, line 159: What the rationale for the avoidance of `Union` and `Optional` in this PR?

nstarman: Neither are necessary with Python's newer language features. I think they're essentially considered legacy, in these contexts, like `typing.Tuple` is obviated by just `tuple`.

kgryte: Given that some of these features are Python 3.10, I suppose we may need to be mindful of backward compat and specifying the minimum supported Python version.

rgommers: I verified that that's correct - this is Python 3.8 compatible. Resolving.

nstarman:
Sorry for the long interlude. I can't seem to resolve 3 linking errors in the docs  https://app.circleci.com/pipelines/github/data-apis/array-api/1088/workflows/ed2dfa75-c3c7-4845-8f28-e01fe35d0c8d/jobs/996?invite=true#step-103-1646

asmeurer:
Sphinx is trying to create links for the type hints. I think to fix that you need to update the list(s) here https://github.com/data-apis/array-api/blob/main/src/_array_api_conf.py#L52-L77

nstarman:
I can squash commits if you don't use Squash & Merge (assuming this is approved, of course :) )

src/array_api_stubs/_draft/array_object.py
```diff
@@ -63,9 +69,10 @@ def ndim(self: array) -> int:
         out: int
             number of array dimensions (axes).
         """
+        ...
 
     @property
-    def shape(self: array) -> Tuple[Optional[int], ...]:
+    def shape(self) -> tuple[int | None, ...]:
```
honno, line 75: Just to say more for myself than anyone else—I initially wasn't too sure if we wanted to use the py3.9+ type hint features, but:
* Forgot that `from future import __annotations__` makes this work for py3.8
* py3.8 doesn't get updates end of next year
* The scientific ecosystem already is adopting py3.9+

src/array_api_stubs/_draft/array_object.py
```diff
@@ -359,7 +376,7 @@ def __dlpack_device__(self: array, /) -> Tuple[Enum, int]:
 
         Returns
         -------
-        device: Tuple[Enum, int]
+        device: Tuple[:class:`~enum.Enum`, int]
```
honno, line 379: I assume this is an artifact from your code editor and can be reverted?

honno (commented):
Nice work!

src/array_api_stubs/_draft/array_object.py
nstarman: I'll revert this and hopefully the docs still compile correctly. This was just a sphinx thing I tried.
I'm also planning to squash my commits, unless y'all use Squash and Merge.

nstarman: Done!

honno (approved):
Looks good on my end!

I think `pre-commit` is a good addition (would love to add hooks for other spec things myself), but probably not too useful until we get a workflow that runs it. That said, I'm not too sure if we'd want to do that just yet :thinking:

nstarman:
> I think `pre-commit` is a good addition (would love to add hooks for other spec things myself), but probably not too useful until we get a workflow that runs it. That said, I'm not too sure if we'd want to do that just yet 🤔

`pre-commit.ci` can be turned on, sans additional GH workflows.

nstarman:
> Looks good on my end!

Thanks @honno for the reviews and discussion.

src/array_api_stubs/_draft/array_object.py
asmeurer: Why is this only done for certain methods?

nstarman: I did it for any methods I looked at over the course of the PR. This is by no means comprehensive and should be done: here or elsewhere.

asmeurer:
Any reason to change `array` to `Array`? I think we might prefer to keep it as `array` like it is now.

nstarman:
The class would have to be called `array`, which is problematic re PEP-8 and also conflicts name wise with the array `TypeVar`. The latter can be changed, but for the former isn't it desirable to refer to `array` instances of some class `Array`?

asmeurer:
The class name is specifically not part of the standard. I always felt that `array` made that a little clearer than `Array`, but maybe that's just me. 

By the way, the protocol isn't going to require the class to be called `Array` is it? If it does, that's a problem.

nstarman:
> By the way, the protocol isn't going to require the class to be called Array is it? If it does, that's a problem.

Not specifically. The protocol is a class and may be named anything. If it's named the same things as another variable, e.g. the `array = TypeVar("array", bound=Array)` then that can be problematic.

As a more zoomed out view. The goal of this and subsequent PRs is to enable the spec to also be a functioning & installable package that can be used for run-time checks of adherence to this spec.

```python
>>> from data_apis.array import Array  # this spec
>>> from numpy import array, ndarray  # post v2.0

>>> issubclass(ndarray, Array)
True

>>> isinstance(array(1), Array)
True
```

To be clear, `Array` cannot be instantiated since it is a protocol without any implementation. Its use is entirely for structural subtyping (e.g. type hinting) and run-time checks.
At some level it's important that the spec adhere to Python standards, e.g. naming conventions. 

> The class name is specifically not part of the standard. 

I'm not sure I understand then what's the problem with calling it `Array`?

src/array_api_stubs/_draft/array_object.py
```diff
@@ -1,25 +1,28 @@
...
+
+
+class Array(Protocol):
+    def __init__(self) -> None:
         """Initialize the attributes for the array object class."""
+        ...
 
     @property
-    def dtype(self: array) -> Dtype:
+    def dtype(self) -> "DType":
```
BvB93, line 25: FYI you won't need any quotation marks if `from __future__ import annotations` is enabled as this (implicitly) stringifies all annotations anyway.

src/array_api_stubs/_draft/array_object.py
```diff
@@ -28,9 +31,10 @@ def dtype(self: array) -> Dtype:
         out: dtype
             array data type.
         """
+        ...
 
     @property
-    def device(self: array) -> Device:
+    def device(self) -> "Device":  # type: ignore[type-var]
```
BvB93, line 37:
```suggestion
    def device(self) -> Any:
```

Type variables (such as `Device` here) don't really make sense when it's only used as return type without using them in either a generic class or another parameter within the function; this is the source of mypy's error here and why you need the `type: ignore`; there must _always_ be at least two of them.

Instead, unless you're willing to make the protocol itself generic w.r.t. the device (which might be a bit excessive here), then it's customary to use `Any`. I'm not sure if this `Device` type is used anywhere else, but if not you could even consider just repurposing it as an `Any` alias in order to keep its more descriptive name.

src/array_api_stubs/_draft/array_object.py
```diff
@@ -39,6 +43,7 @@ def device(self: array) -> Device:
         out: device
             a ``device`` object (see :ref:`device-support`).
         """
+        ...
 
     @property
     def mT(self: array) -> array:
```
BvB93, line 49: As a heads up: the [PEP 673](https://peps.python.org/pep-0673/) `Self` type (a special type variable that's always bound to the respective classes `self` parameter) is now also supported by mypy and the likes, so you could, if so desired, express this as:
```python
if TYPE_CHECKING:
    from typing_extensions import Self
    
 class Array(Protocol):
    def mT(self) -> Self:
```

src/array_api_stubs/_draft/linalg.py
```diff
@@ -272,7 +272,7 @@ def matrix_norm(
     /,
     *,
     keepdims: bool = False,
-    ord: Optional[Union[int, float, Literal[inf, -inf, "fro", "nuc"]]] = "fro",
+    ord: Optional[Union[int, float, Literal[inf, -inf, "fro", "nuc"]]] = "fro",  # type: ignore
```
BvB93, line 275: Note that there's currently no way to express literal floats in the type system, hence the need for the `# type: ignore` comment. Could maybe consider adding two `Inf` and `NInf` types (or something along those lines) and declare them as `float` aliases?
``` python
Inf = float
NInf = float
```

src/array_api_stubs/_draft/array_object.py
```diff
@@ -145,8 +155,9 @@ def __abs__(self: array, /) -> array:
         .. versionchanged:: 2022.12
             Added complex data type support.
         """
+        ...
 
-    def __add__(self: array, other: Union[int, float, array], /) -> array:
+    def __add__(self: array, other: int | float | array, /) -> array:
```
BvB93, line 160: Just checking: if `other` is an array, is it always intended for it to be of the same type as `self`? No `other` subclasses? No different `self` subclass? If no, then it'd be best to annotate it without use of the type variable:
```suggestion
    def __add__(self: array, other: int | float | Array, /) -> array:
```

src/array_api_stubs/_draft/array_object.py
```diff
@@ -1031,9 +1073,10 @@ def __xor__(self: array, other: Union[int, bool, array], /) -> array:
         .. note::
            Element-wise results must equal the results returned by the equivalent element-wise function :func:`~array_api.bitwise_xor`.
         """
+        ...
 
     def to_device(
-        self: array, device: Device, /, *, stream: Optional[Union[int, Any]] = None
+        self: array, device: "Device", /, *, stream: int | Any | None = None  # type: ignore[type-var]
```
BvB93, line 1079: There's been some minor discussion about this before (xref https://github.com/numpy/numpy/pull/19083#discussion_r638069860), but considering this class is now becoming a fully fledged protocol and not just a documentation device the `int` type in the `stream` annotation is starting to become a bit problematic as it makes it mandatory for any (concrete) subtype to support it (while this is, in fact, optional).

src/array_api_stubs/_draft/array_object.py
nstarman: I have a note already
```
array = TypeVar("array", bound="Array")
# NOTE: when working with py3.11+ this can be ``typing.Self``.
```

But I'm happy to change it now. I was avoiding adding additional dependencies, even if they aren't runtime and 
are made by the core python devs.

src/array_api_stubs/_draft/array_object.py
nstarman: This occurred while trying to get the docs to render. Many things were tested, so I don't know if this is strictly necessary (I'll check), but it was not related to type hint stringification.

src/array_api_stubs/_draft/linalg.py
nstarman: I agree it's annoying Python can't do `Literal[<float>]`. I considered doing this, but I think I prefer not to change the spec in this PR. Changing to alias values will require additional docs and explanation since `Inf = float` is not the same as `math.inf`, and the latter is part of the spec.

src/array_api_stubs/_draft/array_object.py
nstarman: Agreed 👍

src/array_api_stubs/_draft/array_object.py
nstarman: A good point. I'm trying to translate this class, not change the spec, so I think this is best done in another PR.

src/array_api_stubs/_draft/array_object.py
nstarman: > unless you're willing to make the protocol itself generic w.r.t. the device (which might be a bit excessive here),

This is planned, but for a followup PR. (See opening comment to this PR).

>  then it's customary to use Any

Alternatively, this PR introduces a Device Protocol which can be used instead of the `TypeVar`. No binding issues.

BvB93: > Alternatively, this PR introduces a Device Protocol which can be used instead of the TypeVar. No binding issues because it's a Protocol. When Array is made generic wrt Device we can change this back to a TypeVar.

I guess that would be sort of possible? Rereading the [Device Support](https://data-apis.org/array-api/latest/design_topics/device_support.html#syntax-for-device-assignment) docs concrete API implementations have _a lot_ of freedom regarding the implementation of their Device type, to the point where only `__eq__` (and `__ne__`) really matter? Honestly, with an interface that minimal it might just be worthwhile to stick to plain `object` rather than `Any` or a protocol, as there's very little (or nothing) to gain by usage of the latter two.

src/array_api_stubs/_draft/linalg.py
BvB93: >  Changing to alias values will require additional docs and explanation since Inf = float is not the same as math.inf, and the latter is part of the spec.

True, though the `float` already being present as a valid `ord` parameter means that the inclusion of `math.inf` is redundant either way (from a typing perspective at least), so to me this doesn't sound like such a change would violate the spec.

src/array_api_stubs/_draft/array_object.py
BvB93: IMO this would not be a spec violation as it'd merely remove an inconsistency between the type signature as implied by the annotations `(..., stream: int | Any | None = ...)` versus the one described in docstring (`..., stream: None = ...`).

As a little bit more background on the nitty gritty part of the world of typing that's going on here: Contrary to the return type of callables (which is covariant and concrete protocol implementations can thus return any sub-type), argument types are contravariant, meaning that concrete implementations will either have to exactly match it _or_ return a supertype thereof (in the most extreme scenario that'd mean that even plain `object` is acceptable, though whether or not that's an actual good practice is another discussion entirely...).

What this means is that in order to proper subtype something like `def (stream: int | None = ...) -> object: ...` then concrete implementation _must_ support both `int` and `None`, something that contracts the actual docstring.

src/array_api_stubs/_draft/array_object.py
nstarman: Yes, currently a `Device` Protocol has only ``__eq__`` in it and as such is effectively equivalent to `object`.
But it's designing with the future in mind ("The choice to not include a standardized Device object may be revisited in a future revision of this standard.") and we'd need to alias `object` anyway to get the correct rendering on the docs...

src/array_api_stubs/_draft/array_object.py
nstarman: Done.

src/array_api_stubs/_draft/linalg.py
nstarman: ✅

nstarman: I added the type ignores only to make my local mypy happier (mypy isn't part of the the CI on this repo). 
I think making these changes should be done in a followup PR. I'm happy to make the changes in another PR.

src/array_api_stubs/_draft/array_object.py
nstarman: Then let's tackle this in a followup, since `Self` will need to be rendered in the docs as `Array`.

src/array_api_stubs/_draft/array_object.py
nstarman: Done!

BvB93:
LGTM, though it might be good to verify if the currently available downstream implementations properly satisfy subtype the protocol as this could potentially reveal some typing-related corner cases that we've missed thus far. Could be something for a follow up though.

src/array_api_stubs/_draft/array_object.py
nstarman: It's not necessary! Removed.

TomNicholas:
It would be great to see this released. It looks very developed to me, and would be useful for us within xarray.

src/array_api_stubs/_draft/linalg.py
```diff
@@ -793,7 +793,7 @@ def vector_norm(
     *,
     axis: Optional[Union[int, Tuple[int, ...]]] = None,
     keepdims: bool = False,
-    ord: Union[int, float, Literal[inf, -inf]] = 2,
+    ord: Union[int, float, Literal[inf, -inf]] = 2,  # type: ignore
```
jorenham, line 796: Perhaps this is out of scope, but this `Literal` is redundant (and as this change implies, also incorrect), since `inf` and `-inf` are both assignable to `float`.

nstarman:
@rgommers. I'd be happy to finish / merge this in. I don't recall why this stalled out.

nstarman:
Rebased and applied fixes for things that had changed.

kgryte:
@nstarman Would you be interested in coming to the [community meeting](https://github.com/data-apis/array-api/issues/853) this week (Nov 28) to discuss static typing and next steps? It would be great to have some perspectives from static typing experts!

nstarman:
I would be very interested to come! It is Thanksgiving, but I should be able to join.

NeilGirdhar:
This PR is awesome.  Really looking forward to it being exposed.

Just one question:  Why is the preference for using protocols instead of ordinary inheritance?

It's not like there are Array API clients that are going to be reluctant to inherit from your `Array` base class.

The problem with using protocols is that if you ever change any definition in the protocol, a client array may no longer see itself as compatible, and then a user program will report type errors at every Array API use.

Also, for the clients, they have to be very careful implementing their class to match the protocol or else it won't be compatible.  With inheritance, they will instead get type errors if they make a mistake.

Finally, protocols are slower for type checkers to check than ordinary class inheritance.

TomNicholas:
Presumably so that array implementation packages can avoid an extra dependency.

NeilGirdhar:
> Presumably so that array implementation packages can avoid an extra dependency.

Right, but are the costs worth it?

And what's the benefit of of avoiding dependencies?  Most users are going to depend on the Array API type annotations anyway.

jorenham:
> Just one question: Why is the preference for using protocols instead of ordinary inheritance?

Inheritance hurts performance, and the additional import that this would require also would. Array libraries are usually heavily optimized, so this is not as realistic ask. And I'm not sure whether it would even be possible in case of e.g. `numpy.ndarray`, which is written in C.

rgommers:
It's not performance, it's that a required dependency is simply unacceptable for most libraries. NumPy has zero runtime dependencies, and it's highly unlikely that any will be added over the next few years - and this kind of thing certainly doesn't meet the bar. Other libraries may have similar constraints, plus there's potential issues with requiring different minimum versions of the same dependency.

ntessore:
> The problem with using protocols is that if you ever change any definition in the protocol, a client array may no longer see itself as compatible, and then a user program will report type errors at every Array API use.

Isn't this exactly the same either way?

> Also, for the clients, they have to be very careful implementing their class to match the protocol or else it won't be compatible. With inheritance, they will instead get type errors if they make a mistake.

Again, isn't this exactly the same either way? The only difference being an `assert_type(MyArray, ArrayApiArray)`?

Generally speaking, I think having Array API protocols is valuable: it allows the user to take responsibility when using an implementation that is not (yet) fully compliant, instead of putting it solely in the hands of the implementation to declare that it is compliant.

It also means implementations can gradually transition to becoming Array API compliant, and they get told exactly where they currently aren't. I found this very useful with @nstarman's repo.

NeilGirdhar:
Sorry, I didn't expect to generate so many replies and flood this PR.  I've created an issue to continue the conversation: https://github.com/data-apis/array-api/issues/899

PR #610, opened 2023-03-13
Make it clearer that the dtype kinds in isdtype are the string inputs

asmeurer:


rgommers (approved):
Thanks Aaron, I agree that this looks better and helps shows that these are the string inputs. In it goes.

PR #625, opened 2023-04-21
Allow negative axes in tensordot

asmeurer:
Both NumPy and PyTorch allow this, and there should be no ambiguity or difficulty in doing so, as long as the specified axes remain unique.


kgryte (approved):
LGTM. Thanks, @asmeurer!

kgryte:
As this PR introduces changes which are consistent with guidance elsewhere in the specification (namely, a valid axis index may be negative, resolving to a positive index according to `N + i`) and libraries within the ecosystem already support negative indices in `tensordot`, will go ahead and merge.

For reference, the original [PR](https://github.com/data-apis/array-api/pull/136) makes no mention regarding the original guidance of nonnegative integers. That the guidance required nonnegative integers and not negative integers was likely an omission/mistake.

PR #632, opened 2023-05-24
Fix invalid escape sequences in some files

asmeurer:
These can be found with the command

python -We:invalid -We::SyntaxWarning -m compileall -f -q src/


rgommers (approved):
Thanks Aaron

rgommers:
As a follow-up, maybe we should add that command in the existing CI job before building the docs? WDYT?

PR #633, opened 2023-05-31
typo fix in abs rfc

hirwa-nshuti:


kgryte (approved):
Thanks, @hirwa-nshuti!

PR #639, opened 2023-06-06
Add missing dtype attribute to iinfo_object and finfo_object

asmeurer:


kgryte:
Thanks, @asmeurer! This appears to have been an oversight in https://github.com/data-apis/array-api/pull/485. Merging...

PR #644, opened 2023-07-05
Fix optional keyword argument in `take` signature

kgryte:
This PR:

- fixes the signature for `take` in which the default value for the `axis` keyword argument should be `None`. Without this change, the `axis` keyword argument is always required (as arguments [must always](https://peps.python.org/pep-3102/) be bound to a value), even for the one-dimensional case. This PR corrects this error in the [original PR](https://github.com/data-apis/array-api/pull/416).
- backports the change to the 2022.12 revision of the array API standard.

Ref: https://github.com/data-apis/array-api-compat/issues/34

### DIFF
--- a/src/array_api_stubs/_2022_12/indexing_functions.py
+++ b/src/array_api_stubs/_2022_12/indexing_functions.py
@@ -1,7 +1,7 @@
 from ._types import Union, array
 
 
-def take(x: array, indices: array, /, *, axis: int) -> array:
+def take(x: array, indices: array, /, *, axis: Optional[int] = None) -> array:
     """
     Returns elements of an array along an axis.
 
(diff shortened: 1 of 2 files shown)


asmeurer:
Looks good. I'm OK with backporting this because it makes the signature match what the text of the specification already said. We will also need to update the compat library and numpy.array_api.

asmeurer:
And the test suite. I'm pretty sure `axis` being optional for 1-D isn't tested because my implementations of `take` would have failed.

rgommers (approved):
LGTM too, in it goes. Thanks @kgryte and @asmeurer. 

There's already https://github.com/data-apis/array-api-compat/issues/34 for the follow-up action.

asmeurer:
I made a fix for numpy.array_api at https://github.com/numpy/numpy/pull/24187

PR #652, opened 2023-07-12
Allow lazy implementations to raise from intrinsically eager functions

### LINKED ISSUE #642: Eager functions in API appear in conflict with lazy implementation
We are looking at adapting this API for a lazy array library built on top of ONNX (and [Spox](https://github.com/Quantco/spox)). It seems to be an excellent fit for most parts. However, certain functions in the specification appear to be in conflict with a lazy implementation. One instance is [`__bool__`](https://data-apis.org/array-api/draft/API_specification/generated/array_api.array.__bool__.html) which is defined as:
```
array.__bool__() → bool
```

The output is documented as "a Python bool object representing the single element of the array". This is problematic for lazy implementations since the value is not available at this point. How should a standard compliant lazy implementation deal with this apparent conflict?

cbourjau:
Allow intrinsically eager functions (such as `array.__bool__`) to raise in lazy implementations. Closes #642

### DIFF
--- a/src/array_api_stubs/_draft/array_object.py
+++ b/src/array_api_stubs/_draft/array_object.py
@@ -239,9 +239,9 @@ class _array:
 
         For complex floating-point operands, special cases must be handled as if the operation is implemented as the logical AND of ``bool(real(self))`` and ``bool(imag(self))``.
 
-        **Lazy implemenations**
+        **Lazy implementations**
 
-        Lazy implementations of the API standard must raise a ``ValueError`` if they are unable eagerly compute the requested value.
+        The Python language requires the return value to be of type ``bool``. Lazy implementations are therefore not able to return any kind of lazy/delayed object here and should raise a ``ValueError`` instead.
 
         .. versionchanged:: 2022.12
             Added boolean and complex data type support.
@@ -277,6 +277,10 @@ class _array:
         - If ``self`` is ``+infinity``, the result is ``+infinity + 0j``.
         - If ``self`` is ``-infinity``, the result is ``-infinity + 0j``.
         - If ``self`` is a finite number, the result is ``self + 0j``.
+
+        **Lazy implementations**
+
+        The Python language requires the return value to be of type ``complex``. Lazy implementations are therefore not able to return any kind of lazy/delayed object here and should raise a ``ValueError`` instead.        
 
         .. versionadded:: 2022.12
         """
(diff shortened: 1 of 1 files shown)


src/array_api_stubs/_draft/array_object.py
```diff
@@ -239,6 +239,10 @@ def __bool__(self: array, /) -> bool:
 
         For complex floating-point operands, special cases must be handled as if the operation is implemented as the logical AND of ``bool(real(self))`` and ``bool(imag(self))``.
 
+        **Lazy implemenations**
+
+        Lazy implementations of the API standard must raise a ``ValueError`` if they are unable eagerly compute the requested value.
```
rgommers, line 244: I suggest to add the following for more context: "the Python language requires the return value to be of type `bool`, hence lazy implementations are not able to return any kind of lazy/delayed object here".

src/array_api_stubs/_draft/array_object.py
```diff
@@ -422,6 +426,10 @@ def __float__(self: array, /) -> float:
         - If ``self`` is ``True``, the result is ``1``.
         - If ``self`` is ``False``, the result is ``0``.
 
+        **Lazy implementations**
+
+        Lazy implementations of the API standard must raise a ``ValueError`` if they are unable eagerly compute the requested value.
```
rgommers, line 431: Could you change "must" to "should"? For why, see https://data-apis.org/array-api/draft/design_topics/exceptions.html

rgommers (commented):
Thanks @cbourjau. This looks like a good start. I added a few minor comments. Then, I think `__complex__` should get the same note, right?

My main suggestion for more context is perhaps something that could go in a short design document on lazy/eager, that could be put under https://data-apis.org/array-api/latest/design_topics/index.html. That also gives us a place to add other future lazy evaluation related topics. I am happy to draft that as a follow-up to this PR; we can get this merged first to clarify things.

cbourjau:
Thanks for the review! I addressed your comments as suggested.

cbourjau:
Is there anything missing in this PR before it could be merged?

rgommers:
> Is there anything missing in this PR before it could be merged?

No, there isn't. My array API review bandwidth disappeared for a bit. In it goes, thanks again @cbourjau!

rgommers:
> My main suggestion for more context is perhaps something that could go in a short design document on lazy/eager, that could be put under https://data-apis.org/array-api/latest/design_topics/index.html. That also gives us a place to add other future lazy evaluation related topics. I am happy to draft that as a follow-up to this PR; we can get this merged first to clarify things.

I'll need to do this now.

rgommers:
Follow-up in gh-708.

Issue #597, opened 2023-02-14
Add support for computing the cumulative sum to the standard


steff456:
This RFC requests to include a new API in the array API specification for the purpose of computing the cumulative sum.

## Overview

Based on array comparison [data](https://github.com/data-apis/array-api-comparison/blob/7ecfa9867ec43a28e914b35868c041443d31d772/data/common_apis.csv#L43), the API is available in all the libraries in the PyData ecosystem.

## Prior art

- NumPy: <https://numpy.org/doc/stable/reference/generated/numpy.cumsum.html>
- PyTorch: https://pytorch.org/docs/stable/generated/torch.cumsum.html
- MXNet: <https://mxnet.apache.org/versions/1.6/api/r/docs/api/mx.nd.cumsum.html>
- TensorFlow: <https://www.tensorflow.org/api_docs/python/tf/math/cumsum>

## Proposal:

```python
def cumsum(x: array, /, *, axis: Optional[Union[int, Tuple[int, ...]]] = None, dtype: Optional[dtype] = None) -> array
```

- `dtype` kwarg is for consistency with `sum` et al

cc @oleksandr-pavlyk

asmeurer:
There's also `np.add.accumulate` which may be relevant when looking at the comparison data.

rgommers:
Based on previous experience with complaints about bad naming (see, e.g., https://github.com/scipy/scipy/issues/12924 and https://www.reddit.com/r/programminghorror/comments/j6sd61/i_was_just_looking_at_the_documentation_for/) I would very much prefer not to enshrine `cumsum` and `cumprod` (as proposed in gh-598) names in this API standard.

This is pretty niche functionality and arguably not "core" enough to implementing an array library for it to be in this standard at all, so I'd vote for leaving it out completely. In a compat layer it could perhaps be named `cumulative_*` like SciPy did with a few methods, if it's desired for these functions to be there.

soraros:
[...]

PR #653, opened 2023-07-13
Add specification for computing the cumulative sum

kgryte:
This PR

- resolves https://github.com/data-apis/array-api/issues/597 by adding a specification for the cumulative sum.
- avoids naming concerns by using the more verbose name `cumulative_sum`. This follows recent convention in the dataframe API standard (ref: https://github.com/data-apis/dataframe-api/pull/193).
- follows `dtype` conventions established in the specifications for `sum`, whereby, if no `dtype` is specified, the data type of the output array must have a range of values at least as large as the default data type of the same kind.
- follows `axis` conventions whereby, for one-dimensional arrays, providing an `axis` is optional; however, if the input array has more than one dimension, providing an `axis` is required. This deviates from NumPy where not providing an `axis` computes a cumulative sum over the flattened array and follows PyTorch where providing an axis is required. The solution here attempts to thread the needle between the two, siding with convention established in `take` where convenience is given greater weight for one-dimensional arrays.
- includes an `include_initial` keyword argument to control whether the initial value (by convention, the additive identity) is included in the output. The `include_initial` naming convention takes inspiration from [NumPy#6044](https://github.com/numpy/numpy/issues/6044) which proposed `include_identity`, but this PR uses `_initial` to align with a future extension in which an `initial` kwarg specifies a custom initial value. By default, `include_initial=False`, which matches the default behavior across the [ecosystem](https://github.com/data-apis/array-api-comparison/blob/2f1ccfcdaca109d2fe74c5e59619c37008b35af6/signatures/statistics/cumulative_sum.md). As of this PR, only TensorFlow supports excluding/including the starting summation value.
- handling of special values (e.g., `NaN` and infinities) must follow special cases as defined in `add`.

## Questions

- Should `axis` be fully optional, as in NumPy?
    - **Update**: during the consortium meeting on 13 July 2023, consensus was to use `take` convention, where NumPy's flattening behavior is **not** supported.
- Bikeshed: alternative naming conventions for `startpoint` and `endpoint`? Or are we okay following specification precedent?
    - **Update**: during the consortium meeting on 13 July 2023, concerns were raised regarding the inclusion of `endpoint`, especially when `endpoint=False` can be achieved by slicing off the last element. Omission of the last element is a common use case (e.g., when computing offsets), but it's explicit support in the API signature did not seem warranted. Further, concerns were raised regarding `startpoint`/`endpoint` naming conventions in that they lacked obviousness. This PR has been updated to use `include_initial` with potential future alignment with future extension in which `initial` kwarg support is added to `cumulative_sum`.


src/array_api_stubs/_draft/statistical_functions.py
```diff
@@ -1,6 +1,66 @@
...
+) -> array:
+    """
+    Calculates the cumulative sum of elements in the input array ``x``.
+
+    Parameters
+    ----------
+    x: array
+        input array. Should have a numeric data type.
+    axis: Optional[int]
+        axis along which a cumulative sum must be computed. If ``axis`` is negative, the function must determine the axis along which to compute a cumulative sum by counting from the last dimension.
```
rgommers, line 21: It seems to me that this should instead match exactly how the `axis` keyword works in `sum` and other reductions?

rgommers (commented):
Thanks @kgryte!

My initial comments:
- I look at `cumulative_sum` like a reduction, which should closely follow what `sum` and other reductions do for its main parameters.
- The use cases about "prepending a value" are tricky because of the output becoming keyword-dependent. As I read it, that is what both `startpoint` and `endpoint` do, right?

   - It kinda breaks our policy regarding output shapes depending on keywords, so I think we  need an explicit okay for this from JAX & co.
   - Most of the discussion was specifically about prepending zero, which is `startpoint`. I'm less convinced by the inclusion of `endpoint`.

- `initial` is indeed useful to leave out here, because if we want to add it we should do that uniformly for all reductions.

src/array_api_stubs/_draft/statistical_functions.py
kgryte: A few comments. (1) NumPy et al and PyTorch don't support providing multiple axes. So that is already an immediate difference. (2) As mentioned on the [RFC](https://github.com/data-apis/array-api/issues/597#issuecomment-1494778565), flattening behavior is a bit odd and has a less, IMO, clear use case than `sum`, where computing the sum of all elements in an array can make sense, but the cumulative sum, less so.

src/array_api_stubs/_draft/statistical_functions.py
```diff
@@ -1,6 +1,66 @@
...
+
+        If the data type (either specified or resolved) differs from the data type of ``x``, the input array should be cast to the specified data type before computing the sum. Default: ``None``.
+
+        .. note::
+           keyword argument is intended to help prevent data type overflows.
+
+    startpoint: bool
+        boolean indicating whether to include the initial value as the first value in the output. Default: ``False``.
+    endpoint: bool
+        boolean indicating whether to include the total sum as the last value in the output. Default: ``True``.
```
seberg, line 42: I will note that `initial=` would be a useful extension to allow starting with something other than 0, especially together with `startpoint=True`.  (It might even be necessary in case an identity isn't defined: such as for timedelta's were there startpoint has to be a datetime.)

Mainly mentioning it in case anyone is inspired about startpoint/endpoint naming, I don't mind adding such parameters (not sure how hard it is to add to numpy, but I don't expect it is hard).

src/array_api_stubs/_draft/statistical_functions.py
```diff
@@ -1,6 +1,66 @@
...
+        -   if the default data type corresponding to the data type "kind" of ``x`` has the same or a larger range of values than the data type of ``x``,
+            -   if ``x`` has a real-valued floating-point data type, the returned array must have the default real-valued floating-point data type.
+            -   if ``x`` has a complex floating-point data type, the returned array must have the default complex floating-point data type.
+            -   if ``x`` has a signed integer data type (e.g., ``int16``), the returned array must have the default integer data type.
+            -   if ``x`` has an unsigned integer data type (e.g., ``uint16``), the returned array must have an unsigned integer data type having the same number of bits as the default integer data type (e.g., if the default integer data type is ``int32``, the returned array must have a ``uint32`` data type).
+
```
seberg, line 33: Can you just refer to `sum`?  (I am still a bit unsure that I find the float behavior useful in practice, since libraries cannot rely on it as the "default" can vary; also true for int's, but there is less prior art).

asmeurer: In general in the spec we duplicate things across function definitions rather than cross-linking.

adityagoel4512:
Is there a reason this PR stalled? I think cumulative sums are quite pervasive and would be a good candidate for standardisation.

rgommers:
No reason beyond limited bandwidth. This function will definitely make it into the next release.

src/array_api_stubs/_draft/statistical_functions.py
kgryte: Additionally, for flattening, given that order matters when computing the cumulative sum, I think it is preferable for users to explicitly reshape prior to calling `cumulative_sum`. While row-major order may the desired order, it's not clear that that is an obvious choice in this case. As such, I don't think we want to replicate how the `axis` kwarg works in `sum`.

kgryte:
**Update**: I added guidance concerning what happens when `include_initial` is `True` and yet `x` is an empty array. My sense is that the returned array should also be empty, but others may have differing opinions. (related comment on NumPy mailing list: https://mail.python.org/archives/list/numpy-discussion@python.org/message/P5ZTUQ3L4Y2EKXVETYWWZLDOE2TGJBPJ/)

kgryte:
**Update**: based on feedback during Consortium discussions, I've updated the guidance concerning what should happen when `include_initial` is `True` and the array is empty.

The guidance now states that, no matter the axis size (>=0), the output array shape must follow the same rule: namely, the size of the axis along which to compute the cumulative sum must be `N+1`, with the first element equal to the initial value. Hence, if the input array `x` is empty and `include_initial` is `True`, then the size of the corresponding dimension in the output array must be `1`.

kgryte:
@seberg Would you be willing to do another round of review? Would be good to get another set of eyes on this before we merge. Thanks!

src/array_api_stubs/_draft/statistical_functions.py
kgryte: I'll go ahead and mark this as resolved.

seberg:
FWIW, I am happy with `include_initial` and all the other things were general things shared with normal sum anyway.

kgryte:
As this PR has been discussed previously and no objections were raised, will go ahead and merge. Any further updates or clarifications can be addressed in follow-ups PRs.

Issue #483, opened 2022-09-23
Adding `moveaxis` & `swapaxes`


jakirkham:
There are a few operations that come up like this periodically (also `rollaxis`). That said, some of these require a doc refresher on how they behave before using them. Still there can be value in doing something simpler than transpose that makes it clearer what changed. Perhaps an alternative approach to adding them to the spec is making sure the NumPy implementations play well with other array implementations.

rgommers:
`swapaxes` is problematic IIRC, see gh-228. That's why we called it [`permute_dims`](https://data-apis.org/array-api/latest/API_specification/generated/signatures.manipulation_functions.permute_dims.html)

rgommers:
We would benefit from a description of how the many ways of doing this in numpy et al. translate to the preferred way of doing this with the array API. That will show also if there's a gap like `moveaxis`. IIRC we looked at this fairly carefully, however I cannot find it back right now.

shoyer:
`permute_dims` is the fundamental operation here. `moveaxis` is convenient for users, but do we also want to include convenience functions in the API standard?

rgommers:
> but do we also want to include convenience functions in the API standard?

I think we parked the many convenience functions previously, and focused on getting the one fundamental operation right. I'd say there is a case for one convenience function here though. Implementing `moveaxis` in terms of `permute_dims` is not completely trivial. Looking at the `numpy.moveaxis` implementation, if you want to get negative axes and other corner cases right, it'd be something like:
```python
def moveaxis(x, /, source, destination):
    source = _normalize_axis_tuple(source, x.ndim, 'source')
    destination = _normalize_axis_tuple(destination, x.ndim, 'destination')
    if len(source) != len(destination):
        raise ValueError('`source` and `destination` arguments must have '
                         'the same number of elements')

    order = [n for n in range(x.ndim) if n not in source]
    for dest, src in sorted(zip(destination, source)):
        order.insert(dest, src)

def _normalize_axis_tuple(axis, ndim, argname=None, allow_duplicate=False):
    if type(axis) not in (tuple, list):
        try:
            axis = [operator.index(axis)]
        except TypeError:
            pass

    axis = tuple([normalize_axis_index(ax, ndim, argname) for ax in axis])
    if not allow_duplicate and len(set(axis)) != len(axis):
        if argname:
            raise ValueError('repeated axis in `{}` argument'.format(argname))
        else:
            raise ValueError('repeated axis')
    return axis
```

For the other functions mentioned:
[...]

PR #656, opened 2023-07-13
Add specification for moving array axes to new positions

kgryte:
This PR

- resolves https://github.com/data-apis/array-api/issues/483 by adding a specification for `moveaxis`.
- [signatures](https://github.com/data-apis/array-api-comparison/blob/ad97bfd878ae7268d668f04fe1e2ace7319c73e1/signatures/manipulation/moveaxis.md) across array libraries are consistent. Only TensorFlow does not have `moveaxis` in its main namespace, but it does include it in its experimental NumPy namespace.

## Questions

- Bikeshed: we intentionally ensure consistent naming with `permute_dims` and `expand_dims`. Should we be consistent here in renaming `moveaxis` to, e.g., `move_dims`? PyTorch predominantly uses the `dims` convention and aliases `moveaxis` as `movedim`, which is not a perfect match given the missing `_` and plural form `dims`.
    - **Update**: leaving name as is as `moveaxis` given widespread array library usage of `moveaxis`. Renaming to something consistent with `permute_dims` and `expand_dims` would likely cause unnecessary churn for insufficient benefit.


rgommers:
This looks fairly straightforward, nice. 

> Bikeshed: we intentionally ensure consistent naming with `permute_dims` and `expand_dims`. Should we be consistent here in renaming `moveaxis` to, e.g., `move_dims`?

Given signatures for all libraries match and the `moveaxis` name is fine, I think this is about as good an example as we can have for "go with prior art".

kgryte:
During the consortium workgroup meeting on 27 July 2023, consensus was to move this API forward and that `moveaxis` is the preferred name given its widespread usage in the array library ecosystem.

Accordingly, this PR should be ready for merge.

kgryte:
As this PR is straightforward, the API has broad support across the ecosystem, and there is working group consensus to move this API forward, I'll go ahead and merge. Any further updates prior to 2023 spec finalization can be made in subsequent PRs. Merging...

PR #659, opened 2023-07-15
Fix documentation for constants

### LINKED ISSUE #601: Bug: descriptions for constants included in the standard are incorrect
See screenshot below:

image

This affects all currently rendered specification versions.

asmeurer:
Sphinx cannot find the docstring corresponding to a constant unless you point it to the exact file where it is, since it doesn't actually live on the object itself. See https://github.com/sphinx-doc/sphinx/issues/6495

Fixes #601


kgryte (approved):
Thanks, @asmeurer!

PR #681, opened 2023-09-07
Use parameter name obj in asarray device doc

aaronmondal:
The asarray creation function uses obj instead of x.


honno (approved):
Good catch!

rgommers (approved):
LGTM too, thanks @aaronmondal

PR #683, opened 2023-09-12
Move `__all__` to PEP-8 recommended location

nstarman:
> Module level "dunders" (i.e. names with two leading and two trailing underscores) such as `__all__`, `__author__`, `__version__`, etc. should be placed after the module docstring but before any import statements except from `__future__` imports. Python mandates that future-imports must appear in the module before any other code except docstrings.

— https://pep8.org


rgommers (approved):
Makes sense, thanks @nstarman

PR #684, opened 2023-09-12
Add pre commit + black

nstarman:
A maintainer will need to turn on `pre-commit.ci` to enforce pre-commit in the CI.


rgommers (commented):
Thanks @nstarman. The `black` reformatting looks like an improvement, in particular the signatures using one line per argument. I'm +1 on that change. 

The larger style problem in this repo is that the docstring text doesn't have proper line breaks. I've looked at that before and unfortunately there's no tool that can deal with that automatically.

For pre-commit: I'd be fine with adding a check to CI (either pre-commit or simply `black` running in a CI job). However, I'd much prefer to get no bot PRs, those are just noise for both contributors watching the repo and in the commit history. Can we get this without `autoupdate_schedule`?

nstarman:
> The larger style problem in this repo is that the docstring text doesn't have proper line breaks. I've looked at that before and unfortunately there's no tool that can deal with that automatically.

At Astropy we've just started to use PyQA's `docformatter` in our pre-commit. `docformatter` can do some text regularization automatically.


> For pre-commit: I'd be fine with adding a check to CI (either pre-commit or simply `black` running in a CI job). However, I'd much prefer to get no bot PRs, those are just noise for both contributors watching the repo and in the commit history. Can we get this without `autoupdate_schedule`?

Yes, using the GH action
https://pre-commit.ci/lite.html
I can add this to the PR.

rgommers:
Thanks! I tried some tools last year, but `docformatter` seems to have seen very active development recently - will try again.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,54 @@
+ci:
+  autofix_prs: false
+  autoupdate_schedule: 'monthly'
```
rgommers, line 3: The [docs](https://pre-commit.ci/#configuration-autoupdate_schedule) for this option seem to say that there's no way to turn this off. However, https://pre-commit.ci/lite.html seems to say that this `ci:` block can go away completely? I'm not sure if those docs are correct, it doesn't really make sense to me that the `lite-action` requires the full action with `uses: pre-commit/action@v3.0.0`.

Could you see if there's a way, and if not then set this to `quarterly` as the slowest document frequency?

rgommers (commented):
Thanks @nstarman, this is looking pretty good to me now. Only one more comment.

.pre-commit-config.yaml
nstarman: It can be removed!

rgommers (approved):
LGTM now, thanks a lot @nstarman.

I'll plan to squash-merge this very soon, and then push a follow-up commit with a new `.git-blame-ignore-revs` entry.

Issue #640, opened 2023-06-15
RFC: add a unified inspection API namespace


kgryte:
This RFC is intended to supersede https://github.com/data-apis/array-api/issues/635, https://github.com/data-apis/array-api/issues/636, https://github.com/data-apis/array-api/issues/637, and https://github.com/data-apis/array-api/issues/638, based on initial consortium feedback, and proposes to consolidate inspection APIs into a single non-public namespace.

# Proposal

```python
info = xp.__array_namespace_info__()
```

Returns a namespace with Array API namespace inspection utilities.

_Note: the `x.__array_namespace__()` API supports a `version` kwarg. We don't follow suit here, as we assume backward compatibility, whereby a consumer may ask for a particular version of the array API standard and a conforming library may return a newer version. That version should still be compliant with the older specified version (backward compat guarantee), and thus, the inspection API return values should still be applicable. Stated more succinctly: conforming array libraries are expected to be "evergreen" and only have a single Array API implementation, not separate implementations for each revision of the standard._

### info.capabilities() -> dict[str,bool]

```python
info.capabilities() -> dict[str,bool]
```

Returns a dictionary of **static** "capabilities", such as whether a conforming library supports boolean indexing or data-dependent output shapes.

Key values should be either `False` (does not support) or `True` (does support). For example,

```text
{
    'boolean_indexing': False | True,
    'data_dependent_shapes': False | True
}
```

### info.default_device() -> device

```python
info.default_device()
```

Returns an object for the default device.

### info.default_dtypes( /, \*, device: device) -> dict\[str, dtype]

```python
info.default_dtypes() -> dict[str, dtype]
info.default_dtypes(device: device) -> dict[str, dtype]
```

Returns a dictionary having the following keys:

- **real floating**: default real floating-point dtype.
- **complex floating**: default complex floating-point dtype.
- **integral**: default integral dtype.
- **indexing**: default index dtype.

```python
>>> info.default_dtypes()
{
    'real floating': <dtype>,
    'complex floating': <dtype>,
    'integral': <dtype>,
    'indexing': <dtype>
}
>>> info.default_dtypes(device=<device>)
{
    'real floating': <dtype>,
    'complex floating': <dtype>,
    'integral': <dtype>,
    'indexing': <dtype>
}
```

More keys could be added in the future, depending on evolution of the standard.

If not provided a `device` argument, the function would return a dictionary based on the current device (context). If provided a `device` argument, the function would return a dictionary as described above, but specific to the specified device.

[...]

Issue #499, opened 2022-10-20
ENH: Should we allow querying certain implementation details


seberg:
I am wondering if it would make sense to standardize a bit of information about the exported namespace.  For example

* immutability: If arrays are always immutable, sometimes a copy may not be needed?
* Is it interesting to know whether in-place operations are actually in-place?

I somewhat suspect that such information will be useful eventually, but not sure that we have anything explicit right now that would be.

asmeurer:
Should this be a query on the namespace or on an array object?

seberg:
I was thinking of namespace, but maybe immutability would need to be array object specific and not namespace...  I suppose we can probably close this for now though, I think it may be useful to "bag" details, but right now I am not sure if users actually have a need for any information :).

kgryte:
@seberg Can you think of any practical use cases where an array API consumer would/should branch based on the implementation info you mention in the OP?

E.g., what would be an example of when an array API consumer should be concerned with whether an array library supports mutation? You mention determining whether a copy is needed; any practical examples of where this would be applicable and a downstream library responsibility rather than in the array library itself?

seberg:
The thing why I was wondering was the discussion about esuring you have a copy:
```
api = getnamespace
if not api.mutable:
    # ensure user input cannot be mutated:
    arr = arr.copy()

# continue using in-place ops:
arr += 1
arr *= 5
```

seberg:
In the meeting there were a few other things that came up and may be interesting:
* Support for non-native byte order
* Actually supported dtypes
* Supported devices

rgommers:
> ```
> api = getnamespace
> if not api.mutable:
>     # ensure user input cannot be mutated:
>     arr = arr.copy()
> ```

I assume the second line was supposed to be `if api.mutable:`? Either way, not sure this is a convincing example, because for immutable arrays `copy()` should be a no-op, so you may as well leave out the check.

For the other things, that seems like potentially useful info when using a single library. But when you're using the standard API, I'm wondering what you are going to do with something like "yes, supports non-native byte order". Because it doesn't come back anywhere else in the API, so it seems more like an FYI.

The actual dtypes and devices I can imagine using; they can be passed on to `dtype=` or `device=` keywords.

rgommers:
I was reminded of this issue because of gh-609. It's still a pain that we have data-dependent features in the API that are optional to implement. So for writing portable and performant code across those libraries that do and don't perform it, we need something like:
```python
[...]

PR #689, opened 2023-09-20
Add Array API inspection utilities

kgryte:
This PR

- resolves https://github.com/data-apis/array-api/issues/499 and resolves https://github.com/data-apis/array-api/issues/640 by adding a unified API for inspecting the Array API capabilities for specification-conforming array libraries. The principle details of this implementation were proposed and discussed in the referenced RFC.


src/array_api_stubs/_draft/info.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+def dtypes(
+    *,
+    device: Optional[device] = None,
+    kind: Optional[Union[str, Tuple[str, ...]]] = None,
+) -> DataTypes:
+    """
+    Returns a dictionary of supported *Array API* data types.
+
```
rgommers, line 98: Perhaps it would be useful to explicitly note that if a dtype is included as supported here, it must have the expected properties? Meaning in particular that JAX must not include `float64` here if it aliases that to `float32`.

kgryte: Good suggestion. Will add.

src/array_api_stubs/_draft/_types.py
```diff
@@ -51,6 +56,65 @@
 # to keep pyflakes happy. https://github.com/python/typeshed/issues/3556
 ellipsis = TypeVar("ellipsis")
 
+DefaultDataTypes = TypedDict(
```
rgommers, line 59: I'm not sure that you can use `TypedDict` like this. It's a type hint only for regular dictionaries, and the docs always use a class which inherits from `TypedDict`: https://docs.python.org/3/library/typing.html#typing.TypedDict.

rgommers:
This looks pretty good to me overall, thanks @kgryte.

src/array_api_stubs/_draft/_types.py
kgryte: Do you have a preferred alternative? I was not sure the best approach for the sake of typing.

src/array_api_stubs/_draft/info.py
kgryte: Done.

src/array_api_stubs/_draft/_types.py
rgommers: I'd be inclined to go with a regular `dict[str: dtype]`. However, let's see first if someone else has a more informed opinion - I have never used `TypedDict` before, so I meant my "I'm not sure" - this could well be fine.

BvB93: It should be alright I think: the `TypedDict` docs do state that this function-style syntax is supported, though the more conventional class-style syntax is generally preferred (the main use case for the function-style is dictionary keys with, _e.g._, hyphens or other characters that would lead to invalid symbols). 

Actually, is there any reason for _not_ using the class-style syntax here?

src/array_api_stubs/_draft/_types.py
```diff
@@ -51,6 +56,65 @@
...
+    },
+    total=False,
+)
+Capabilities = TypedDict(
+    "Capabilities", {"boolean_indexing": bool, "data_dependent_shapes": bool}
+)
+
+
+@dataclass
+class Info:
```
BvB93, line 93: Small note mostly relevant for downstream usage: dataclasses very much follow a nominal subtyping pattern (as opposed to protocols), so the only way downstream `Info` implementations can properly subtype `Info` is by directly inheriting from this very specific class here. Assuming this is undesired, might it be an idea to type it as a protocol instead?

Also what's going on with all the `@staticmethod` decorators? Is this something that's necessary due to some dataclass-specific reasons?

src/array_api_stubs/_draft/_types.py
kgryte: @BvB93 Some of the dictionary keys contain spaces (e.g., `real floating` and `complex floating`).

For the other two dictionaries, I used the function syntax for consistency. I am happy to update if folks have strong opinions here.

src/array_api_stubs/_draft/_types.py
kgryte: @BvB93 No, no dataclass-specific reasons. Do you have recommendations for how to type returning a namespace? Specifically, this `Info` class is intended to document the return value for `__array_namespace_info__()`. If you have recommendations for the right way to type this, LMK.

kgryte: In this instance, my intent was to type an object with various static methods.

kgryte: It's quite possible that typing as a protocol makes more sense, as proposed in https://github.com/data-apis/array-api/pull/685.

BvB93: Yeah, I'd definitely recommend protocol usage here as, in its most basic forms, those are used to denote "an arbitrary object with X attributes and Y methods", so they're fully duck-typable.

kgryte: Refactored using protocols, so I'll mark this as resolved.

src/array_api_stubs/_draft/_types.py
kgryte: As the functional syntax for `TypedDict` is supported and `DefaultDataTypes` contains keys which are not valid identifiers (`real floating`), I'll leave as is and mark this as resolved.

kgryte:
I've updated this PR to use a protocol instead of a data class. I believe this PR should be ready for final review.

rgommers (commented):
The way this PR is written and without looking at the linked discussions, it's not quite clear to me whether `__array_namespace_info__` is an array method or a function in the namespace returned by `array.__array_namespace__`.

In addition, it could be useful to clarify in the `__array_namespace_info__` docs that the returned object from that function/method may be either a namespace or a class; as long as the user can do this:
```python
info = __array_namespace_info__()
info.capabilities()
info.devices()
info.dtypes()
info.default_dtypes()
```

rgommers:
The rendered docs look fine, maybe some explanation under `Inspection APIs` and right above the table of function/method names will help?

image

rgommers:
IIRC there were a few more things discussed for inclusion in the inspection API:
- the max number of dimensions: gh-694
- whether or not read-only arrays are supported
- eager/lazy (maybe, not sure)

kgryte:
> IIRC there were a few more things discussed for inclusion in the inspection API

I was planning on submitting follow-up PR(s) to avoid potential bikesheds and avoid piggybacking on this PR, but I can include some of the aforementioned additional capabilities in this PR. I don't have a strong opinion, except wanting to get this in sooner rather than later.

src/array_api_stubs/_draft/info.py
```diff
@@ -0,0 +1,149 @@
...
+    .. note::
+       While specification-conforming array libraries may support additional data types which are not present in this specification, data types which are not present in this specification should not be included in the returned dictionary.
+
+    .. note::
+       Specification-conforming array libraries must only return supported data types having expected properties as described in :ref:`data-types`. For example, if a library decides to alias ``float32`` as ``float64``, that library must not include ``float64`` in the dictionary of supported data types.
+
+    Parameters
+    ----------
+    kind: Optional[Union[str, Tuple[str, ...]]]
```
leofang, line 107: Silly question irrelevant of this PR: Can't Sphinx just check the type annotations in the function signature, and populate it here? Seems repetitive to me having to copy/paste. Just curious...

src/array_api_stubs/_draft/info.py
```diff
@@ -0,0 +1,149 @@
...
+) -> DefaultDataTypes:
+    """
+    Returns a dictionary containing default data types.
+
+    The dictionary should have the following keys:
+
+    -   **real floating**: default real floating-point data type.
+    -   **complex floating**: default complex floating-point data type.
+    -   **integral**: default integral data type.
+    -   **indexing**: default index data type.
```
leofang, line 74: I don't recall how this one is supposed to work, e.g., why would users care, which API does this impact, etc?

src/array_api_stubs/_draft/info.py
```diff
@@ -0,0 +1,149 @@
...
+    *,
+    device: Optional[device] = None,
+) -> DefaultDataTypes:
+    """
+    Returns a dictionary containing default data types.
+
+    The dictionary should have the following keys:
+
+    -   **real floating**: default real floating-point data type.
+    -   **complex floating**: default complex floating-point data type.
```
leofang, line 72: Q: I noticed the keys here (and later) have space between words, while those in `capabilities()` have underscores, maybe we can unify the style?

src/array_api_stubs/_draft/info.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+def __array_namespace_info__() -> Info:
+    """
+    Returns a namespace with Array API namespace inspection utilities.
+
+    Returns
+    -------
+    out: Info
+        Namespace with Array API namespace inspection utilities.
```
leofang, line 31: xref: https://github.com/data-apis/array-api/pull/689#pullrequestreview-1816074339 

We should try to expand the explanation for the returned object. Either here or in `inspection.rst` is fine to me.

leofang (commented):
Done my first pass, LGTM so far!

src/array_api_stubs/_draft/_types.py
leofang: > As the functional syntax for `TypedDict` is supported and `DefaultDataTypes` contains keys which are not valid identifiers (`real floating`), I'll leave as is and mark this as resolved.

Reopen as per my question below: https://github.com/data-apis/array-api/pull/689#discussion_r1449754371

src/array_api_stubs/_draft/info.py
kgryte: Not sure. Thus far, we've always just duplicated the types. I've left a copy on https://github.com/data-apis/array-api/issues/533#issuecomment-1909414609, as a follow-up maintenance task.

src/array_api_stubs/_draft/info.py
kgryte: These match the accepted "kind" values in the `isdtype` API. In hindsight, it may have been better to have the "kind" values be snakecase, as arguably more idiomatic, but that ship has sailed. In which case, I'll update the capabilities to use spaces.

src/array_api_stubs/_draft/info.py
kgryte: This API provides a means of programmatically querying the default data types for the different defaults described in the specification. E.g., for a given dtype "kind", this API would allow downstream libraries to check if the default dtype is of a lower (or higher) precision than desired, or what the default dtype is for a specific device (e.g., on CPU, this could be "float64", but on GPU, this could be "float32", etc). When working with a library supporting multiple device types, a downstream library wanting a specific minimum precision could then choose a device based on which device automatically supports operations having the desired default device type.

kgryte:
> The way this PR is written and without looking at the linked discussions, it's not quite clear to me whether `__array_namespace_info__` is an array method or a function in the namespace returned by `array.__array_namespace__`.

I modeled the documentation of `__array_namespace_info__` based on [`__array_api_version__`](https://data-apis.org/array-api/latest/API_specification/version.html), which is documented as an attribute in the namespace returned by `array.__array_namespace__`. Accordingly, `__array_namespace_info__` is intended as a top-level function. During discussions, we made it a dunder as it is not intended to be a downstream end-user facing API.

kgryte:
> In addition, it could be useful to clarify in the `__array_namespace_info__` docs that the returned object from that function/method may be either a namespace or a class...

I've added a corresponding note to the `__array_namespace_info__` docstring.

src/array_api_stubs/_draft/info.py
kgryte: I've added a note in the function docstring and also added an explanation in `inspection.rst`. Will go ahead and mark this as resolved.

kgryte:
@rgommers The updated `inspection.rst` file now renders as

Screenshot 2024-01-25 at 1 55 06 AM

kgryte:
@rgommers @leofang I believe I have addressed your most recent review comments, and this PR should be ready for another round of review.

src/array_api_stubs/_draft/info.py
```diff
@@ -0,0 +1,163 @@
...
+        a dictionary containing supported data types.
+
+        .. note::
+           Dictionary keys must only consist of canonical names as defined in :ref:`data-types`.
+    """
+
+
+def devices() -> List[device]:
+    """
+    Returns a list of supported devices.
```
rgommers, line 157: May want to clarify that this is a runtime thing? E.g., for a library that supports both CPU and GPU, it should not return a GPU device that is not available on the current machine.

So maybe it should be worded as "available" instead of "supported"?

kgryte: Updated the copy to

> Returns a list of supported devices which are available at runtime.

src/array_api_stubs/_draft/info.py
```diff
@@ -0,0 +1,163 @@
...
+
```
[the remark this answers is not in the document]
kgryte: I would think it should return whatever is applicable when setting the `device` kwarg in the creation APIs.

src/array_api_stubs/_draft/info.py
```diff
@@ -0,0 +1,163 @@
...
+    kind: Optional[Union[str, Tuple[str, ...]]] = None,
+) -> DataTypes:
+    """
+    Returns a dictionary of supported *Array API* data types.
+
+    .. note::
+       While specification-conforming array libraries may support additional data types which are not present in this specification, data types which are not present in this specification should not be included in the returned dictionary.
+
+    .. note::
+       Specification-conforming array libraries must only return supported data types having expected properties as described in :ref:`data-types`. For example, if a library decides to alias ``float32`` as ``float64``, that library must not include ``float64`` in the dictionary of supported data types.
```
rgommers, line 117: Most relevant for JAX I think, so pinging @jakevdp for visibility. I think it'll actually be useful to be able to write portable code also for JAX in SciPy & co, while being fully backwards-compatible for JAX users since the current `float64` aliasing to `float32` is unlikely to change. WDYT @jakevdp?

src/array_api_stubs/_draft/info.py
leofang: Thanks for the explanation @kgryte. I guess I shouldn't have been vague. I understand the purpose of default data types (int/float), they all sound good. My specific question is on the `indexing` one, which I am clueless how knowing this would help (ex: which API mentions default indexing type?), and how a compliant library should report, which there's dynamic mode switching (ex: CuPy can decide on the fly the indexing type used in a JIT'd kernel)?

src/array_api_stubs/_draft/info.py
leofang: ^^^^^ @jakevdp (in case you missed the ping) 🙂

jakevdp: If returning `float64` in certain cases is a requirement for the array API, then it sounds like JAX is only fully-compatible with the Array API when `JAX_ENABLE_X64=True`. That seems fine, I think.

src/array_api_stubs/_draft/info.py
kgryte: @leofang We discuss the default array index data type in the following locations:

- Data Types: https://data-apis.org/array-api/2022.12/API_specification/data_types.html#default-data-types
- argmax: https://data-apis.org/array-api/2022.12/API_specification/generated/array_api.argmax.html#array_api.argmax
- argmin: https://data-apis.org/array-api/2022.12/API_specification/generated/array_api.argmin.html#array_api.argmin
- nonzero: https://data-apis.org/array-api/2022.12/API_specification/generated/array_api.nonzero.html#array_api.nonzero
- argsort: https://data-apis.org/array-api/2022.12/API_specification/generated/array_api.argsort.html#array_api.argsort
- unique\*: https://data-apis.org/array-api/2022.12/API_specification/generated/array_api.unique_all.html#array_api.unique_all

In short, wherever we return indices, we state that the default return data type be the default array index data type.

As for dynamic mode switching, I don't think we have considered this. But wouldn't this apply similarly, in theory, to all default data types? Couldn't a library choose to dynamically switch the "default" floating-point data type used in a JIT'd kernel?

kgryte: But also, correct me if I am wrong, dynamic mode switching seems more like an implementation detail, whereas this API concerns the default array type used for indexing into an array in Array API consuming code.

Additionally, in, e.g., `take`, I can imagine Array API consumers wanting to allocate the `indices` array according to the respective array index data type to avoid unnecessary casts. Having a way to query that data type would help ensure alignment.

Am I missing something here?

leofang: Thanks, @kgryte, I guess I totally overlooked (or forgot about) the indexing data type. It makes sense for the `arg*` or any index-returning APIs to consider.

> But wouldn't this apply similarly, in theory, to all default data types? Couldn't a library choose to dynamically switch the "default" floating-point data type used in a JIT'd kernel?

Yes, in theory. However, I would think it's a lot easier to fix the default int/float data types, and only vary the indexing data type dynamically based on the input size (which offers a natural condition for run-time dispatching). Though, in practice it seems NumPy/CuPy both do have the notion of a default indexing data type, so my job should be easy and I won't complain 😄
```py
>>> import numpy as np
>>> a = np.random.random(2**15)
>>> b = a.argsort()
>>> b.dtype  # it's int64, despite the problem fits in int32
dtype('int64')
```

leofang: Sorry, didn't noticed your 2nd reply until I hit send. No, I think we're good here. I resolved this thread.

kgryte: Thanks, @leofang!

src/array_api_stubs/_draft/info.py
```diff
@@ -0,0 +1,163 @@
...
+    "devices",
+    "dtypes",
+]
+
+from ._types import (
+    Optional,
+    Union,
+    Tuple,
+    List,
+    device,
```
leofang, line 15: I stared this `device` for a few seconds, and realized that elsewhere (e.g. `array_object.py`) we do `import device as Device` and use the capitalized name. Nothing is wrong, just a minor note 🙂

leofang (approved):
LGTM from my perspective. I'm keen to see this PR merged so that we can start addressing follow-up works (ex: https://github.com/data-apis/array-api/pull/689#issuecomment-1887613258). 

Ping @oleksandr-pavlyk for review, due to its relevance 🙂

src/array_api_stubs/_draft/info.py
kgryte: We are considering relaxing dtype requirements, as this is one of the main friction points among array libraries. Part of the impetus for the inspection API is to allow dynamically querying dtype support, and we'll revisit mandatory dtype support in 2024.

The main point of the referenced comment is that array libraries are not allowed to alias dtypes in the list of returned dtypes which don't actually correspond to the underlying storage. E.g., exposing a "float64" dtype, where in reality it is "float32". 

I think we are good here, so I'll go ahead and mark as resolved.

src/array_api_stubs/_draft/info.py
kgryte: Alas, in the creation function documentation, we use lowercase `device`, similar to `dtype`, and `array`. I cannot recall why we aliased as `Dtype` and `Device` in the array object spec. 😞

src/array_api_stubs/_draft/info.py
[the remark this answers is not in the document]
kgryte: As discussed a previous workgroup meeting, I added a note (see https://github.com/data-apis/array-api/pull/689/commits/98a3b9002dac020bdeb05e255a2226f52741d6a3) stating that each returned device object must be valid as a kwarg in creation functions. I don't believe we need to state anything more than this, so I'll go ahead and mark as resolved.

src/array_api_stubs/_draft/info.py
kgryte: I'll mark this as resolved, as the only place where we alias is in `array_object.py`, which is, I believe, to avoid name conflicts and make Sphinx happy.

spec/draft/API_specification/inspection.rst
```diff
@@ -0,0 +1,40 @@
...
+   :toctree: generated
+   :template: method.rst
+
+   __array_namespace_info__
+
+
+Inspection APIs
+---------------
+
+In the namespace (or class) returned by ``__array_namespace__info__``, a conforming implementation of the array API standard must provide and support the following functions (or methods) for programmatically querying data type and device support, capabilities, and other specification-defined implementation-specific behavior, as documented in the functions described below.
```
rgommers, line 27: typo: `__array_namespace__info__` has one underscore too many before `info`.

rgommers (approved):
LGTM too now, modulo the one typo.

spec/draft/API_specification/inspection.rst
kgryte, line 27:
```suggestion
In the namespace (or class) returned by ``__array_namespace_info__``, a conforming implementation of the array API standard must provide and support the following functions (or methods) for programmatically querying data type and device support, capabilities, and other specification-defined implementation-specific behavior, as documented in the functions described below.
```

kgryte: Addressed.

oleksandr-pavlyk (approved):
LGTM!

src/array_api_stubs/_draft/info.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+def capabilities() -> Capabilities:
+    """
+    Returns a dictionary of array library capabilities.
+
+    The dictionary should contain the following keys:
+
+    -   **boolean indexing**: boolean indicating whether an array library supports boolean indexing.
+    -   **data-dependent shapes**: boolean indicating whether an array library supports data-dependent output shapes.
```
asmeurer, line 56: I probably missed the discussion on this, but what's the difference between these two things?

asmeurer: Also it would be good to be specific here about what a `True` value for the capability must entail.

kgryte: Boolean indexing is related, but a library can support APIs which have data-dependent output shapes (e.g., `unique`), but, in principle, not boolean indexing, and vice versa.

leofang: Q: @asmeurer is your question resolved?

kgryte: @asmeurer I included guidance about what a `True` value must entail. Does this satisfy your concern?

asmeurer: Yes.

src/array_api_stubs/_draft/info.py
```diff
@@ -0,0 +1,170 @@
...
+) -> DefaultDataTypes:
+    """
+    Returns a dictionary containing default data types.
+
+    The dictionary must have the following keys:
+
+    -   **real floating**: default real floating-point data type.
+    -   **complex floating**: default complex floating-point data type.
+    -   **integral**: default integral data type.
+    -   **indexing**: default array index data type.
```
leofang, line 88: nit: in terms of style, `kind` below has monospace strings for keys, maybe we should do it here too?
```suggestion
    -   `"real floating"`: default real floating-point data type.
    -   `"complex floating"`: default complex floating-point data type.
    -   `"integral"`: default integral data type.
    -   `"indexing"`: default array index data type.
```

src/array_api_stubs/_draft/info.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+def capabilities() -> Capabilities:
+    """
+    Returns a dictionary of array library capabilities.
+
+    The dictionary must contain the following keys:
+
+    -   **boolean indexing**: boolean indicating whether an array library supports boolean indexing. If a conforming implementation fully supports boolean indexing in compliance with this specification (see :ref:`indexing`), the corresponding dictionary value must be ``True``; otherwise, the value must be ``False``.
+    -   **data-dependent shapes**: boolean indicating whether an array library supports data-dependent output shapes. If a conforming implementation fully supports all APIs included in this specification (excluding boolean indexing) which have data-dependent output shapes, as explicitly demarcated throughout the specification, the corresponding dictionary value must be ``True``; otherwise, the value must be ``False``.
```
leofang, line 56: ditto
```suggestion
    -   `"boolean indexing"`: boolean indicating whether an array library supports boolean indexing. If a conforming implementation fully supports boolean indexing in compliance with this specification (see :ref:`indexing`), the corresponding dictionary value must be ``True``; otherwise, the value must be ``False``.
    -   `"data-dependent shapes"`: boolean indicating whether an array library supports data-dependent output shapes. If a conforming implementation fully supports all APIs included in this specification (excluding boolean indexing) which have data-dependent output shapes, as explicitly demarcated throughout the specification, the corresponding dictionary value must be ``True``; otherwise, the value must be ``False``.
```

leofang (approved):
Two nits on formatting, but content-wise we're good to go!

kgryte:
Thanks, @leofang! Made the updates.

@rgommers I think we are good to go here.

leofang:
Thanks to Athan for driving this and all for help!

Issue #672, opened 2023-08-07
Device-specific type promotion rules


oleksandr-pavlyk:
For devices that have different capabilities (support for float64/float16 may or may not be present), 
the type promotion graph adopted by array-API must be different.

Functions `xp.can_cast` ([ref](https://data-apis.org/array-api/latest/API_specification/generated/array_api.can_cast.html)) and `xp.result_type` ([ref](https://data-apis.org/array-api/latest/API_specification/generated/array_api.result_type.html)) allow for signatures where all inputs are data types. 

For such inputs the device information is not available, but `xp.can_cast(xp.float32, xp.float64)` is `True` for devices that support float64, but should be `False` to devices that do not.

The only variability in data type support across devices, per SYCL spec, exists for 'float64' and 'float16'. In SYCL these capabilities are queries via [device aspects](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:device-aspects). This is only relevant since standardization in SYCL gives us confidence that every other data-type is going to be supported.

So I would like to open for discussion the following proposal
  - `xp.can_cast` and `xp_result_type` need to acquire `device` keyword support with default value of `None` to mean that device capabilities are determined from argument arrays, if present, and or from capabilities of the default device otherwise.

Alternative name for the keyword is acceptable too, as long it can used to convey device capabilities.

oleksandr-pavlyk:
@rgommers Yes, if `x` is an array, the device information is available. It is the use case `can_cast(dtype, xp.float64)`, or rather `result_type(xp.int64, xp.uint64)` that I worry about. This should promote to the default floating point data type which is device specific.

kgryte:
@oleksandr-pavlyk Circling back here. Did you have further thoughts you'd like to share here or on the linked PR? 

https://github.com/data-apis/array-api/pull/691

oleksandr-pavlyk:
The issue was again discussed at a meeting, and it was decided that adding `device` keyword is going to be a burden to current libraries.

The way to query device specific type promotion graph is to use array objects instead of dtype objects. Perhaps a note in the specification for these functions may be added requiring implementations to do that, and noting that output of `xp.can_cast(array, to_)` may differ from output of `xp.can_cast(array.dtype, to_)`.

PR #691, opened 2023-09-21
Add `device` kwarg support to `can_cast` and `result_type`

kgryte:
This PR

- resolves https://github.com/data-apis/array-api/issues/672 by adding `device` keyword argument support to `can_cast` and `result_type`.
- adds guidance indicating that, by default, device capabilities should **not** be considered when applying type promotion rules. Currently, the specification is mum on whether device capabilities should be considered. For both functions, when `device` is `None`, only Array API type promotion rules may be applied. When `device` is a device object, consideration can be made as to whether particular type promotion rules are possible. E.g., if a device only supports `float32`, then
  
  ```python
  can_cast(xp.float32, xp.float64, device=<device>)
  ```

  can return `false`. Otherwise, users have to workaround in which they apply `can_cast` and then use the proposed inspection APIs (https://github.com/data-apis/array-api/pull/689) to determine whether a device supports the promoted type. This PR makes this operation more direct.

### DIFF
--- a/src/array_api_stubs/_draft/data_type_functions.py
+++ b/src/array_api_stubs/_draft/data_type_functions.py
@@ -85,7 +85,7 @@ def can_cast(
     Returns
     -------
     out: bool
-        ``True`` if the cast is possible; otherwise, ``False``.
+        ``True`` if the cast can occur; otherwise, ``False``.
     """
 
 
--- a/src/array_api_stubs/_draft/data_type_functions.py
+++ b/src/array_api_stubs/_draft/data_type_functions.py
@@ -60,7 +60,13 @@ def astype(
     """
 
 
-def can_cast(from_: Union[dtype, array], to: dtype, /) -> bool:
+def can_cast(
+    from_: Union[dtype, array],
+    to: dtype,
+    /,
+    *,
+    device: Optional[device] = None,
+) -> bool:
     """
     Determines if one data type can be cast to another data type according :ref:`type-promotion` rules.
 
@@ -70,11 +76,16 @@ def can_cast(from_: Union[dtype, array], to: dtype, /) -> bool:
         input data type or array from which to cast.
     to: dtype
         desired data type.
+    device: Optional[device]
+        device on which to perform a cast. If ``device`` is ``None``, the function must apply :ref:`type-promotion` rules irrespective of device capabilities. If ``device`` is a device object, the function must determine whether a cast can be performed on the specified device. Default: ``None``.
+
+        .. note::
+           While specification-conforming array libraries are expected to support all data types included in this specification, array libraries may support devices which do not have full data type support. Accordingly, the inclusion of a ``device`` keyword argument allows downstream array API consumers to test casting support for particular devices.
 
     Returns
     -------
     out: bool
-        ``True`` if the cast can occur according to :ref:`type-promotion` rules; otherwise, ``False``.
+        ``True`` if the cast is possible; otherwise, ``False``.
     """
 
 
@@ -206,7 +217,10 @@ def isdtype(
     """
 
 
-def result_type(*arrays_and_dtypes: Union[array, dtype]) -> dtype:
+def result_type(
+    *arrays_and_dtypes: Union[array, dtype],
+    device: Optional[device] = None,
+) -> dtype:
     """
     Returns the dtype that results from applying the type promotion rules (see :ref:`type-promotion`) to the arguments.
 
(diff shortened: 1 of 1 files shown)


rgommers (changes requested):
I'm not sure I like to add the `device=` keyword here. As discussed in gh-672, one can already make this work by passing in an array. So why not add a note instead that in case one worries about device-specific behavior, use an array on the device rather than a dtype as the first input? That avoids the need to add the keyword to these APIs in all array libraries, for most of which this will be a no-op. And it's anyway a better idea imho, because the only way to obtain a device instance is from an array, so `can_cast(x, xp.float64)` should be preferred over `can_cast(xp.float32, xp.float32, device=x.device)`. In case you want this kind of logic without having an array at hand, use `can_cast(xp.empty([], dtype=..., device=...), xp.float64)`.

kgryte:
@rgommers Consider `result_type` which can take multiple operands. Which device takes precedence in the scenario where you provide arrays belonging to different devices?

For `can_cast`, if we require that `can_cast` infer the device from `from`, how can you resolve the cast result independent of a device? E.g., what if I want to know if I can safely cast a GPU allocated array to a CPU allocated array of greater precision? I can use `astype` (given the recent PR) and catch the exception, but this still seems clunky to me. 

IMO, by default, these APIs should return consistent values--namely, by strictly applying type promotion rules as defined in the specification--for the sake of predictability. The `device` kwarg would allow an opt-in to device-specific return values.

kgryte:
Also, adding a note here would imply that returning device-specific return values was somehow the expected behavior in prior versions of the spec. I don't think that is true given the existing language.

> Determines if one data type can be cast to another data type according :ref:`type-promotion` rules.

> ...from applying the type promotion rules...

If we update the spec to infer the device from input arrays and then allow returning results only in accordance with that device, I'd consider that a breaking change to the spec.

rgommers:
> Consider `result_type` which can take multiple operands. Which device takes precedence in the scenario where you provide arrays belonging to different devices?

It can raise or be undefined behavior to have multiple arrays on different devices. We don't allow combining arrays that aren't on the same device in any other API either.

> For `can_cast`, if we require that `can_cast` infer the device from `from`, how can you resolve the cast result independent of a device? E.g., what if I want to know if I can safely cast a GPU allocated array to a CPU allocated array of greater precision? I can use `astype` (given the recent PR) and catch the exception, but this still seems clunky to me.

You can use `can_cast(x.dtype, ...)` to get the device-independent behavior?

> Also, adding a note here would imply that returning device-specific return values was somehow the expected behavior in prior versions of the spec. I don't think that is true given the existing language. ...  I'd consider that a breaking change to the spec.

I don't quite agree. It is ill-specified now, it says nothing of relevance either way in case a dtype is missing on a specific device. We didn't consider that case at all until recently. It is and remains a bit of a corner case, so we're clarifying the expectation here. It doesn't seem reasonable to me to take a long of preciseness for a corner case to extrapolate that we must add non-useful keywords to libraries like numpy.

oleksandr-pavlyk:
The type promotion rules result in a graph where dtypes are the nodes. Functions `can_cast` and `result_type` query this graph. 

Some data types may be unsupported on certain devices, changing the graph. The `device` keyword in these functions aim to select the graph to use in the query.

What remains to clarify is how to aggregate `device` information from the `device` keyword and from input arrays (which also carry device information).

1. `can_cast(from_: dtype, to_: dtype, device=None)` uses full graph (as if the device supports all data types in the specification.
2. `can_cast(from_: dtype, to_: dtype, device=<device>)` uses promotion graph associated with the specified device
3. `can_cast(from_: array, to_: dtype, device=None)` uses graph corresponding to the device of input array

1. `result_types(*dtypes, device=None)` uses full graph
2. `result_type(*arrays, device=None)` requires all arrays to have the same device, or exception is raised, uses promotion rules applicable for that device.
3. `result_type(*array_or_dtypes, device=<device>)` requires that given device be the device where `arrays` were created, or exception is raised, uses promotion rules applicable for that device.

I think the spec should also provide some rules about type promotion graphs for a proper subset of dtypes.

rgommers:
Those rules seem reasonable to me.

kgryte:
@rgommers To be clear, you originally raised objections to adding `device` kwarg support. Does  https://github.com/data-apis/array-api/pull/691#issuecomment-1800240575 sufficiently address your concerns? If so, I'll update this PR accordingly so that we can move things forward.

rgommers:
Actually, re-reading the whole discussion, I think no one answered my comments around this already being possible without adding device keywords to these functions. @oleksandr-pavlyk wrote on the linked issue:

> It is the use case `can_cast(dtype, xp.float64)`, or rather `result_type(xp.int64, xp.uint64)` that I worry about. This should promote to the default floating point data type which is device specific.  

There is no way to get a device object in a library-independent way other than taking it from an array. So this use case may not be relevant, and if it is then I'd say that it's as easy to do:
```
can_cast(xp.float32, xp.float64, x)
```
or 
```
can_cast(xp.float32, xp.float64, asarray([], device=_my_current_device))
```
as it is to do 
```
can_cast(xp.float32, xp.float64, device=_my_current_device)
```

Hence, unless I am missing something, this is still not compelling and already supported - we should reject this change I believe.

kgryte:
As discussed in https://github.com/data-apis/array-api/issues/672#issuecomment-1850016256, current consensus is to **not** add a `device` kwarg to `can_cast` and `result_type`. 

However, we will add, in a separate PR, some clarification to the text regarding how the respective functions should behave when provided input arrays, rather than dtypes, where the array device should be taken into consideration and reflect the device-specific type promotion graph.

Closing this PR...

Issue #655, opened 2023-07-13
RFC: add support for creating a new array by tiling a provided array


kgryte:
This RFC proposes adding support to the array API specification for creating a new array by tiling a provided array.

## Overview

Based on array comparison [data](https://github.com/data-apis/array-api-comparison/blob/fa017050ea8079c0fbca1142dcfd130bdcbb6136/signatures/manipulation/tile.md), the API is available in all array libraries.

## Prior art

- NumPy: <https://numpy.org/doc/stable/reference/generated/numpy.tile.html>
- PyTorch: <https://pytorch.org/docs/stable/generated/torch.tile.html>
- TensorFlow: <https://www.tensorflow.org/api_docs/python/tf/tile>
- JAX: <https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.tile.html>
- CuPy: <https://docs.cupy.dev/en/stable/reference/generated/cupy.tile.html>
- Dask: <https://docs.dask.org/en/stable/generated/dask.array.tile.html>

## Proposal

```python
def tile(x: array, repetitions: Union[int, Tuple[int, ...]], /)
```

- **repetitions**: the number of repetitions of `x` along each axis. Let `N = len(repetitions)` and let `M = len(x.shape)`.

  - if `M < N`, then `x` must be broadcasted to be `N` dimensional by prepending new axes.
  - if `M > N`, then `repetitions` must be broadcasted by prepending new axes.

## Questions

- Neither PyTorch nor TensorFlow support providing an `int` for `repetitions`. TensorFlow requires `repetitions` to be a `Tensor`. All array libraries apart from TensorFlow allow `repetitions` to be a tuple. Should `repetitions` be typed more strictly?

## Related

- Mention of xarray's need for `tile`: <https://github.com/data-apis/array-api/issues/187#issuecomment-1553615779>

rgommers:
> * Neither PyTorch nor TensorFlow support providing an `int` for `repetitions`. TensorFlow requires `repetitions` to be a `Tensor`. All array libraries apart from TensorFlow allow `repetitions` to be a tuple. Should `repetitions` be typed more strictly?

I'd support tuple-only. It can always be loosened in the future, and this function isn't used all that much hence it's not much of a burden to type `(3,)` instead of `3`.

Should `repetitions` be positional-only perhaps? I noticed that NumPy uses `reps` rather than `repetitions`.

kgryte:
[...]

PR #692, opened 2023-09-21
Add support for tiling an array to the specification

kgryte:
This PR

- resolves https://github.com/data-apis/array-api/issues/655 by adding an API for creating a new array by tiling an input array a specified number of repetitions.
- requires that the second positional argument (`repetitions`) be a tuple of `ints`. NumPy et al support providing an integer; however, PyTorch and TensorFlow do not. PyTorch requires a tuple. TensorFlow requires an array. This PR seeks to find common ground by requiring `repetitions` to be a tuple.
- supports prepend behavior, as supported by NumPy et al and PyTorch. I.e., if `len(repetitions)` is less than the rank of `x`, then ones are prepended to `repetitions`. If the rank of `x` is less than `len(repetitions)`, singleton dimensions are prepended to `x`. TensorFlow does not support this behavior, requiring that `repetitions.shape[0]` equal the rank of `x`.

### DIFF
--- a/src/array_api_stubs/_draft/manipulation_functions.py
+++ b/src/array_api_stubs/_draft/manipulation_functions.py
@@ -260,7 +260,7 @@ def stack(arrays: Union[Tuple[array, ...], List[array]], /, *, axis: int = 0) ->
 
 def tile(x: array, repetitions: Tuple[int, ...], /):
     """
-    Constructs an array by tiling a provided array.
+    Constructs an array by tiling the input array.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


src/array_api_stubs/_draft/manipulation_functions.py
```diff
@@ -257,6 +258,30 @@ def stack(arrays: Union[Tuple[array, ...], List[array]], /, *, axis: int = 0) ->
     """
 
 
+def tile(x: array, repetitions: Tuple[int, ...], /):
+    """
+    Constructs an array by tiling a provided array.
```
oleksandr-pavlyk, line 263:
```suggestion
    Constructs an array by tiling the input array.
```

kgryte:
As this PR has been discussed during Consortium meetings and no objections have been raised either here or in the original RFC (https://github.com/data-apis/array-api/issues/655), will merge. Any further changes can be addressed in follow-up PRs...

Issue #593, opened 2023-02-14
Add `copysign` function to the standard


steff456:
This RFC requests to include a new API in the array API specification for the purpose of returning a value with the magnitude of `x` but the sign of `y`.

## Overview

Based on array comparison [data](https://github.com/data-apis/array-api-comparison/blob/7ecfa9867ec43a28e914b35868c041443d31d772/data/common_apis.csv#L37), the API is available in the majority of libraries in the PyData ecosystem.

The Array API specification does not currently include a convenient method of copying the sign from one value to another value via the IEEE 754 function `copysign`. This function is commonly used in the implementations of transcendental functions involving argument reductions.

While this can be implemented in terms of `where`, doing so is overly cumbersome.

## Prior art

- C99: <https://en.cppreference.com/w/c/numeric/math/copysign>
- NumPy: <https://numpy.org/doc/stable/reference/generated/numpy.copysign.html>
- PyTorch: <https://pytorch.org/docs/stable/generated/torch.copysign.html>
- MXNet: <https://mxnet.apache.org/versions/master/api/python/docs/api/np/generated/mxnet.np.copysign.html#mxnet.np.copysign>

## Proposal:

```python
def copysign(x1: array, x2:array, /) -> array
```

cc @kgryte

rgommers:
This seems like a reasonable addition. Do TensorFlow and MXNet have alternatives or open feature requests?

kgryte:
MXNet does have a `copysign` function. TF is the odd one out and does not appear to have a `copysign` API even in its experimental NumPy namespace.

asmeurer:
I don't know if this can actually be replicated with `where` because of `-0.0`.

PR #693, opened 2023-09-21
Add support for `copysign` to the specification

kgryte:
This PR

- resolves https://github.com/data-apis/array-api/issues/593 by adding support for composing a floating-point value having the magnitude of `x1_i` and the sign of `x2_i` for each element in `x1_i`.
- chooses to follow strict IEEE 754 semantics, in which the sign bit of `NaN` may be manipulated.
- limits to only real-valued floating-point data types.
- requires that input arguments participate in type promotion.

## Questions

- Are we comfortable limiting to only floating-point data types? Libraries would be free, of course, to support other dtypes (e.g., NumPy), but this would not be considered portable according to the spec.
- Are we comfortable with embracing full IEEE 754 NaN semantics, namely signed NaNs?

### DIFF
--- a/spec/draft/API_specification/elementwise_functions.rst
+++ b/spec/draft/API_specification/elementwise_functions.rst
@@ -34,6 +34,7 @@ Objects in API
    bitwise_xor
    ceil
    conj
+   copysign
    cos
    cosh
    divide
--- a/src/array_api_stubs/_draft/elementwise_functions.py
+++ b/src/array_api_stubs/_draft/elementwise_functions.py
@@ -16,6 +16,7 @@ __all__ = [
     "bitwise_xor",
     "ceil",
     "conj",
+    "copysign",
     "cos",
     "cosh",
     "divide",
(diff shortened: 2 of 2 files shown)


rgommers:
> Are we comfortable limiting to only floating-point data types?

Yes, because real-valued floating-point dtypes is all that's in C99.

> Are we comfortable with embracing full IEEE 754 NaN semantics, namely signed NaNs?

If it's supported by all libraries, then it should be fine I think. I quickly checked NumPy, PyTorch and JAX, all three seem to support it.

seberg:
I see nothing wrong with "supporting" signed NaNs.  In practice, I doubt libraries take care with having meaningful signs.  Python `float("nan")` is always positive, `np.nan` I _suspect_ is, but I am not 100% right now (I would be happy to ensure it though).

The C `NAN` macro does *not* have a defined sign though.  If you really want to worry more about signs, you could specify the sign of `namespace.nan` (I would say its low prio until someone shouts).

kgryte:
@seberg To clarify, while `np.nan` is a double whose signbit is `0`, NumPy does support signed NaNs:

```
In [1]: np.signbit(-np.nan)
Out[1]: True

In [2]: np.signbit(np.nan)
Out[2]: False

In [3]: np.signbit(np.copysign(np.nan, -1.0))
Out[3]: True

In [4]: np.signbit(np.copysign(np.nan, 1.0))
Out[4]: False

In [5]: np.signbit(np.copysign(-np.nan, 1.0))
Out[5]: False

In [6]: np.signbit(np.copysign(np.nan, -np.nan))
Out[6]: True

In [7]: np.signbit(np.copysign(np.nan, np.nan))
Out[7]: False
```

kgryte:
As this PR has been reviewed and discussed without objections, will go ahead and merge. Any subsequent revisions can be addressed in follow-up PRs...

asmeurer:
> Python float("nan") is always positive,

I'm curious, do you know if this is guaranteed by the language, or does it just happen to be true for all implementations?

seberg:
CPython ensures this and I am pretty sure even tests for it (I has once looked at simplifying the code responsible). I.e., IIRC, we can rely on Mark Dickinson to ensure this.
I suppose other Python implementations may not, but I would guess that if you point out CPython does, you can ask them to change it (if it matters).

PR #696, opened 2023-10-09
Fix documented dtype inputs for rfft-like functions

honno:
For the real fft family of functions:

- Replace instances of "complex-valued" in documented input with "real-valued"
- Use "must" language instead of "should"—this is to follow how [`rfft`](https://data-apis.org/array-api/draft/extensions/generated/array_api.fft.rfft.html#array_api.fft.rfft) is right now, although I could see maybe that it should all be "should"?

### DIFF
--- a/src/array_api_stubs/_2022_12/fft.py
+++ b/src/array_api_stubs/_2022_12/fft.py
@@ -279,7 +279,7 @@ def irfft(
     norm: Literal["backward", "ortho", "forward"] = "backward",
 ) -> array:
     """
-    Computes the one-dimensional inverse of ``rfft`` for complex-valued input.
+    Computes the one-dimensional inverse of ``rfft`` for real-valued input.
 
     .. note::
        Applying the one-dimensional inverse discrete Fourier transform for real-valued input to the output of this function must return the original (i.e., non-transformed) input array within numerical accuracy (i.e., ``irfft(rfft(x)) == x``), provided that the transform and inverse transform are performed with the same arguments (axis and normalization mode) and consistent length.
@@ -287,7 +287,7 @@ def irfft(
     Parameters
     ----------
     x: array
-        input array. Should have a complex-valued floating-point data type.
+        input array. Must have a real-valued floating-point data type.
     n: int
         length of the transformed axis of the **output**. If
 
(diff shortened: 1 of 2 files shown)


kgryte:
These changes should likely be backported, correct?

leofang (changes requested):
This is incorrect. Rfft is for r2c transforms, and irfft for c2r.

honno:
> These changes should likely be backported, correct?

I should've remembered to do that :sweat_smile: Backported to `2022.12`.


> This is incorrect. Rfft is for r2c transforms, and irfft for c2r.

Oh okay, did only skim over #189 so I might of missed relevant discussion? I was looking at [`np.rfft`](https://numpy.org/doc/stable/reference/generated/numpy.fft.rfft.html) and [`np.irfft`](https://numpy.org/doc/stable/reference/generated/numpy.fft.irfft.html), which documents that the input is real-valued, e.g.

> Compute the one-dimensional discrete Fourier Transform for real input.

buuuut I 'spose that means it ignores the imaginary components of complex input?

leofang:
> > These changes should likely be backported, correct?
> 
Sorry for brevity @honno, I left the comment on cell in a rush...

> > This is incorrect. Rfft is for r2c transforms, and irfft for c2r.
> 
> Oh okay, did only skim over #189 so I might of missed relevant discussion? I was looking at [`np.rfft`](https://numpy.org/doc/stable/reference/generated/numpy.fft.rfft.html), which documents that the input is real-valued, e.g. `np.rfft`'s docstring starts with "Compute the one-dimensional discrete Fourier Transform for real input." But I 'spose that means **it can take complex input and just ignores the imaginary components?**

No, that's a misnomer. `rfft` means to do Fourier transform on a real input, and `irfft` means to do the inverse operation of it such that `irfft(rfft(arr)) == arr` (subject to caveats noted everywhere, array API or NumPy). Note that after irfft you need to get back a real input.

Now, repeat after me 🙂
- `rfft` refers to the real-to-complex (r2c) transform 
- `irfft` refers to the complex-to-real (c2r) transform 

This is the only way that `irfft(rfft(arr)) == arr` could make sense. As long as you memorize this, I can assure you success for writing docs, sample codes and tests. This would also help when you cross-reference low-level FFT library docs (ex: FFTW or cuFFT).

leofang:
> But I 'spose that means **it can take complex input and just ignores the imaginary components?**

Missed this one. No, this is a behavior that we purposely forbid. Ignoring the imaginary part (either silently or raising a warning) was a poor behavior in NumPy, and I pushed during the discussions to ensure array API does not pick it up.

kgryte:
@honno Have you had the chance to follow up here?

honno:
Here's a summary of input-output behaviour

| Function | Array API                 | NumPy docs | NumPy actual                                              |
|----------|---------------------------|------------|-----------------------------------------------------------|
| rfft     | Must have real input      | Real input | Also accepts complex input, always returns complex output |
| rfftn    | Must have real input      | Real input | Also accepts complex input, always returns complex output |
| irfft    | Should have complex input | Real input | Also accepts complex input, always returns real output    |
| irfftn   | Should have complex input | Real input | Also accepts complex input, always returns real output    |

I think I see what @leofang means now, so disregard this PR! :sweat_smile:

Issue #688, opened 2023-09-20
RFC: add `searchsorted` to the specification


kgryte:
This RFC proposes adding support to the array API specification for finding the indices where elements should be inserted in order to maintain order.

## Overview

Based on array comparison [data](https://github.com/data-apis/array-api-comparison/blob/29ec55b6d0a1a3fe29a6d6106a48d2d08bdf14d6/signatures/searching/searchsorted.md), the API is available across all considered libraries.

Furthermore, all considered libraries support the `side` keyword argument, and all considered libraries, except for TensorFlow, support the `sorter` keyword argument.

JAX supports an additional kwarg, `method`, which is used based on device/size performance optimization considerations. PyTorch and TensorFlow support specifying the output data type, but differ in naming conventions.

All array libraries support one-dimensional arrays. PyTorch and TensorFlow generalize to any n-dimensional ndarray (stacking).

## Prior Art

- NumPy: <https://numpy.org/doc/stable/reference/generated/numpy.searchsorted.html>
- PyTorch: <https://pytorch.org/docs/stable/generated/torch.searchsorted.html>
- TensorFlow: <https://www.tensorflow.org/api_docs/python/tf/searchsorted>
- JAX: <https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.searchsorted.html>
- CuPy: <https://docs.cupy.dev/en/stable/reference/generated/cupy.searchsorted.html>
- Dask: <https://docs.dask.org/en/stable/generated/dask.array.searchsorted.html>

## Proposal

```python
def searchsorted(x1: array, x2: array, /, *, side="left", sorter=None)
```

- **x1**: one-dimensional array. If `sorter` is `None`, `x1` must be sorted in ascending order.
- **x2**: one-dimensional array.
- **side**:  if "left", the returned index `i` satisfies `x1[i-1] < x2[j] <= x1[i]`. Otherwise, if "right", the returned index `i` satisfies `x1[i-1] <= x2[j] < x1[i]`. If no suitable index, then `i` is either `0` or `N`, respectively, were `N` is the length of `x1`.
- **sorter**: array of integer indices that sort `x1` in ascending order (e.g., as might be produced via `argsort`).

## Questions

- Should the API be extended to support stacking as in PyTorch/TensorFlow?
- Should the API support a scalar value for `x2`? NumPy, PyTorch, JAX, Dask support scalars. CuPy and TensorFlow do not.

rgommers:
[...]

PR #699, opened 2023-11-06
Add `searchsorted` to the specification

kgryte:
This PR

- resolves https://github.com/data-apis/array-api/issues/688 by adding a specification for `searchsorted` to the Array API specification.
- requires that `x1` be a one-dimensional array. PyTorch and TensorFlow allow `x1` to be multi-dimensional; however, NumPy et al do not. In this PR, I've chosen to restrict `x1` to a one-dimensional array, as this is the lowest common denominator. The specification can be extended in the future to accommodate multi-dimensional `x1` (stacking) in the future.
- requires that `x2` be an array. CuPy and TensorFlow require that `x2` be an array, while NumPy, PyTorch, and others allow `x2` to be a scalar. In this PR, I've chosen to restrict `x2` to be an array. For the scalar use case, downstream consumers can wrap scalar values as zero-dimensional arrays.

### DIFF
--- a/src/array_api_stubs/_draft/searching_functions.py
+++ b/src/array_api_stubs/_draft/searching_functions.py
@@ -131,7 +131,7 @@ def searchsorted(
     Notes
     -----
 
-    For real-valued floating-point arrays, the sort order of NaNs and signed zeros is unspecified and thus implementation-dependent. Accordingly, when a real-valued floating-point array contains NaNs and signed zeros, what constitutes ascending order may vary among specification-conforming array libraries.
+    For real-valued floating-point arrays, the sort order of NaNs and signed zeros is unspecified and thus implementation-dependent. Accordingly, when a real-valued floating-point array contains NaNs and signed zeros, what constitutes ascending order may vary among specification-conforming array libraries and behavior is implementation-dependent.
 
     Specification-conforming libraries should, however, ensure consistency with ``sort`` and ``argsort`` (i.e., if a value in ``x2`` is inserted into ``x1`` according to the corresponding index in the output array and ``sort`` is invoked on the resultant array, the sorted result should be an array in the same order).
     """
--- a/src/array_api_stubs/_draft/searching_functions.py
+++ b/src/array_api_stubs/_draft/searching_functions.py
@@ -1,7 +1,7 @@
 __all__ = ["argmax", "argmin", "nonzero", "where"]
 
 
-from ._types import Optional, Tuple, array
+from ._types import Optional, Tuple, Literal, array
 
 
 def argmax(x: array, /, *, axis: Optional[int] = None, keepdims: bool = False) -> array:
--- a/src/array_api_stubs/_draft/searching_functions.py
+++ b/src/array_api_stubs/_draft/searching_functions.py
@@ -105,7 +105,21 @@ def searchsorted(
     x2: array
         array containing search values.
     side: Literal['left', 'right']
-        if ``'left'``, then each returned index ``i`` must satisfy ``x1[i-1] < x2[m][n]...[j] <= x1[i]``, and, if, for an element ``x2[m][n]...[j]``, no index satisfies the index condition, then the returned index for that element must be ``0``. Otherwise, if ``'right'``, then each returned index ``i`` must satisfy ``x1[i-1] <= x2[m][n]...[j] < x1[i]``, and, if, for an element ``x2[m][n]...[j]``, no index satisfies the index condition, then the returned index for that element must be ``N``, where ``N`` is the number of elements in ``x1``. Default: ``'left'``.
+        argument controlling which index is returned if a value lands exactly on an edge.
+
+        Let ``x`` be an array of rank ``N`` where ``v`` is an individual element given by ``v = x2[n,m,...,j]``.
+
+        If ``side == 'left'``, then
+
+        - each returned index ``i`` must satisfy the index condition ``x1[i-1] < v <= x1[i]``.
+        - if no index satisfies the index condition, then the returned index for that element must be ``0``.
+
+        Otherwise, if ``side == 'right'``, then
+
+        - each returned index ``i`` must satisfy the index condition ``x1[i-1] <= v < x1[i]``.
+        - if no index satisfies the index condition, then the returned index for that element must be ``N``, where ``N`` is the number of elements in ``x1``.
+
+        Default: ``'left'``.
     sorter: Optional[array]
         array of indices that sort ``x1`` in ascending order. The array must have the same shape as ``x1`` and have an integer data type. Default: ``None``.
 
--- a/src/array_api_stubs/_draft/searching_functions.py
+++ b/src/array_api_stubs/_draft/searching_functions.py
@@ -101,9 +101,9 @@ def searchsorted(
     Parameters
     ----------
     x1: array
-        input array. Must be a one-dimensional array. If ``sorter`` is ``None``, must be sorted in ascending order; otherwise, ``sorter`` must be an array of indices that sort ``x1`` in ascending order.
+        input array. Must be a one-dimensional array. Should have a real-valued data type. If ``sorter`` is ``None``, must be sorted in ascending order; otherwise, ``sorter`` must be an array of indices that sort ``x1`` in ascending order.
     x2: array
-        array containing search values.
+        array containing search values. Should have a real-valued data type.
     side: Literal['left', 'right']
         argument controlling which index is returned if a value lands exactly on an edge.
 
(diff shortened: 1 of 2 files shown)


src/array_api_stubs/_draft/searching_functions.py
```diff
@@ -87,6 +87,35 @@ def nonzero(x: array, /) -> Tuple[array, ...]:
...
+) -> array:
+    """
+    Finds the indices into ``x1`` such that, if the corresponding elements in ``x2`` were inserted before the indices, the order of ``x1``, when sorted in ascending order, would be preserved.
+
+    Parameters
+    ----------
+    x1: array
+        input array. Must be a one-dimensional array. If ``sorter`` is ``None``, must be sorted in ascending order; otherwise, ``sorter`` must be an array of indices that sort ``x1`` in ascending order.
+    x2: array
+        array containing search values.
```
rgommers, line 106: Any reason why this doesn't say that `x2` must be a one-dimensional array?

src/array_api_stubs/_draft/searching_functions.py
```diff
@@ -87,6 +87,35 @@ def nonzero(x: array, /) -> Tuple[array, ...]:
...
+    side: Literal['left', 'right']
+        if ``'left'``, then each returned index ``i`` must satisfy ``x1[i-1] < x2[m][n]...[j] <= x1[i]``, and, if, for an element ``x2[m][n]...[j]``, no index satisfies the index condition, then the returned index for that element must be ``0``. Otherwise, if ``'right'``, then each returned index ``i`` must satisfy ``x1[i-1] <= x2[m][n]...[j] < x1[i]``, and, if, for an element ``x2[m][n]...[j]``, no index satisfies the index condition, then the returned index for that element must be ``N``, where ``N`` is the number of elements in ``x1``. Default: ``'left'``.
```
rgommers, line 108: The notation `[m][n]...[j]` is confusing to me, it's not normal indexing notation. Please consider replacing it with something clearer, like what the `numpy.searchsorted` docstring has:
```
======  ============================
`side`  returned index `i` satisfies
======  ============================
left    ``a[i-1] < v <= a[i]``
right   ``a[i-1] <= v < a[i]``
======  ============================
```

src/array_api_stubs/_draft/searching_functions.py
```diff
@@ -87,6 +87,35 @@ def nonzero(x: array, /) -> Tuple[array, ...]:
...
+    sorter: Optional[array]
+        array of indices that sort ``x1`` in ascending order. The array must have the same shape as ``x1`` and have an integer data type. Default: ``None``.
+
+    Returns
+    -------
+    out: array
+        an array of indices with the same shape as ``x2``. The returned array must have the default array index data type.
+    """
```
rgommers, line 116: Can you add `.. versionadded:: 2023.12` in a Notes section?

rgommers (commented):
Thanks @kgryte. I have a few minor comments, but +1 overall to this addition.

The issue is linked to from a scikit-learn PR where `searchsorted` was needed, and SciPy needs it as well (Cc @mdhaber for visibility).

src/array_api_stubs/_draft/searching_functions.py
kgryte: These annotations will be added at the end, prior to specification finalization, as we did last year. We have to do this for multiple APIs and can do in one pass.

src/array_api_stubs/_draft/searching_functions.py
kgryte: I followed the PyTorch [docs](https://pytorch.org/docs/stable/generated/torch.searchsorted.html) on using nested bracket notation. I can rework to something else.

src/array_api_stubs/_draft/searching_functions.py
kgryte: Because `x2` is [allowed](https://numpy.org/doc/stable/reference/generated/numpy.searchsorted.html) to be multi-dimensional. Only `x1` is required to be one-dimensional. The output array shape should be the same as `x2`. I don't see a reason why we'd need to constrain `x2` to be one-dimensional.

rgommers: Okay, the proposal in gh-688 said 1-D, but this seems right. I checked that all of NumPy, CuPy, JAX, PyTorch and Dask support n-D arrays here - resolving.

src/array_api_stubs/_draft/searching_functions.py
```diff
@@ -87,6 +87,35 @@ def nonzero(x: array, /) -> Tuple[array, ...]:
...
+
```
emcastillo, line 117: I am wondering if its worth it to add a note regarding the treatment of `NaN` values?

https://numpy.org/doc/stable/reference/generated/numpy.searchsorted.html
NumPy has two sentences about it. Unfortunately in CuPy we don't have it documented though...

kgryte: @emcastillo `NaN` sort order is discussed in the specification: https://data-apis.org/array-api/2022.12/API_specification/sorting_functions.html.

As in other functions, such as `sort` and `argsort`, the sort order of NaNs and signed zeros is implementation-dependent. In which case, I don't think we have anything normative to add here.

However, I will add a note indicating that order is likely to vary between implementations.

kgryte: Following-up on this, NumPy sorts `NaN` values to the end. However, the specification leaves the sort order of `NaN` values as unspecified. In which case, the following NumPy array would be a valid sort according to the specification:

```
In [1]: x = np.asarray([1.0, 2.0, np.nan, 3.0, 4.0, np.nan, 5.0])

In [2]: x
Out[2]: array([ 1.,  2., nan,  3.,  4., nan,  5.])
```

When invoking `np.searchsorted` on this array, NumPy always sorts search values to indices before `NaN` elements.

```
In [3]: np.searchsorted(x, np.asarray([0.5, 1.5, 2.5, 3.5, 4.5, 5.5]))
Out[3]: array([0, 1, 2, 4, 5, 7])

In [4]: np.searchsorted(x, np.asarray([0.5, 1.5, 2.5, 3.5, 4.5, 5.5]), side="left")
Out[4]: array([0, 1, 2, 4, 5, 7])

In [5]: np.searchsorted(x, np.asarray([0.5, 1.5, 2.5, 3.5, 4.5, 5.5]), side="right")
Out[5]: array([0, 1, 2, 4, 5, 7])
```

This behavior does violate the index condition; namely, that each returned index must satisfy `x1[i-1] < v <= x1[i]`, as `v < NaN` is always `False`.

I'll raise this concern in the upcoming workgroup meeting.

src/array_api_stubs/_draft/searching_functions.py
kgryte: This should now be resolved.

src/array_api_stubs/_draft/searching_functions.py
rgommers: Is there anything you still want to change here @kgryte?

kgryte: I added a note concerning real-valued floating-point sort order as relates to `NaNs` and signed zeros. Namely, as sort order is implementation-dependent, by implication, so too is `searchsorted` behavior. However, returned indices should be consistent with `sort` and `argsort`.

kgryte:
@rgommers This PR should be ready for final sign off.

rgommers (approved):
Reviewed again, all comments were addressed and this PR looks good to me. I'll plan to merge it later today or tomorrow.

asmeurer:
>  In this PR, I've chosen to restrict x1 to a one-dimensional array, as this is the lowest common denominator. The specification can be extended in the future to accommodate multi-dimensional x1 (stacking) in the future.

Shouldn't the language be "x1 **should** be a one-dimensional array" instead of "x1 **must** be a one-dimensional array"?

src/array_api_stubs/_draft/searching_functions.py
```diff
@@ -87,6 +87,56 @@ def nonzero(x: array, /) -> Tuple[array, ...]:
...
+        - if no index satisfies the index condition, then the returned index for that element must be ``0``.
+
+        Otherwise, if ``side == 'right'``, then
+
+        - each returned index ``i`` must satisfy the index condition ``x1[i-1] <= v < x1[i]``.
+        - if no index satisfies the index condition, then the returned index for that element must be ``N``, where ``N`` is the number of elements in ``x1``.
+
+        Default: ``'left'``.
+    sorter: Optional[array]
+        array of indices that sort ``x1`` in ascending order. The array must have the same shape as ``x1`` and have an integer data type. Default: ``None``.
```
asmeurer, line 124: Should we only require this to be the default array index type?

kgryte: Why would that be necessary?

Issue #588, opened 2023-02-03
unique_all doesn't specify that indices should be a flattened index


asmeurer:
[unique_all](https://data-apis.org/array-api/latest/API_specification/generated/array_api.unique_all.html#array_api.unique_all):

> second element must have the field name indices and must be an array containing the indices (first occurrences) of x that result in values. The array must have the same shape as values and must have the default array index data type.

This is under specified because it isn't clear that the indices are for a *flattened* `x`, e.g.:

```py
>>> import numpy.array_api as xp
<stdin>:1: UserWarning: The numpy.array_api submodule is still experimental. See NEP 47.
>>> x = xp.asarray([[1, 0, 1], [0, 2, 1]])
>>> xp.unique_all(x)
UniqueAllResult(values=Array([0, 1, 2], dtype=int64), indices=Array([1, 0, 4], dtype=int64), inverse_indices=Array([[1, 0, 1],
       [0, 2, 1]], dtype=int64), counts=Array([2, 3, 1], dtype=int64))
```

PR #700, opened 2023-11-06
Clarify flattening behavior and fix required output shape for inverse indices in `unique_*` APIs

kgryte:
This PR

- resolves https://github.com/data-apis/array-api/issues/588 by clarifying flattening behavior and fixing the required output shape for returned inverse indices.
- makes explicit in the draft specification that flattening should be in row-major, C-style order. This is not backported to prior revisions. While this flattening behavior may have been assumed at the time of writing the specifications, we shouldn't be post-writing this assumption into prior revisions. Instead, we clarify expected behavior going forward. A strict reading of the prior revisions by downstream consumers should entail no guarantees regarding flattening order.
- makes explicit in the draft specification that the count order should match the order in `values` (i.e., the returned array of unique values). This was not made explicit in prior revisions, and its omission might indicate that `counts` could be returned in any order. While unlikely, we prohibit arbitrary count order for subsequent revisions.
- backports the following changes to the 2021 and 2022 revisions of the Array API standard:
    - correction to the required output shape of `inverse_indices`. Prior revisions stated that this should have the same shape as `x`, but this is only true if `x` is one-dimensional. For multi-dimensional `x`, the output shape should the shape of a flattened `x`.
    - clarifications regarding the contents of returned indices (first occurrences). Namely, that the returned array should include indices into a flattened `x`. For the case where `x` is one-dimensional, the guidance is the same. For multi-dimensional `x`, this clarifies that returned indices are for the flattened `x` and not the original input array.

### DIFF
--- a/src/array_api_stubs/_2021_12/set_functions.py
+++ b/src/array_api_stubs/_2021_12/set_functions.py
@@ -32,8 +32,8 @@ def unique_all(x: array, /) -> Tuple[array, array, array, array]:
         a namedtuple ``(values, indices, inverse_indices, counts)`` whose
 
         - first element must have the field name ``values`` and must be an array containing the unique elements of ``x``. The array must have the same data type as ``x``.
-        - second element must have the field name ``indices`` and must be an array containing the indices (first occurrences) of ``x`` that result in ``values``. The array must have the same shape as ``values`` and must have the default array index data type.
-        - third element must have the field name ``inverse_indices`` and must be an array containing the indices of ``values`` that reconstruct ``x``. The array must have the same shape as ``x`` and must have the default array index data type.
+        - second element must have the field name ``indices`` and must be an array containing the indices (first occurrences) of a flattened ``x`` that result in ``values``. The array must have the same shape as ``values`` and must have the default array index data type.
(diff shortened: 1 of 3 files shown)


kgryte:
@asmeurer Would you mind reviewing since you submitted the original issue? Thanks!

rgommers:
Eh, now seeing this right after NumPy changed it - consensus in the discussion that flattening was considered a bug, see https://github.com/numpy/numpy/pull/25553. I didn't really follow that PR, so let me ping @jakevdp here for his opinion on what to do with this.

jakevdp:
The goal of that PR was to make `np.unique_all` and `np.unique_inverse` compatible with the `2022.12` array API. After I sent the PR, reviewers suggested we change the output shape in `np.unique` as well.

For the array API, I think the current specification (i.e. not flattening `inverse_indices`) makes the most sense, because then you can always reconstruct the input using `unique_values[inverse_indices]`.

leofang (changes requested):
(Mark it in red to avoid accidental merge.) So should we just close this PR now that NumPy devs also agreed we don't need flattening?

leofang:
Q: @kgryte it seems we can just close this PR?

kgryte:
@leofang Yes, I will close and move the clarifications regarding flattening behavior to a new PR.

PR #701, opened 2023-11-06
Add note clarifying behavior for out-of-bounds indices in `take`

kgryte:
This PR

- addresses a comment made concerning out-of-bounds index behavior in `take` (see https://github.com/data-apis/array-api/issues/177#issuecomment-1513903987). Namely, a note is added to indicate that behavior for out-of-bounds indices is left unspecified. This is consistent with current [guidance](https://data-apis.org/array-api/2022.12/API_specification/indexing.html#single-axis-indexing) for integer indexing.


kgryte:
@asmeurer As you are the original comment author, would you mind reviewing and signing off? Thanks!

kgryte:
As this PR is a straightforward textual change which matches the language in the indexing specification, will go ahead and merge.

@asmeurer If you have further comments, please feel free to open a new issue and we can address in a separate follow-up PR.

Issue #479, opened 2022-09-20
Array objects of arbitrary rank are infeasible - require a reasonable range of ranks instead


Zac-HD:
https://github.com/data-apis/array-api/blob/main/spec/API_specification/array_object.rst#17 says:

> Furthermore, a conforming implementation of the array API standard must support array objects of arbitrary rank `N` (i.e., number of dimensions), where `N` is greater than or equal to zero.

Unfortunately this is infeasible: on any computer there's going to be a finite maximum dimensionality, and in practice it's usually much smaller - e.g. `numpy.array_api` is limited to 32-dimensional arrays.  When it's impossible to comply with the standard, implementations will quite reasonably choose whatever noncompliant option makes sense for them.

I'd therefore propose that this should instead require implementations to support arrays with between zero and 32 dimensions inclusive (or another reasonable constant TBD), and permit implementations to support a larger number of dimensions if they wish.

tylerjereddy:
From the `pykokkos` side, I asked @crtrott on Kokkos team, since we ultimately depend on the compiled C++ Kokkos bindings under the hood.

> Kokkos right now supports rank 8

> Mdspan supports arbitrary

> In practice your compiler will hit recursion limits at between rank 100 and 200 or so

> But mdspan is the multidimensional array which comes the closest to truly arbitrary as far as i know

rgommers:
That seems like a reasonable change to make. I think the absolute minimum number of dimensions that are needed to support the features used in the standard (e.g., linear algebra routines with a batch dimension) is 3. In practice, >3 are used though, so it'd be good to set the number higher. 

Is there any library that does not support 8 dimensions?

leofang:
8 is way too small for einsum. We've hit several issues with cuQuantum in which NumPy/CuPy only support up to 32 ndim. cuTENSOR has supported up to 64 and PyTorch has no limitation.

kgryte:
@Zac-HD What is motivating this request? Am I correct that this stems from a concern in Hypothesis?

Without further context, I am leery about imposing any explicit limits on array rank in the specification, as does not make sense to me to arbitrarily impose constraints based on present practice, when we don't have a crystal ball for the future.

[...]

Issue #694, opened 2023-09-25
RFC: add inspection support for querying the maximum number of supported dimensions


kgryte:
In https://github.com/data-apis/array-api/issues/640, we proposed adding a unified inspection API namespace, and, in that proposal, we included support for querying array library capabilities.

That RFC includes two capabilities:

- **data_dependent_shapes**
- **boolean_indexing**

In this RFC, I'd like to propose a third; namely, the ability to query the maximum number of supported dimensions. This RFC is related to https://github.com/data-apis/array-api/issues/479, where it was requested that the specification provide better guidance regarding the minimum upper bound array rank. During Consortium working group meetings, a minimum upper bound of `4` has been discussed.

As such, different array libraries are likely to support different upper bounds, and downstream libraries should have an ergonomic means for determine what that upper bound is in order to ensure portable code. This RFC proposes that we add another field to the dictionary returned by `info.capabilities()`: **max_rank**.

As currently written, https://github.com/data-apis/array-api/issues/640 requires dictionary values be booleans. This RFC would require that requirement to be relaxed, such that dictionary values may, in general, be any value, but specific dictionary keys must be assigned values of a particular type. In this case, **max_rank** should have an integer value.

rgommers:
Thanks @kgryte. Seems reasonable to me to add this.

> In this case, **max_rank** should have an integer value.

It is not unlikely that there is no such value for some libraries, it may be infinite. If you can instantiate an empty array with a fixed memory cost rather than it increasing with the number of dimensions, there's no reason that there must be a max rank. So allow `None`?

leofang:
IIRC PyTorch allows it to be infinite (through broadcast). But why not just use library-specific `inf` if it exists?

seberg:
Allowing `float('inf')` sounds reasonable, although I doubt there is any lib at this time with truly no limit, it might just be somthing very large like `2**31` or even up to `2**64` (which is bigger than anything plausible anyway).

rgommers:
It doesn't matter too much, but `Optional[int]` communicates the intent better and seems a little nicer static typing-wise than `Union[int, float]` for something that inherently is an integer if it exists.

leofang:
Sounds good to me. I support this RFC.

PR #702, opened 2023-11-06
Revise guidance to require a minimum upper bound for supported ranks

kgryte:
This PR

- resolves https://github.com/data-apis/array-api/issues/479 by revising the guidance requiring conforming implementations to support arrays of arbitrary rank and specifying that array libraries must support, at minimum, arrays of rank `4`. Additionally, array libraries will be required to explicitly document their maximum supported rank.

## Related

- <https://github.com/data-apis/array-api/issues/694>

### DIFF
--- a/spec/draft/API_specification/array_object.rst
+++ b/spec/draft/API_specification/array_object.rst
@@ -7,7 +7,7 @@ Array object
 
 A conforming implementation of the array API standard must provide and support an array object having the following attributes and methods.
 
-Furthermore, a conforming implementation of the array API standard must support array objects of arbitrary rank ``N`` (i.e., number of dimensions), where ``N`` is greater than or equal to zero.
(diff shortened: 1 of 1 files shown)


kgryte:
cc @leofang @Zac-HD

kgryte:
@Zac-HD Indeed, that is part of the rationale behind https://github.com/data-apis/array-api/issues/694. Once we merge the inspection API in, we'll add support for querying the maximum number of allowed dimensions.

rgommers (approved):
This LGTM, thanks @kgryte. I believe this aligns with previous discussions, and won't be a problem for any implementation (while `4` was the default maxdims supported by an implementation based on Kokkos IIRC).

rgommers:
This is pretty straightforward. I'll aim to merge this in a few days unless an objection appears.

leofang (approved):

kgryte:
As this PR has received multiple approvals and no objections, will go ahead and merge. Any further changes can be addressed in follow-up PRs...

PR #703, opened 2023-11-06
Add `hypot` specification for computing the square root of the sum of squares

kgryte:
This PR

- resolves https://github.com/data-apis/array-api/issues/544 by adding a specification for computing the square root of the sum of squares.
- follows IEEE 754 guidance that, if one of the operands is `+-infinity`, the result is infinity, regardless of whether the other operand is `NaN`. Intentionally, this deviates from a naive `sqrt(x**2 + y**2)` implementation.
- only allows underflow when both arguments are subnormal and the correct result is also subnormal. This follows C99 and the POSIX standard and effectively prohibits a naive implementation. The entire purpose of this API is to avoid underflow/overflow during intermediate steps of computation and allowing a naive implementation defeats that purpose.
- follows `sqrt` in requiring floating-point data types. The guidance uses "should", not "must", to allow implementations to support non-floating-point data types. The reason for the floating-point restriction is that the output data type must be a real-valued floating-point data type and promoting from, e.g., an integral to a floating-point data type is implementation-defined. For defined promotion semantics, a consumer should provide input arrays having floating-point data types.

### DIFF
--- a/spec/draft/API_specification/elementwise_functions.rst
+++ b/spec/draft/API_specification/elementwise_functions.rst
@@ -45,6 +45,7 @@ Objects in API
    floor_divide
    greater
    greater_equal
+   hypot
    imag
    isfinite
    isinf
--- a/src/array_api_stubs/_draft/elementwise_functions.py
+++ b/src/array_api_stubs/_draft/elementwise_functions.py
@@ -1354,6 +1354,8 @@ def hypot(x1: array, x2: array, /) -> array:
     Notes
     -----
 
+    The purpose of this function is to avoid underflow and overflow during intermediate stages of computation. Accordingly, conforming implementations should not use naive implementations.
+
     **Special Cases**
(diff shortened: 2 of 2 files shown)


leofang:
I am a bit concerned about mentioning subnormal here. In CuPy we always set the compiler flag `-ftz=true` which would flush all denormals to zero. This helps significantly with floating point operations on GPU. Would this PR imply that we can no longer do this?

kgryte:
@leofang I've added a note concerning subnormal behavior and the fact that actual hardware support may vary. Does this allay your concerns?

leofang (approved):

oleksandr-pavlyk (approved):
:+1:

rgommers:
Great, looks like everyone is happy now - let's hit the green button. Thanks @kgryte & reviewers!

Issue #670, opened 2023-08-01
Need for signbit in the array API spec


oleksandr-pavlyk:
The array specification for `atan2` requires implementations to output `-0.0` for some special values, e.g.

> If `x1_i` is `-0` and x2_i is `+0`, the result is `-0`.

which brings up a question on how implementors can distinguish between `-0.0` and `+0.0` in their test suite.

Equality testing spec [mandates](https://data-apis.org/array-api/latest/API_specification/generated/array_api.equal.html) that these values be indistinguishable, and for a good reason.

The `sign` [function](https://data-apis.org/array-api/latest/API_specification/generated/array_api.sign.html) also clamps these two values into `+0.0`.

I think this calls for the standard to have the `signbit` function:
  - `numpy.signbit` [ref](https://numpy.org/doc/stable/reference/generated/numpy.signbit.html)
  - `cupy.signbit` [ref](https://docs.cupy.dev/en/stable/reference/generated/cupy.signbit.html)
  - `torch.signbit` [ref](https://pytorch.org/docs/stable/generated/torch.signbit.html)
  - `tf.experimental.numpy.signbit` [ref](https://www.tensorflow.org/api_docs/python/tf/experimental/numpy/signbit)
  - `dask.array.signbit` [ref](https://docs.dask.org/en/stable/generated/dask.array.signbit.html)
  - JAX does not seem to have `signbit` function, although it is implementable ([example](https://github.com/google/jax/issues/10621) in one of JAX's issues)

honno:
Yeah it is a bit clunky right now. In `array-api-tests` we cast 0d array floats to builtin floats, before using `math.copysign(1, your_builtin_float) == 1` to see whether its positively signed or not. Maybe it's not too big of a deal to add a function for it (or possibly modifying `sign`), don't have a sense for that.

asmeurer:
There is a proposal for copysign at https://github.com/data-apis/array-api/issues/593.

rgommers:
> which brings up a question on how implementors can distinguish between `-0.0` and `+0.0` in their test suite.

For an implementer to test something, it isn't required to use only functions in the standard. This is checking the value of a bit, which can be done with a private function even.

Not saying that it's not a good idea to have this function - but especially given that JAX doesn't have it, it'd be nice to have a stronger reason.

I checked the SciPy code base. There's a lot of use of the C/C++ versions of `signbit` in compiled code, but only a couple in Python code:
```
scipy/special/tests/test_wrightomega.py
[...]

PR #705, opened 2023-11-15
Add `signbit` specification for determining whether a sign bit is set

kgryte:
This PR

- resolves https://github.com/data-apis/array-api/issues/670 by adding a specification for `signbit` for determining whether a sign bit is set for each floating-point number in a provided input array.
- follows Python conventions in which a boolean is returned. In [C99](https://en.cppreference.com/w/c/numeric/math/signbit), this is a macro which returns an integral value, where a positive result can be any non-zero integral value and a negative result must be zero.
- limits the input data type to real-valued floating-point data types. The current text uses "should", indicating that array libraries may extend support to additional data types (e.g., signed integer data types) for, e.g., backward compatibility reasons. This PR does not use the more liberal real-valued data type, as the intent of this function is for distinguishing `+0` and `-0` and the determining the sign of `NaN`, values which are not possible outside of floating-point data types. For complex data types, users need to provide the real and imaginary components separately.
- as this [API](https://github.com/data-apis/array-api-comparison/blob/20880a575994c2cc5d9c0b2be5a5d4c3b4f73544/signatures/mathematical-functions-elementwise/signbit.md) is commonly implemented across array libraries without deviation, there are no open questions requiring resolution before specification inclusion.

### DIFF
--- a/spec/draft/API_specification/elementwise_functions.rst
+++ b/spec/draft/API_specification/elementwise_functions.rst
@@ -69,6 +69,7 @@ Objects in API
    remainder
    round
    sign
+   signbit
    sin
    sinh
    square
--- a/src/array_api_stubs/_draft/elementwise_functions.py
+++ b/src/array_api_stubs/_draft/elementwise_functions.py
@@ -51,6 +51,7 @@ __all__ = [
     "remainder",
     "round",
     "sign",
+    "signbit",
     "sin",
     "sinh",
     "square",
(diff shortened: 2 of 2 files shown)


rgommers (approved):
LGTM overall, thanks @kgryte. One sentence I'd add to the returns section is that for negative values the result is expected to be `True`, and for positive values `False`. That won't be completely obvious otherwise to readers not familiar with the low-level details.

leofang (approved):
I assume input shape-preserving is an implicit assumption, given that `signbit` just follows other elementwise docs?

kgryte:
> One sentence I'd add to the returns section is that for negative values the result is expected to be True, and for positive values False. That won't be completely obvious otherwise to readers not familiar with the low-level details.

@rgommers Isn't that already covered in the special cases section?

```
    - If ``x_i`` is a positive (i.e., greater than ``0``) finite number, the result is ``False``.
    - If ``x_i`` is a negative (i.e., less than ``0``) finite number, the result is ``True``.
```

Ref: https://github.com/data-apis/array-api/pull/705/files#diff-153d068dcea18980d39e9465019663509adce3ee4842a48738b37a2a64c693f5R2244-R2245

kgryte:
> I assume input shape-preserving is an implicit assumption, given that signbit just follows other elementwise docs?

@leofang That is correct. We don't explicitly mention the expected output shape across the element-wise specifications, but we should probably be explicit. I can submit a follow-up PR to address this more broadly.

rgommers:
> Isn't that already covered in the special cases section?

No, special cases are special for a reason, and should not be needed to deduce the main behavior of a function. The text always has to make sense if you'd simply delete the special cases section.

kgryte:
> The text always has to make sense if you'd simply delete the special cases section.

IMO this is splitting hairs. The main behavior is precisely as described, as the API should not be confused with the [signum](https://data-apis.org/array-api/2022.12/API_specification/generated/array_api.sign.html) function. Meaning, using `signbit` to test less than or greater than zero is likely an anti-pattern; one can use one of `sign`, `less`, or `greater` instead. Based on a survey of docs,

- [PyTorch](https://pytorch.org/docs/stable/generated/torch.signbit.html) does not include the requested language.
- [NumPy](https://numpy.org/doc/stable/reference/generated/numpy.signbit.html) uses misleading language. A value does not have to be less than zero in order for the sign bit to be set.
- [CuPy](https://docs.cupy.dev/en/stable/reference/generated/cupy.signbit.html) copies over the same misleading description from NumPy.
- [Dask](https://docs.dask.org/en/stable/generated/dask.array.signbit.html) copies over the same misleading description from NumPy.
- [JAX](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.signbit.html) copies over the same misleading description from NumPy.

rgommers:
It is not a big ask to add something like: "if the sign bit of a value is set, i.e. has value 1 or the value has a minus sign, the result will be False". Or add the math expression like the `sign` function (`x / |x|`) for clarity. I didn't ask for fun, I needed to actually double check when I first read it.

PR #706, opened 2023-11-15
Fix typo in input data type for `conj` specification

kgryte:
This PR

- fixes a typo for the input data type in the `conj` specification. Currently, the hyphen is misplaced, and this PR corrects that mistake. This mistake was identified in https://github.com/data-apis/array-api-tests/pull/208#discussion_r1394205425, requiring an undesired workaround in the test suite which attempts to use pattern matching for deriving test strategies.
- backports the typo correction to the `2022.12` specification release in which `conj` was added to the specification.


rgommers (approved):
Thanks @kgryte

Issue #711, opened 2023-11-25
Update 'Purpose and Scope' for complex dtype support


lucascolley:
'Purpose and Scope' still states that complex dtypes will be included in a future version. 

> 4. Standardization of these dtypes is out of scope: bfloat16, complex, extended
   precision floating point, datetime, string, object and void dtypes.

   _Rationale: these dtypes aren't uniformly supported, and their inclusion at
   this point in time could put a significant implementation burden on
   libraries. It is expected that some of these dtypes - in particular
   `bfloat16`, `complex64`, and `complex128` - will be included in a future
   version of the standard._

PR #714, opened 2023-11-30
Remove outdated content from Purpose and Scope document

kgryte:
This PR

- resolves https://github.com/data-apis/array-api/issues/711 by removing outdated content from the Purpose and Scope document. Namely, the document currently states that complex dtype support will be included in a future revision of the specification; however, complex dtype support was added in the 2022 revision.
- backports the changes to the 2022 revision of the Array API Standard.


rgommers (approved):
LGTM, thanks @kgryte

Issue #718, opened 2023-12-06
Specify output shapes for fft functions


asmeurer:
The fft functions do not specify what the output axis size should be. This is not necessarily always the same as the input axis size (or `n`). 

For example, NumPy's `ihfft` uses an output shape of `n//2 + 1` (https://numpy.org/doc/stable/reference/generated/numpy.fft.ihfft.html). This size is not mentioned anywhere in the spec.

leofang:
Yes, I think it'd be nice to mention this for `rfft`, `rfftn`, and `ihfft`. Usually, the `n` or `s` input argument defines the length of the output along the transformed axis (and other axes have the same lengths), but for the 3 APIs doing R2C transform, `n`/`s` is for the input not output.

Issue #159, opened 2021-04-07
Summary of FFT APIs


steff456:
This issue gathers all the information of the current APIs in NumPy, CuPy, Dask, JAX, MXNet, pytorch and tensorflow in the FFT module. The main outcome of this issue is to open the discussion of the potential APIs that are going to be included in the spec.

The current APIs involved in this issue are:
- Standard FFT: fft, ifft, fft2, ifft2, fftn, ifftn
- Real FFT: rfft, irfft, rfft2, irfft2, rfftn, irfftn
- Hermitian FFT: hfft, ihfft
- Helper Routines: fftfreq, rfftfreq, fftshift, ifftshift

### fft/ifft

| argument/keyword | NumPy | CuPy | Dask | JAX | MXNet | Pytorch | TF |
|---|---|---|---|---|---|---|---|
| n | Y | Y | Y | Y | N | Y | N |
| axis/dim | Y | Y | Y | Y | N | Y | N |
| norm | Y | Y | N | Y | N | Y | N |
| data | N | N | N | N | Y | N | N |
| out | N | N | N | N | Y | N | N |
| name | N | N | N | N | Y | N | Y |

### fft2/ifft2

| argument/keyword | NumPy | CuPy | Dask | JAX  | Pytorch | TF |
|---|---|---|---|---|---|---|
| s | Y | Y | Y | Y | Y | N |
| axes/dim | Y | Y | Y | Y | Y | N |
| norm | Y | Y | N | Y | Y | N |
| name | N | N | N | N | N | Y |

_These functions are not implemented in MXNet_.

### fftn/ifftn

| argument/keyword | NumPy | CuPy | Dask | JAX  | Pytorch |
|---|---|---|---|---|---|
| n | Y | Y | N | N | Y |
| s | N | N | Y | Y | N |
| axis/axes/dim | Y | Y | Y | Y | Y |
| norm | Y | Y | N | Y | Y |

_These functions are not implemented in MXNet and TF_.

### rfft/irfft

| argument/keyword | NumPy | CuPy | Dask | JAX  | Pytorch | TF |
|---|---|---|---|---|---|---|
| n | Y | Y | Y | Y | Y | N |
| axis/dim | Y | Y | Y | Y | Y | N |
| norm | Y | Y | N | Y | Y | N |
| fft_length | N | N | N | N | N | Y |
| name | N | N | N | N | N | Y |

_These functions are not implemented in MXNet_.

### rfft2/irfft2

[...]

PR #720, opened 2023-12-12
Clean up some wording in the fft extension

### LINKED ISSUE #717: Return dtype for fftfreq and rfftfreq
The specs for [fftfreq](https://data-apis.org/array-api/draft/extensions/generated/array_api.fft.fftfreq.html#array_api.fft.fftfreq) and [rfftfreq](https://data-apis.org/array-api/draft/extensions/generated/array_api.fft.rfftfreq.html#array_api.fft.rfftfreq) say

"The returned array must have a real-valued floating-point data type determined by [Type Promotion Rules](https://data-apis.org/array-api/draft/API_specification/type_promotion.html#type-promotion)."

But there is no type promotion here, since none of the input arguments are arrays. Presumably it should say that the return type should have the default real floating-point dtype. 

But that also begs the question of whether these functions should have a `dtype` argument like the other array creation functions.

asmeurer:
Fixes #717.

I also plan to fix https://github.com/data-apis/array-api/issues/718 here.

Note: this should also be backported, but I want to make sure the wording here looks good first before I do that.

Also, I haven't done it here, but we should consider whether we want to add a `dtype` argument to `fftfreq` and `rfftfreq` for consistency with the other creation functions. Right now, they return the default real floating-point dtype, but there's no way to change that. However, note that np.fft does not implement a dtype argument for these functions.

Ref: https://github.com/data-apis/array-api/pull/189

### DIFF
--- a/src/array_api_stubs/_2022_12/fft.py
+++ b/src/array_api_stubs/_2022_12/fft.py
@@ -491,19 +491,16 @@ def ihfft(
     ----------
     x: array
         input array. Must have a real-valued floating-point data type.
-    n: int
-        length of the transformed axis of the **input**. If ``n`` is not ``None`` and
+    n: Optional[int]
+        number of elements over which to compute the transform along the axis (dimension) specified by ``axis``. Let ``M`` be the size of the input array along the axis specified by ``axis``. When ``n`` is ``None``, the function must set ``n`` equal to ``M``.
 
-        - ``n`` is greater than the length of the input array along ``axis``, the input array along ``axis`` is zero-padded to length ``n``.
-        - ``n`` is less than the length of the input array along ``axis``, the input array along ``axis`` is trimmed to length ``n``.
-
-        If ``n`` is ``None`` (not provided), the full length of the input array along ``axis`` must be used. The length along the transformed axis must equal ``x.shape[axis]//2 + 1``.
+        -   If ``n`` is greater than ``M``, the axis specified by ``axis`` must be zero-padded to size ``n``.
+        -   If ``n`` is less than ``M``, the axis specified by ``axis`` must be trimmed to size ``n``.
+        -   If ``n`` equals ``M``, all elements along the axis specified by ``axis`` must be used when computing the transform.
 
         Default: ``None``.
     axis: int
-        axis (dimension) over which to compute the Fourier transform. If not set, the last axis (dimension) is used.
-
-        Default: ``-1``.
+        axis (dimension) of the input array over which to compute the transform. A valid ``axis`` must be an integer on the interval ``[-N, N)``, where ``N`` is the rank (number of dimensions) of ``x``. If an ``axis`` is specified as a negative integer, the function must determine the axis along which to compute the transform by counting backward from the last dimension (where ``-1`` refers to the last dimension). Default: ``-1``.
     norm: Literal['backward', 'ortho', 'forward']
         normalization mode. Should be one of the following modes:
 
@@ -516,7 +513,7 @@ def ihfft(
     Returns
     -------
     out: array
-        an array transformed along the axis (dimension) indicated by ``axis``. The returned array must have a complex floating-point data type whose precision matches the precision of ``x`` (e.g., if ``x`` is ``float64``, then the returned array must have a ``complex128`` data type). The length along the transformed axis is ``n//2 + 1`` if ``n`` is given, or ``x.shape[axis]//2 + 1`` otherwise. The lengths along the un-transformed axes remain unchanged.
+        an array transformed along the axis (dimension) specified by ``axis``. The returned array must have a complex floating-point data type whose precision matches the precision of ``x`` (e.g., if ``x`` is ``float64``, then the returned array must have a ``complex128`` data type). The returned array must have the same shape as ``x``, except for the axis specified by ``axis`` which must have size ``n//2 + 1``.
 
     Notes
     -----
--- a/src/array_api_stubs/_2022_12/fft.py
+++ b/src/array_api_stubs/_2022_12/fft.py
@@ -41,7 +41,7 @@ def fft(
 
         -   If ``n`` is greater than ``M``, the axis specified by ``axis`` must be zero-padded to size ``n``.
         -   If ``n`` is less than ``M``, the axis specified by ``axis`` must be trimmed to size ``n``.
-        -   If ``n`` equals ``M``, all elements along the axis specified by ``axis`` must be used when computing the discrete Fourier transform.
+        -   If ``n`` equals ``M``, all elements along the axis specified by ``axis`` must be used when computing the transform.
 
         Default: ``None``.
     axis: int
--- a/src/array_api_stubs/_2022_12/fft.py
+++ b/src/array_api_stubs/_2022_12/fft.py
@@ -526,7 +526,7 @@ def fftfreq(n: int, /, *, d: float = 1.0, device: Optional[device] = None) -> ar
     """
     Computes the discrete Fourier transform sample frequencies.
 
-    For a Fourier transform of length ``n`` and length unit of ``d`` the frequencies are described as:
+    For a Fourier transform of length ``n`` and length unit of ``d``, the frequencies are described as:
 
     .. code-block::
 
@@ -558,7 +558,7 @@ def rfftfreq(n: int, /, *, d: float = 1.0, device: Optional[device] = None) -> a
     """
     Computes the discrete Fourier transform sample frequencies (for ``rfft`` and ``irfft``).
 
-    For a Fourier transform of length ``n`` and length unit of ``d`` the frequencies are described as:
+    For a Fourier transform of length ``n`` and length unit of ``d``, the frequencies are described as:
 
     .. code-block::
 
@@ -590,7 +590,7 @@ def rfftfreq(n: int, /, *, d: float = 1.0, device: Optional[device] = None) -> a
 
 def fftshift(x: array, /, *, axes: Optional[Union[int, Sequence[int]]] = None) -> array:
     """
-    Shift the zero-frequency component to the center of the spectrum.
+    Shifts the zero-frequency component to the center of the spectrum.
 
     This function swaps half-spaces for all axes (dimensions) specified by ``axes``.
 
@@ -607,7 +607,7 @@ def fftshift(x: array, /, *, axes: Optional[Union[int, Sequence[int]]] = None) -
     Returns
     -------
     out: array
-        the shifted array. The returned array must have the same data type as ``x``. The returned array must have the same shape as the input array ``x``.
+        the shifted array. The returned array must have the same data type and shape as ``x``.
 
     Notes
     -----
@@ -635,7 +635,7 @@ def ifftshift(
     Returns
     -------
     out: array
-        the shifted array. The returned array must have the same data type as ``x``. The returned array must have the same shape as the input array ``x``.
+        the shifted array. The returned array must have the same data type and shape as ``x``.
 
     Notes
     -----
--- a/src/array_api_stubs/_2022_12/fft.py
+++ b/src/array_api_stubs/_2022_12/fft.py
@@ -37,7 +37,7 @@ def fft(
     x: array
         input array. Should have a complex-valued floating-point data type.
     n: Optional[int]
-        number of elements in the transformed axis of the output array. Let ``M`` be the size of the input array array along the axis (dimension) specified by ``axis``. When ``n`` is ``None``, the function must set ``n`` equal to ``M``.
+        number of elements in the transformed axis of the output array. Let ``M`` be the size of the input array along the axis (dimension) specified by ``axis``. When ``n`` is ``None``, the function must set ``n`` equal to ``M``.
 
         -   If ``n`` is greater than ``M``, the axis specified by ``axis`` must be zero-padded to size ``n``.
         -   If ``n`` is less than ``M``, the axis specified by ``axis`` must be trimmed to size ``n``.
@@ -86,7 +86,7 @@ def ifft(
     x: array
         input array. Should have a complex-valued floating-point data type.
     n: Optional[int]
-        number of elements in the transformed axis of the output array. Let ``M`` be the size of the input array array along the axis (dimension) specified by ``axis``. When ``n`` is ``None``, the function must set ``n`` equal to ``M``.
+        number of elements in the transformed axis of the output array. Let ``M`` be the size of the input array along the axis (dimension) specified by ``axis``. When ``n`` is ``None``, the function must set ``n`` equal to ``M``.
 
         -   If ``n`` is greater than ``M``, the axis specified by ``axis`` must be zero-padded to size ``n``.
         -   If ``n`` is less than ``M``, the axis specified by ``axis`` must be trimmed to size ``n``.
@@ -135,22 +135,17 @@ def fftn(
     x: array
         input array. Should have a complex-valued floating-point data type.
     s: Optional[Sequence[int]]
-        length of each transformed axis of the output. If
+        number of elements in the transformed axes (dimensions) of the output array. Let ``i`` be the index of the nth axis specified by ``axes`` and ``M[i]`` be the size of the input array along axis ``i``. When ``s`` is ``None``, the function must set ``s`` equal to a sequence of integers, such that, for all ``i``, ``s[i]`` equals ``M[i]``.
 
-        - ``s[i]`` is greater than the length of the input array along the corresponding axis (dimension) ``i``, the input array along the axis ``i`` is zero-padded to length ``s[i]``.
-        - ``s[i]`` is less than the length of the input array along a corresponding axis (dimension) ``i``, the input array along the axis ``i`` is trimmed to length ``s[i]``.
-        - ``s[i]`` is ``-1``, the whole input array along the axis ``i`` is used (no padding/trimming).
-        - ``s`` is not provided, the length of each transformed axis (dimension) in the output array must equal the length of the corresponding axis in the input array.
+        -   If ``s[i]`` is greater than ``M[i]``, axis ``i`` must be zero-padded to size ``s[i]``.
+        -   If ``s[i]`` is less than ``M[i]``, axis ``i`` must be trimmed to size ``s[i]``.
+        -   If ``s[i]`` equals ``M[i]`` or ``-1``, all elements along axis ``i`` must be used when computing the transform.
 
-        If ``s`` is not ``None``, ``axes`` must not be ``None`` either, and ``s[i]`` corresponds to the length along the transformed axis specified by ``axes[i]``.
-
-        Default: ``None``.
+        If ``s`` is not ``None``, ``axes`` must not be ``None``. Default: ``None``.
     axes: Optional[Sequence[int]]
-        axes (dimensions) over which to compute the Fourier transform. If ``None``, all axes must be transformed.
+        axes (dimensions) over which to compute the transform. A valid axis must be an integer on the interval ``[-N, N)``, where ``N`` is the rank (number of dimensions) of ``x``. If an axis is specified as a negative integer, the function must determine the axis along which to compute the transform by counting backward from the last dimension (where ``-1`` refers to the last dimension).
 
-        If ``s`` is specified, the corresponding ``axes`` to be transformed must be explicitly specified too.
-
-        Default: ``None``.
+        If ``s`` is provided, the corresponding ``axes`` to be transformed must also be provided. If ``axes`` is ``None``, the function must compute the transform over all axes. Default: ``None``.
     norm: Literal['backward', 'ortho', 'forward']
         normalization mode. Should be one of the following modes:
 
@@ -165,7 +160,7 @@ def fftn(
     Returns
     -------
     out: array
-        an array transformed along the axes (dimension) indicated by ``axes``. The returned array must have the same data type as ``x``. The length along each transformed axis ``i`` is ``s[i]``, if ``s`` is given and ``s[i]`` is not ``-1``, or the length of the corresponding axis of the input array otherwise. The lengths along the un-transformed axes remain unchanged.
+        an array transformed along the axes (dimensions) specified by ``axes``. The returned array must have the same data type as ``x`` and must have the same shape as ``x``, except for the axes specified by ``axes`` which must have size ``s[i]``.
 
     Notes
     -----
--- a/src/array_api_stubs/_2022_12/fft.py
+++ b/src/array_api_stubs/_2022_12/fft.py
@@ -404,7 +404,7 @@ def irfftn(
     axes: Optional[Sequence[int]]
        axes (dimensions) over which to compute the transform. A valid axis must be an integer on the interval ``[-N, N)``, where ``N`` is the rank (number of dimensions) of ``x``. If an axis is specified as a negative integer, the function must determine the axis along which to compute the transform by counting backward from the last dimension (where ``-1`` refers to the last dimension).
 
-        If ``s`` is provided, the corresponding ``axes`` to be transformed must also be provided. If ``axes`` is ``None``, the function must compute the transform over all axes. Default: ``None``.
+       If ``s`` is provided, the corresponding ``axes`` to be transformed must also be provided. If ``axes`` is ``None``, the function must compute the transform over all axes. Default: ``None``.
     norm: Literal['backward', 'ortho', 'forward']
         normalization mode. Should be one of the following modes:
 
--- a/src/array_api_stubs/_2022_12/fft.py
+++ b/src/array_api_stubs/_2022_12/fft.py
@@ -398,7 +398,7 @@ def irfftn(
 
         -   If ``n`` is greater than ``M[i]``, axis ``i`` of the input array must be zero-padded to size ``n``.
         -   If ``n`` is less than ``M[i]``, axis ``i`` of the input array must be trimmed to size ``n``.
-        -   If ``n`` equals ``M[i]`` or ``-1``, all elements along the axis ``i`` of the input array must be used when computing the transform.
+        -   If ``n`` equals ``M[i]`` or ``-1``, all elements along axis ``i`` of the input array must be used when computing the transform.
 
         If ``s`` is not ``None``, ``axes`` must not be ``None``. Default: ``None``.
     axes: Optional[Sequence[int]]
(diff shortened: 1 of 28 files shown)


asmeurer:
"trimmed to length n" also isn't very clear. Does this mean it should take the first `n` elements?

src/array_api_stubs/_draft/fft.py
```diff
@@ -60,7 +60,7 @@ def fft(
     Returns
     -------
     out: array
-        an array transformed along the axis (dimension) indicated by ``axis``. The returned array must have a complex floating-point data type determined by :ref:`type-promotion`.
+        an array transformed along the axis (dimension) indicated by ``axis``. If ``x`` has a complex floating-point data type, the returned array must have the same data type as ``x``. If ``x`` has a real-valued floating-point data type, the returned array must have a complex floating-point data type whose precision matches the precision of ``x`` (e.g., if ``x`` is ``float64``, then the returned array must have a ``complex128`` data type).
```
leofang, line 63: I am against the change to accept real input here. I think in the past meetings we've decided to not allow that, and instead users must do explicit casting here.

asmeurer: This is based on what was already specified (https://github.com/data-apis/array-api/pull/720/files#diff-31eaf0d090ecc4df2af89ab3d3a33b91ff8846925b579fd64bf065aea146a230R38).

src/array_api_stubs/_draft/fft.py
```diff
@@ -329,7 +329,7 @@ def irfft(
     Returns
     -------
     out: array
-        an array transformed along the axis (dimension) indicated by ``axis``. The returned array must have a real-valued floating-point data type determined by :ref:`type-promotion`. The length along the transformed axis is ``n`` (if given) or ``2*(m-1)`` (otherwise).
+        an array transformed along the axis (dimension) indicated by ``axis``. The returned array must have the same data type as ``x``. The length along the transformed axis is ``n`` (if given) or ``2*(m-1)`` (otherwise).
```
leofang, line 332: This is incorrect, irfft/irfftn mean complex-to-real (C2R) transforms, and they must return real data types.

src/array_api_stubs/_draft/fft.py
leofang: If it leaves room for alternative interpretation, we must shut down those loopholes, and here is not one 🙂

asmeurer: I don't think this is an ambiguity. "floating-point data type" in the spec right now means "real or complex" (see https://data-apis.org/array-api/draft/API_specification/data_types.html#floating-point-data-types). Right now the spec does allow both types. We should revisit this if it was supposed to be otherwise.

asmeurer:
I've also pushed an update to specify output lengths. Please check carefully.

I've also tried to clean up the wording a bit. However, I'm not sure if my wording here is always consistent with wording elsewhere in the spec, so let me know if something should be changed. 

I noticed that we aren't consistent throughout the spec with "length" vs. "size" to refer to the size of an axis.

src/array_api_stubs/_draft/fft.py
leofang: (fixed a typo...)

honno: Previous discussion at https://github.com/data-apis/array-api/pull/696#issuecomment-1754325857

asmeurer: That discussion is for `rfft` and friends. But Leo is saying that the main `fft` should only accept `complex`, not `float`.

rgommers:
Okay, this took a bit of digging, but I believe that @leofang is largely correct. The current state of the `fft` docs is inconsistent, and has been in that state since the initial merge of gh-189. The problem is that the _Parameters_ section says that the first input `x` must have a "floating-point dtype" (so may be real), while the prominent note right above it says:

_"Applying the one-dimensional inverse discrete Fourier transform to the output of this function must return the original (i.e., non-transformed) input array within numerical accuracy (i.e., `ifft(fft(x)) == x`), provided that the transform and inverse transform are performed with the same arguments (length, axis, and normalization mode)."_

Clearly that is not the case if `x` would be real, because the roundtripping changes the dtype from real to complex (e.g., `float64` to `complex128`). Hence real input should be rejected.

So far the spec - let's update the wording to reflect that. For implementations:
- The reference implementation should raise on non-complex input dtypes to `fft`/`fftn`.
- For NumPy 2.0, that'd be one bridge too far I think, so let's open an issue first - there are more issues with `numpy.fft` that need addressing. Accepting non-complex input and treating it like complex with a zero imaginary component will simply be one oddity that perhaps can be addressed over time.

asmeurer:
Would we want to backport this to 2022.12, or just add it to the draft?

asmeurer:
I've pushed a change that updates these functions to only require complex-valued inputs. I think this will also require a changelog entry, but I didn't add it yet because I don't know if it will be part of 2023.12 only or backported.

This PR also requires some review for other the wording changes that have been made, as discussed above.

kgryte:
IMO, I think it is a bug in the 2022.12 revision and should be backported. For the changelog entry, we can add that in a subsequent PR along with the other specification changes which should be documented similarly.

rgommers:
These changes all LGTM. I think if they are backported and added to the Changelog, this is good to merge.

leofang:
Sorry I dropped the ball here. I think the PR looks good now from a quick glance. I can do a more thorough look tomorrow, but don't let me stop the progress 🙂 Thanks Aaron and all for the improvement!

asmeurer:
I've backported everything to 2022.12. I've left out the changelog based on https://github.com/data-apis/array-api/pull/720#issuecomment-1884048884.

rgommers (approved):
Approving and moving ahead with the matching implementation from @asmeurer in `numpy.array_api`: https://github.com/numpy/numpy/pull/25317.

I'll hold off from merging for a bit more since @leofang wanted to have a closer look still.

src/array_api_stubs/_draft/fft.py
```diff
@@ -387,7 +387,7 @@ def rfftn(
     Returns
     -------
     out: array
-        an array transformed along the axes (dimension) indicated by ``axes``. The returned array must have a complex-valued floating-point data type determined by :ref:`type-promotion`.
+        an array transformed along the axes (dimension) indicated by ``axes``. The returned array must have a complex floating-point data type whose precision matches the precision of ``x`` (e.g., if ``x`` is ``float64``, then the returned array must have a ``complex128`` data type). The length along each transformed axis ``i`` is ``s[i]``, if ``s`` is given and ``s[i]`` is not ``-1``, or the length of the corresponding axis of the input array otherwise.
```
asmeurer, line 390: This doesn't match what NumPy does. From [the NumPy docs for `rfftn`](https://numpy.org/doc/stable/reference/generated/numpy.fft.rfftn.html):

> The length of the last axis transformed will be s[-1]//2+1, while the remaining transformed axes will have lengths according to s, or unchanged from the input.

(here `s` defaults to the shape of the input along the given axes if not given)

src/array_api_stubs/_2022_12/fft.py
```diff
@@ -261,7 +261,7 @@ def rfft(
     Returns
     -------
     out: array
-        an array transformed along the axis (dimension) indicated by ``axis``. The returned array must have a complex-valued floating-point data type determined by :ref:`type-promotion`.
+        an array transformed along the axis (dimension) indicated by ``axis``. The returned array must have a complex floating-point data type whose precision matches the precision of ``x`` (e.g., if ``x`` is ``float64``, then the returned array must have a ``complex128`` data type). The length along the transformed axis is ``n``, if given, or the length of the input along the axis specified by ``axis`` otherwise.
```
asmeurer, line 264: This also disagrees with [NumPy's `rfft`](https://numpy.org/doc/stable/reference/generated/numpy.fft.rfft.html):

> The truncated or zero-padded input, transformed along the axis indicated by axis, or the last one if axis is not specified. If n is even, the length of the transformed axis is (n/2)+1. If n is odd, the length is (n+1)/2.

asmeurer:
Found two more places where the specified output shape didn't match the numpy docs/implementation (`rfft` and `rfftn`). Need someone to confirm that the spec really should be changed here to match what NumPy does. Note that technically, the spec doesn't currently say what the output shape for those functions should be at all.

asmeurer:
The spec for `irfft` says

> Applying the one-dimensional inverse discrete Fourier transform for real-valued input to the output of this function must return the original (i.e., non-transformed) input array within numerical accuracy (i.e., ``irfft(rfft(x)) == x``), provided that the transform and inverse transform are performed with the same arguments (axis and normalization mode) and consistent length.

But the NumPy docs say 

>  In other words, irfft(rfft(a), len(a)) == a to within numerical accuracy. (See Notes below for why len(a) is necessary here.)

See the "notes" section at https://numpy.org/doc/stable/reference/generated/numpy.fft.irfft.html

I don't know if this caveat applies to other functions here as well.

(as an aside, is it intentional that `n` is not allowed to be positional in fft functions?)

asmeurer:
Another one: NumPy 2.0 has a new deprecation warning if `axes` is `None` and `s` is not `None`:

```
>>> np.fft.irfftn(np.asarray([0.+0.j, 0.+0.j]), s=(1,), axes=None)
<stdin>:1: DeprecationWarning: `axes` should not be `None` if `s` is not `None` (Deprecated in NumPy 2.0). In a future version of NumPy, this will raise an error and `s[i]` will correspond to the size along the transformed axis specified by `axes[i]`. To retain current behaviour, pass a sequence [0, ..., k-1] to `axes` for an array of dimension k.
```

The spec doesn't seem to disallow this.

asmeurer:
> (as an aside, is it intentional that n is not allowed to be positional in fft functions?)

I didn't see any discussion of this in https://github.com/data-apis/array-api/pull/189 or https://github.com/data-apis/array-api/issues/159.

Just based on a cursory [GitHub code search](https://github.com/search?q=%2Fnp.fft.fft%5C%28.*%2C+.*%5C%29%2F++path%3A*.py&type=code), it seems like passing `n` as positional is common. See also https://github.com/data-apis/python-record-api/blob/master/data/typing/numpy.fft.py (does that data distinguish between when an argument is passed positionally or as a keyword?).

With that being said, I'm haven't used the fft APIs so I can't say myself whether n being positional is something we should aim for or not. 

I also want to clarify that I don't think this is an issue for NumPy either way. Right now NumPy defines `n` as a positional argument, but there's not really an incompatibility with doing so when the standard says keyword-only (at least in my opinion).

asmeurer:
@leofang do you have any comments on my suggestions in my most recent comments? If there are no objections, I can update the spec accordingly.

src/array_api_stubs/_2022_12/fft.py
```diff
@@ -312,7 +312,7 @@ def irfft(
     Returns
     -------
     out: array
-        an array transformed along the axis (dimension) indicated by ``axis``. The returned array must have a real-valued floating-point data type determined by :ref:`type-promotion`. The length along the transformed axis is ``n`` (if given) or ``2*(m-1)`` (otherwise).
+        an array transformed along the axis (dimension) indicated by ``axis``. The returned array must have a real-valued floating-point data type whose precision matches the precision of ``x`` (e.g., if ``x`` is ``complex128``, then the returned array must have a ``float64`` data type). The length along the transformed axis is ``n``, if given or ``2*(m-1)`` otherwise.
```
leofang, line 315: I think NumPy's `irfft` behavior is correct.
```suggestion
        an array transformed along the axis (dimension) indicated by ``axis``. The returned array must have a real-valued floating-point data type whose precision matches the precision of ``x`` (e.g., if ``x`` is ``complex128``, then the returned array must have a ``float64`` data type). The length along the transformed axis is ``n`` if given, or ``2*(m-1)`` otherwise. Therefore, to get an odd number of output points along ``axis``, ``n`` must be specified.
```

src/array_api_stubs/_2022_12/fft.py
leofang: Yes, would you mind fixing it Aaron?

src/array_api_stubs/_2022_12/fft.py
```diff
@@ -428,7 +428,7 @@ def irfftn(
     Returns
     -------
     out: array
-        an array transformed along the axes (dimension) indicated by ``axes``. The returned array must have a real-valued floating-point data type determined by :ref:`type-promotion`. The length along the last transformed axis is ``s[-1]`` (if given) or ``2*(m - 1)`` (otherwise), and all other axes ``s[i]``.
+        an array transformed along the axes (dimension) indicated by ``axes``. The returned array must have a real-valued floating-point data type whose precision matches the precision of ``x`` (e.g., if ``x`` is ``complex128``, then the returned array must have a ``float64`` data type). The length along the last transformed axis is ``s[-1]`` if given, or ``2*(m - 1)`` otherwise, and all other axes ``s[i]``.
```
leofang, line 431:
```suggestion
        an array transformed along the axes (dimension) indicated by ``axes``. The returned array must have a real-valued floating-point data type whose precision matches the precision of ``x`` (e.g., if ``x`` is ``complex128``, then the returned array must have a ``float64`` data type). The length along the last transformed axis is ``s[-1]`` if given, or ``2*(m - 1)`` otherwise, and all other axes ``s[i]``. Therefore, to get an odd number of output points along the last axis, ``s`` should be specified.
```

src/array_api_stubs/_draft/fft.py
leofang: Yes, would you mind fixing it Aaron?

leofang:
For this question 
> The spec for `irfft` says
> 
> > Applying the one-dimensional inverse discrete Fourier transform for real-valued input to the output of this function must return the original (i.e., non-transformed) input array within numerical accuracy (i.e., `irfft(rfft(x)) == x`), provided that the transform and inverse transform are performed with the same arguments (axis and normalization mode) and consistent length.
> 
> But the NumPy docs say
> 
> > In other words, irfft(rfft(a), len(a)) == a to within numerical accuracy. (See Notes below for why len(a) is necessary here.)
> 
> See the "notes" section at https://numpy.org/doc/stable/reference/generated/numpy.fft.irfft.html
> 
> I don't know if this caveat applies to other functions here as well.

I think we were being sloppy back then, as the idea of performing round-trip transforms to get back the original data was clear, but a lot of caveats need to be handled carefully such as `s` and `n` choices. I didn't think I could come up with a concise expression to describe it accurately enough. If anyone has suggestion, please help phrase it better...

> (as an aside, is it intentional that `n` is not allowed to be positional in fft functions?)

I prefer making it kw-only, but no strong opinion.

leofang (changes requested):

leofang:
> Another one: NumPy 2.0 has a new deprecation warning if `axes` is `None` and `s` is not `None`:
> 
> ```
> >>> np.fft.irfftn(np.asarray([0.+0.j, 0.+0.j]), s=(1,), axes=None)
> <stdin>:1: DeprecationWarning: `axes` should not be `None` if `s` is not `None` (Deprecated in NumPy 2.0). In a future version of NumPy, this will raise an error and `s[i]` will correspond to the size along the transformed axis specified by `axes[i]`. To retain current behaviour, pass a sequence [0, ..., k-1] to `axes` for an array of dimension k.
> ```
> 
> The spec doesn't seem to disallow this.

We did say "If s is not None, axes must not be None either." Do I misread your question Aaron?
https://github.com/data-apis/array-api/blob/95332bb71159cea68e15e7d20364a3e07060a402/src/array_api_stubs/_draft/fft.py#L425

leofang:
> @leofang do you have any comments on my suggestions in my most recent comments? If there are no objections, I can update the spec accordingly.

@asmeurer so sorry for my unexpected delay 🙇 I just left some comments. Yes, please go ahead and fix them, and ping me for re-review when you're done. During an internal discussion it came to my attention that we need to make the output shape of C2R (i.e. `irfft`, `irfftn`) clearer too. I made suggestions above.

asmeurer:
> I think we were being sloppy back then, as the idea of performing round-trip transforms to get back the original data was clear, but a lot of caveats need to be handled carefully such as s and n choices. I didn't think I could come up with a concise expression to describe it accurately enough. If anyone has suggestion, please help phrase it better...

Actually, I didn't notice it already caveats this in the text:

> Applying the one-dimensional inverse discrete Fourier transform for real-valued input to the output of this function must return the original (i.e., non-transformed) input array within numerical accuracy (i.e., ``irfft(rfft(x)) == x``), **provided that the transform and inverse transform are performed with the same arguments (axis and normalization mode) and consistent length.**

(emphasis mine)

So I think we don't need to change anything here.

asmeurer:
> We did say "If s is not None, axes must not be None either." Do I misread your question Aaron?

It looks like you're right. I must have missed this text previously.

asmeurer:
Also noticed another issue, quite a few of the type hints do not properly use `Optional` for arguments that could be `None`.

src/array_api_stubs/_2022_12/fft.py
asmeurer: Do you mean ``s`` must be specified?

leofang: This is `irfft` not `irfftn` so I think it's `n` not `s`. 

<rant> we made a bad choice in the `s`/`n` argument names, NumPy was a bad model, is it too late to change? </rant>

asmeurer:
@leofang I just pushed a commit with the proposed fixes. If you could review the text and make sure it is both correct and sufficiently unambiguous, that would help. I'm particularly not sure about the best way to word things with the `s` given that there are so many conditions (`s` could be `None` or `s[i]` could be `-1`). 

I have only updated `_draft` for now. Once we agree with the changes I will backport all changes to `_2022_12`.

asmeurer:
One final (hopefully) thing which I have not done here is adding a `dtype` argument to the fft creation fucntions `fftfreq` and `rfftfreq`. This would be an actual API change, and one for which NumPy is not currently compatible (although torch is). So if we want to make this change it should not be backported. Perhaps it would be best to do this only for the 2024 spec, given that it hasn't actually been requested by anyone yet, and we are close to tagging 2023. I will open a new issue about this.

asmeurer:
Actually https://github.com/data-apis/array-api/issues/717 can already be the issue for this.

asmeurer:
@kgryte could also use your review here to make sure the wording looks good everywhere. I modified quite a few sentences to make them less ambiguous. Again, please only look at `_draft` for now. I will backport everything to `_2022_12` once it all looks good.

leofang:
I am reviewing now

src/array_api_stubs/_draft/fft.py
```diff
@@ -278,7 +278,8 @@ def rfft(
     Returns
     -------
     out: array
-        an array transformed along the axis (dimension) indicated by ``axis``. The returned array must have a complex-valued floating-point data type determined by :ref:`type-promotion`.
+        an array transformed along the axis (dimension) indicated by ``axis``. The returned array must have a complex floating-point data type whose precision matches the precision of ``x`` (e.g., if ``x`` is ``float64``, then the returned array must have a ``complex128`` data type). If ``n`` is even, the length of the transformed axis is ``(n/2)+1``. If ``n`` is odd, the length is ``(n+1)/2``. If ``n`` is ``None``, it defaults to ``x.shape[axis]``.
+
```
leofang, line 282:
```suggestion
```

src/array_api_stubs/_draft/fft.py
```diff
@@ -387,7 +388,8 @@ def rfftn(
     Returns
     -------
     out: array
-        an array transformed along the axes (dimension) indicated by ``axes``. The returned array must have a complex-valued floating-point data type determined by :ref:`type-promotion`.
+        an array transformed along the axes (dimension) indicated by ``axes``. The returned array must have a complex floating-point data type whose precision matches the precision of ``x`` (e.g., if ``x`` is ``float64``, then the returned array must have a ``complex128`` data type). The length along the last transformed axis must be equal to ``s[-1]//2+1`` and the length along the remaining transformed axes ``i``  must be equal to ``s[i]``, where ``s`` defaults to ``x.shape`` if the parameter ``s`` is ``None`` and ``s[i]`` defaults to ``x.shape[i]`` if ``s[i]`` is ``-1``.
+
```
leofang, line 392:
```suggestion
```

leofang:
@asmeurer @kgryte I could use extra pairs of eyes 🥺

kgryte:
@leofang Of course. Will take a look tonight.

leofang:
> I'm particularly not sure about the best way to word things with the `s` given that there are so many conditions (`s` could be `None` or `s[i]` could be `-1`).

In commit [a84c293](https://github.com/data-apis/array-api/pull/720/commits/a84c293dfa163816ee3d999240f2da007f8b4f67) I tried to separate out the `None` case, as it could be a bit nasty especially with `irfftn`.

src/array_api_stubs/_2022_12/fft.py
```diff
@@ -18,13 +35,13 @@ def fft(
...
-        input array. Should have a floating-point data type.
+        input array. Should have a complex-valued floating-point data type.
     n: int
         length of the transformed axis of the output. If
 
-        - ``n`` is greater than the length of the input array, the input array is zero-padded to length ``n``.
-        - ``n`` is less than the length of the input array, the input array is trimmed to length ``n``.
-        - ``n`` is not provided, the length of the transformed axis of the output must equal the length of the input along the axis specified by ``axis``.
+        - ``n`` is greater than the length of the input array along ``axis``, the input array along ``axis`` is zero-padded to length ``n``.
+        - ``n`` is less than the length of the input array along ``axis``, the input array along ``axis`` is trimmed to length ``n``.
```
kgryte, line 43:
```suggestion
        - ``n`` is less than the length of the input array along ``axis``, the input array along ``axis`` must be trimmed to length ``n``.
```

src/array_api_stubs/_2022_12/fft.py
```diff
@@ -18,13 +35,13 @@ def fft(
...
+        - ``n`` is not provided, the full length of the input array along ``axis`` must be used.
```
kgryte, line 44:
```suggestion
        - ``n`` is not ``None``, the full length of the input array along ``axis`` must be used.
```

src/array_api_stubs/_2022_12/fft.py
```diff
@@ -18,13 +35,13 @@ def fft(
...
     x: array
-        input array. Should have a floating-point data type.
+        input array. Should have a complex-valued floating-point data type.
     n: int
         length of the transformed axis of the output. If
 
-        - ``n`` is greater than the length of the input array, the input array is zero-padded to length ``n``.
-        - ``n`` is less than the length of the input array, the input array is trimmed to length ``n``.
-        - ``n`` is not provided, the length of the transformed axis of the output must equal the length of the input along the axis specified by ``axis``.
+        - ``n`` is greater than the length of the input array along ``axis``, the input array along ``axis`` must be zero-padded to length ``n``.
```
kgryte, line 42: In general, in order to refer to the number of elements along an axis, we have used, tmk, "size", rather than "length". We didn't adhere to that in the original FFT PR, likely as we more or less copied the phrasing from NumPy docs.

If we are comfortable adopting consistent language (i.e., using "size"), I can do a pass over this PR and update accordingly.

leofang: Yup if you have time, thx 🙏

asmeurer: Well I was trying to normalize things here, but in the other direction (using "length" instead of "size"). The reason I would prefer "length" to "size" is that `size` is an attribute on the array object that refers to the total number of elements. There was also a related discussion in NumPy https://github.com/numpy/numpy/pull/25140

kgryte: @asmeurer Thanks for that context. I am not sure I agree with the conclusions of that discussion, but I don't think this PR is the right place to change course. As "size" is the current term use throughout the specification when referring to the number of elements along a dimension, we should, IMO, continue to adhere to that convention in this PR.

As a follow-up, I suggest opening a separate issue outlining your concerns and arguments for changing the terminology. Should the reception prove positive, we can then make the changes across the entire specification more generally with potential inclusion in the v2024 revision.

kgryte: If this works for you, I can revert the s/size/length/ changes when I make some additional copy edits to align the language used in this PR with the rest of the specification.

leofang: I support sticking to the current array API convention for now, and open a new issue for terminology discussion.

asmeurer: Yes please proceed with pushing whatever changes are necessary to get this merged.

leofang:
btw I dunno if it's a me problem, but every PR that I touched (this one and others) failed in the CI/CD...

kgryte:
@leofang An unlucky touch? 😅 I'll make a note to investigate once we get these in.

kgryte:
@leofang Looking at the CI logs, looks like you need to locally install dev deps in order to run the pre-commit hook which will apply black formatting before pushing.

src/array_api_stubs/_2022_12/fft.py
```diff
@@ -18,13 +35,13 @@ def fft(
...
+        - ``n`` is less than the length of the input array along ``axis``, the input array along ``axis`` must be trimmed to length ``n``.
```
kgryte, line 43: @leofang Sorry for my ignorance, but "trimmed" in this case means the last `N-n` elements are removed?

kgryte: Let's address this in a small follow-on PR.

leofang:
@kgryte thanks, I'll check again later today. One thing I noticed: the diff contains a lot of non-FFT-related changes, is this expected?

kgryte:
@leofang GitHub seems confused. The actual diff only includes FFT changes. See https://github.com/data-apis/array-api/commit/937d3b8e22ef42d2420a951cefe6ec28107f31bf.

leofang:
Interesting, I thought only Gitlab would be confused not GitHub 😀 We squash-merged, right?

kgryte:
> We squash-merged, right?

Yes, we did. 😀

leofang:
Thanks for all the editing, Aaron/Athan. Looks nice! I sent a tiny follow-up: #743

Issue #621, opened 2023-04-18
How to translate `nanmin` and friends to the Array API?


betatim:
Is there advice around handling NaNs and how to translate Numpy code to using the Array API?

In particular I have code like `np.nanmin(X, axis=0)` that I would like to rewrite so that it works with Numpy, Torch, etc arrays.

To me the "obvious" translation seems to be `xp.min(X[~xp.isnan(X)], axis=0)` but this doesn't work if `X` has a shape like `(7500, 4)` (here `xp` is the namespace of the array `X`). Another option I looked for is a `where=` argument to [`min()`](https://data-apis.org/array-api/latest/API_specification/generated/array_api.min.html), but that doesn't exist unfortunately.

Does anyone have advice on this topic or knows if there is work planned on this?

rgommers:
Good question, thanks @betatim. I think we want a general pattern here that can apply to any reduction. That's a bit of an unsolved problem even in NumPy, where we rejected more niche `nan*` feature requests like `nanptp` and `nancov`.

This discussion seems relevant: https://github.com/pytorch/pytorch/issues/38349#issuecomment-629837734. SciPy's `nan_policy` is now in better shape, but I'm still not sure on what the best approach in general is. Options include:
- a `skip_nan` keyword
- a context manager or other generic method to do achieve the same effect
- a compat library approach with a fast path for libraries known to offer specific functions like `nanmin` and a generic fallback implementation that iterates over the specified `axis=`.

A generic implementation is pretty fiddly
```python
def nanreduction(func, x, axis=None):
    if axis is None:
        return func(x[~np.isnan(x)])
    else:
        # normalize axis, then:
        out_shape = list(x.shape).remove(x.shape[axis])
        out = np.empty_like(x, shape=out_shape)
        for i in range(axis):
            # call `func` in a loop here. this is pretty annoying to get right
            ...
```
NumPy's machinery is pure Python so can be looked at for guidance. It does also raise questions like how to deal with all-nan slices (e.g., https://github.com/numpy/numpy/blob/main/numpy/lib/nanfunctions.py#L355-L360); those warnings tend to not be useful.

ogrisel:
> how to deal with all-nan slices

I think the reduction should allow to pass an explicit scalar value (e.g. typically 0.) as a kwarg and raise an error if all-nan slices occur while this kwarg was left to `None`.

rgommers:
[...]

PR #722, opened 2023-12-14
feat: add API specification for returning the `k` largest elements

### LINKED ISSUE #629: Standard API for topk and / or argpartition
numpy provides an indirect way to compute the smallest (or largest) values of an array [numpy.argpartition](https://numpy.org/doc/stable/reference/generated/numpy.argpartition.html) 

There is also a proposal to implement a simpler (arg)topk in numpy (without allocating a result array of the size of the input array when `k` is small):

- https://github.com/numpy/numpy/pull/19117

Here is a quick review of some available implementations in related libraries:

- [torch.topk](https://pytorch.org/docs/stable/generated/torch.topk.html) (no such thing as `torch.argpartition`)
   - returns a tuple of values and indices
- [jax.lax.top_k](https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.top_k.html)
   - returns a tuple of values and indices
- [dask.array.topk](https://docs.dask.org/en/stable/generated/dask.array.topk.html)
   - returns only the values, I did not find a way to get the indices :(
- [cupy.argpartition](https://docs.cupy.dev/en/stable/reference/generated/cupy.argpartition.html) but internally computes a full `cupy.argsort` which makes it very inefficient  for large arrays and small `k`: O(nlog(n)) instead of O(n).

Motivation: (arg)topk is needed by popular baseline data-science workloads (e.g. k-nearest neighbors classification in scikit-learn) and is surprisingly non trivial to implement efficiently. For instance on GPUs, the fastest implementations are based on some kind of partial radix sort while CPU implementations would use more traditional partial sorting algorithms (as implemented in `std:partial_sort` or `std::nth_element`).

kgryte:
**Update**: This PR

- resolves https://github.com/data-apis/array-api/issues/629 by adding one new API to the Array API specification

  - `top_k`: returns a tuple whose first element is an array containing the top `k` largest (or smallest) values and whose second element is an array containing the indices of those `k` values.

The design decisions largely follow the discussion below. The specialized methods `top_k_indices` and `top_k_values` have been dropped from the specification.

---

This PR

- resolves https://github.com/data-apis/array-api/issues/629 by adding `3` new APIs to the Array API specification

  - `top_k`: returns a tuple whose first element is an array containing the top `k` largest (or smallest) values and whose second element is an array containing the indices of those `k` values.
  - `top_k_indices`: returns an array containing the indices of the `k` largest (or smallest) values.
  - `top_k_values`: returns an array containing the `k` largest (or smallest) values.

## Prior Art

As illustrated in the [API comparison](https://github.com/data-apis/array-api-comparison/blob/4bc91bb2ff0d71a1aeb2ddfd15616fb627b57f6a/signatures/searching/top_k.md), there is currently no consistent API across array libraries for returning the `k` largest or smallest values.

- NumPy has `partition` and `argpartition`, but these return full arrays. When `axis` is `None`, NumPy operates on a flattened input array. To get the top `k` values, one must index and, if wanting sorted values, sort.
- CuPy has `partition` and `argpartition` and follows NumPy; however, for `argpartition`, for implementation reasons, it performs a full sort.
- Dask has `topk` and switches "modes" (largest or smallest) based on whether `k` is positive or negative.
- JAX has `top_k` which ~~only returns values~~ always returns both values and indices, as well as NumPy equivalent `partition` and `argpartition` APIs (however, JAX differs in how it handles NaNs). The function only supports searching along the last axis.
- PyTorch has `topk` which always returns both values and indices.
- TensorFlow has `top_k` which always returns both values and indices and only supports searching along the last axis.

## Proposed APIs

This PR attempts to synthesize the common themes and best ideas for "top k" APIs as observed among array libraries and attempts to define APIs which adhere to specification precedent in order to promote consistent design and reduce cognitive load.

### top_k

```python
def top_k(
    x: array,
    k: int,
    /,
    *,
    axis: Optional[int] = None,
    mode: Literal["largest", "smallest"] = "largest",
) -> Tuple[array, array]
```

Returns a tuple containing the `k` largest (or smallest) elements in `x`. 

```python
def top_k_indices(
    x: array,
    k: int,
    /,
    *,
    axis: Optional[int] = None,
    mode: Literal["largest", "smallest"] = "largest",
) -> array
```

Returns an array containing the indices of the `k` largest (or smallest) elements in `x`.

```python
def top_k_values(
    x: array,
    k: int,
    /,
    *,
    axis: Optional[int] = None,
    mode: Literal["largest", "smallest"] = "largest",
) -> array
```

Returns an array containing the `k` largest (or smallest) elements in `x`.

## Design Decision Rationale

- The default for `axis` is `None` in order to match `min`, `max`, `argmin`, and `argmax`. In those APIs, when `axis` is `None` (the default), the functions operate over a flattened array. Given that `top_k*` may be considered a generalization of the mentioned APIs, ensuring consistency seemed preferable to requiring users to remember a separate set of rules for `top_k*`.
- `axis` only supports `int` and `None` in order to match `argmin` and `argmax`. In `min` and `max`, the specification supports specifying multiple axes. Support for multiple axes can be a future specification extension.
- `top_k` was chosen over `topk` due to naming concerns discussed [elsewhere](https://mail.python.org/archives/list/numpy-discussion@python.org/thread/F4P5UVTAKRJJ3OORI6UOWFSUEE5CNTSC/#PELUDW5ACUBHBNK5IVGWIWTQHBM2HXUP) (namely `top k` vs `to pk`). Furthermore, "top k" follows ML conventions, as opposed to `maxk`/`max_k` or `nlargest`/`nsmallest` as found in other languages.
- The PR includes three separate APIs following the lead of `unique`. In that case, rather than support polymorphic return values (e.g., returning values, returning values and indices, return values and counts, etc), we chose define specific API which are monomorphic in their output. We innovated there, and the thinking that went into those design decisions seemed applicable here, where a user may want only values, indices, or both.
- The PR follows the `unique_*` naming convention, rather than the `arg*` naming convention, as there are three different return value situations: values, indices, and indices and values. Hence, using a suffix to describe what is returned as in `unique_*` seems reasonable and follows existing precedent in the specification.
- The APIs include a "mode" option to specify the type (largest or smallest) of values to return. Most existing array libraries supporting a "top k" API return only the largest values; however, PyTorch supports returning either the smallest or largest and does so via a `largest` keyword argument. This PR chooses to name the kwarg `mode` in order to be more explicit (what does `largest=False` mean to the lay reader?) and follows precedent elsewhere in the specification (e.g., `linalg.qr`) where `mode` is used to toggle between different operating modes.
- The PR does not include a `sorted` kwarg in order to instruct the API to return sorted values (or indices corresponding to sorted values) because (a) the kwarg is not universally supported currently, (b) downstream users can, at least for values, explicitly call `sort` (except in Dask which doesn't currently support full sorting) after calling `top_k` or `top_k_values`, and (c) can be addressed in a future specification extension. Additionally, if we support `sorted`, we may also want to support a `stable` kwarg as in `sort` to allow ensuring that returned indices are consistent when provided the same input array.
- Leaves unspecified what should happen when `k` exceeds the number of elements, as different behaviors seem acceptable (e.g., raising an exception or returning `m < k` values).

## Questions

- Should we be more strict in specifying what should happen when `k` exceeds the number of elements?
- Should zero-dimensional arrays be supported?
- In `argmin` and `argmax`, the specification requires returning the index of first occurrence when a minimum/maximum value occurs multiple times. Given that `top_k*` can be implemented as a partial sort, presumably we do not want specify a first occurrence restriction. Is this a reasonable assumption?
- Should we defer adding support for specifying multiple axes until a future revision of the specification or should we go ahead and add now for parity with `min` and `max`?
- Are we okay with `None` being the default for `axis`, where the default behavior is searching over a flattened array?

## Considerations

The APIs included in this PR have implications for the following array libraries:

- **NumPy**: these will be new APIs, and, similar to `unique_*`, will need to be added to the main namespace.
- **CuPy**: same as NumPy.
- **Dask**: will need to introduce new APIs; however, the new APIs can be implemented as lightweight wrappers around Dask's existing `topk` and `argtopk`.
- **JAX**: were JAX to place the APIs in its `lax` namespace, this PR would introduce breaking changes, as ~~JAX would need to return both values and indices, by default,~~ and JAX would need to flatten by default rather than search along the last dimension. However, if implemented in its `numpy` namespace, these will simply be new APIs. In both scenarios, JAX will need to add support for `axis` and `mode` behavior.
- **PyTorch**: these will be new APIs; however, the new APIs can be implemented as lightweight wrappers around PyTorch's existing `topk`.
- **TensorFlow**: additional APIs (`top_k_values` and `top_k_indices`). If implemented in its `math` namespace, this PR would introduce breaking changes as TensorFlow would need to flatten by default. However, if implemented in its `numpy` namespace, these will simply be new APIs. In both scenarios, TensorFlow will need to add support for `axis` and `mode` behavior.

## Related Links

- <https://github.com/data-apis/array-api/issues/629>
- <https://github.com/numpy/numpy/pull/19117>
- <https://github.com/numpy/numpy/issues/15128>
- <https://mail.python.org/archives/list/numpy-discussion@python.org/thread/F4P5UVTAKRJJ3OORI6UOWFSUEE5CNTSC/#PELUDW5ACUBHBNK5IVGWIWTQHBM2HXUP>


kgryte:
cc @ogrisel for visibility since you originally opened #722.

kgryte:
@ogrisel Do you have opinions on whether having three separate APIs is preferable to having just `argtop_k` and `top_k`?

I tried tracking down actual usages of top `k` in the wild, but I wasn't able to get a good sense on whether having only two APIs suffices or having three separate APIs is more desirable.

There may also be other combinations. The PR currently specifies

- `top_k` returns `[ values, indices ]`
- `top_k_values` returns `values`
- `top_k_indices` returns `indices`

We could, e.g., only specify 

- `top_k` returns `[ values, indices ]`
- `argtop_k` returns `indices`

or strictly complementary

- `top_k` returns `values`
- `argtop_k` returns `indices`

If you have a feel for what is preferable, that would be great to hear!

ogrisel:
Thank you very much for the survey of current implementations and API proposal. From a potential API consumer point of view the main proposal seems good for me.

About the name: `topk` seems to be a bit more popular than `top_k` in existing libraries. Using `top_k` might help reducing breaking changes when adopting this spec but it might be better to hear from library implementers.


> Should we be more strict in specifying what should happen when k exceeds the number of elements?

I believe so. As a user I would accept the call to fail with a standard exception type such as `ValueError`.

> In argmin and argmax, the specification requires returning the index of first occurrence when a minimum/maximum value occurs multiple times. Given that top_k* can be implemented as a partial sort, presumably we do not want specify a first occurrence restriction. Is this a reasonable assumption?

I think it's fine to allow topk to be faster by not enforcing this constraint. It would always be possible to add `stable=False` bool kwarg later to make it possible for the users to request stability of the results maybe at the cost of a performance penalty as done for the [xp.sort](https://data-apis.org/array-api/latest/API_specification/generated/array_api.sort.html#sort) function.

> Should we defer adding support for specifying multiple axes until a future revision of the specification or should we go ahead and add now for parity with min and max?

I wouldn't like this requirement to hurt speed of adoption by backing libraries. I don't think many users need that in practice.

> Are we okay with None being the default for axis, where the default behavior is searching over a flattened array?

I don't think users have an need for this: then can flatten the input by themselves if need.  But I have the impression that NumPy (and therefore Array API) often `axis=None` as a default convention to work on 1d flatten array anyway so I am fine with staying consistent in that regard.

However for this particular case, the default in `numpy.argpartition` is `axis=-1` instead of `axis=None`. No strong opinion of which is most natural.

> Should we defer adding support for specifying multiple axes until a future revision of the specification or should we go ahead and add now for parity with min and max?

That would remove a lot of value by preventing efficient parallelization by the underlying backend when running topk on a 2D array with `axis=0` or `axis=1`. This pattern was the original motivation for #629 (e.g. k-nearest neighbors classification in scikit-learn).

> I tried tracking down actual usages of top k in the wild, but I wasn't able to get a good sense on whether having only two APIs suffices or having three separate APIs is more desirable.

The benefit for the 3 function API would be to always be able to optimize memory usage by not allocating unnecessary arrays when k is large enough for this to matter. However I have no good feeling about how much this would really be a problem in practice (and how much the underlying implementation would be able to skip the extra contiguous memory allocation internally).


Something that seems missing from this spec is to specify the handling of NaN values. Maybe and extra kwarg is needed to specify if they should be considered as either smallest or largest, or if they need to be filtere out from the result (but then the result size would be data-dependent and potentially empty arrays which might also cause problems).

Also I assume that nan values are always smaller than +inf and larger than -inf but maybe not all libraries agree on that.

rgommers:
Revisiting this topic in preparation of helping it move forward. Quick first comment on:

> JAX has `top_k` which only returns values,

I am not sure if it changed in the meantime or you just misread the JAX docs at the time, but https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.top_k.html says that both `values` and `indices` are returned. 

I had a peek at [the jax.lax implementation](https://github.com/google/jax/blob/3fe73777192c39b1d97b190222b3c0aca2096a0c/jax/_src/lax/lax.py#L1227-L1246) as well, but could tell so quickly because I'm not familiar with the `.bind` method. @jakevdp would you mind confirming that the docs are correct here?

So it looks like PyTorch, JAX and TF all return `(values, indices)`. That seems to align well enough with the `top_k` definition proposed here.

jakevdp:
JAX's current `top_k` functions are in the `jax.lax` namespace, while the array API implementations will be in the `jax.numpy` namespace. So there is no issue with having different API conventions here.

JuliaPoo:
I don't mind drafting up an implementation for:
```py
def top_k(
    x: array,
    k: int,
    /,
    *,
    axis: Optional[int] = None,
    mode: Literal["largest", "smallest"] = "largest",
) -> Tuple[array, array]
```

If it helps further the discussion. Some notes about implementation:

I think `top_k` implementation can simply call `argpartition` internally, and we can let that decide a bunch of details about `top_k`:

> Should we be more strict in specifying what should happen when `k` exceeds the number of elements?

`argpartition` does raise a `ValueError` when that happens.

> Are we okay with `None` being the default for axis, where the default behavior is searching over a flattened array?

Currently the default for `argparitition` is `axis=-1`, which performs the partitioning along the last axis rather than flattening. I think `top_k` for now can have a default `axis=-1` similarly until consensus is reached.

> Something that seems missing from this spec is to specify the handling of NaN values.

From experimenting with `np.max` and `np.min` it seems numpy returns false for any comparison with `nan`. Since `argpartition` uses this comparison as well it doesn't seem like `argpartition` is particularly intentional in its treatment of `nan`.

In addition to the current API, following [torch.topk](https://pytorch.org/docs/stable/generated/torch.topk.html) and [tf.math.top_k](https://www.tensorflow.org/api_docs/python/tf/math/top_k), we can consider the `out` keyword to specify an output array, and the `sorted` keyword. Putting it together can look something like:

```py
def top_k(
    x: array,
    k: int,
    /,
    *,
    axis: Optional[int] = -1,
    mode: Literal["largest", "smallest"] = "largest",
    sorted: bool = True,
    out: Optional[array] = None
) -> Tuple[array, array]
```

rgommers:
Here is a PR with a draft implementation for NumPy: https://github.com/numpy/numpy/pull/26666, aligned with the `top_k` signature in this PR.

seberg:
The most sane NaN handling IMO, is to sort NaNs always to the end, which however means that if you implement `sort="desc"` or largest values here, NaNs should end up also at the end, which is opposite of what happens for ascending sort/smallest values.

The annoyance with that is that it means sort behavior diverges for asc/desc sort beyong a `[::-1]` also for unstable sort.

Of course one can just leave it unspecified here.  OTOH, I dunno how much that limits the usability.

src/array_api_stubs/_draft/searching_functions.py
```diff
@@ -137,6 +146,132 @@ def searchsorted(
...
+def top_k(
+    x: array,
+    k: int,
+    /,
+    *,
+    axis: Optional[int] = None,
+    mode: Literal["largest", "smallest"] = "largest",
+) -> Tuple[array, array]:
+    """
+    Returns the ``k`` largest (or smallest) elements of an input array ``x`` along a specified dimension.
```
ndgrigorian, line 158:
```suggestion
    Returns the values and indices of the ``k`` largest (or smallest) elements of an input array ``x`` along a specified dimension.
```
Otherwise it would be identical to the description of `top_k_values`.

src/array_api_stubs/_draft/searching_functions.py
```diff
@@ -168,6 +176,50 @@ def searchsorted(
     """
 
 
+def top_k(
+    x: array,
+    k: int,
+    /,
+    *,
+    axis: Optional[int] = None,
+    mode: Literal["largest", "smallest"] = "largest",
```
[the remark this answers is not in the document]
rgommers: I don't think we're expecting other options. String literals is simply a standard way of doing this when only certain string values are valid.

> Why is string literals better than using an enum?

Enums are awful for defining a public API. You'll need to make the enums themselves public so the users of your API can use them, meaning you will increase the size of your API for every argument with a fixed set of options you add. Keeping the API surface small and easy to understand is an important goal of this standard - it's mostly functions and a few constants and other objects.

src/array_api_stubs/_draft/searching_functions.py
```diff
@@ -168,6 +176,50 @@ def searchsorted(
     """
 
 
+def top_k(
```
rparolin, line 179: Generally this look good. Other implementations I've seen pass a DeviceContext as well but I'm not sure if we want to tackle that as part of the initial implementation.

jakevdp:
A comment: since the above comments, `jax.lax.top_k` has added an `axis` argument.

jakevdp:
The main open question here I think is what to do with `NaN` values. Since `NaN`s are sorted to the end of a list, it may make sense for `top_k` to match these semantics and return the `NaN` entries if present. This is what JAX currently does:
```python
In [1]: import jax

In [2]: import numpy as np

In [3]: x = np.array([np.nan, 1, 2, 3, np.inf, -np.inf])

In [4]: np.sort(x)[::-1][:3]  # top 3 based on NumPy sorting semantics
Out[4]: array([nan, inf,  3.])

In [5]: jax.lax.top_k(x, 3)[0]  # top 3 in JAX are equivalent
Out[5]: Array([nan, inf,  3.], dtype=float32)
```
I think it would also be reasonable to change the semantics for NaNs such that they are ignored for the purpose of `top_k`; either way I think that's something the array API should specify explicitly.

seberg:
For NumPy I lean strongly that the only good thing is to prefer an "omit" style and sorting should always place NaN at the end (which NumPy does, i.e. also for descending sorts the NaNs are at the end).

For `top_k` that means NaNs are only returned if not enough non-NaN values are present.  That said, I have doubts you want to formalize that here when it isn't formalized for sorting as well?

(And yes, this a preference fpr "omit" style rather than "propagate" style NaN handling for `top_k`/sorting, which is the default for dataframes but somewhat debatable for arrays when normal reductions default to propagating.  Maybe I strongly lean towards it because it is what we have for sorting. -- also for sorting, I could see this being an option, but it adds yet another parameter...)

EDIT: But I guess... for `partition` I actually care for it to mirror sorts.  For `top_k` it would be totally OK to differ, since we can just use `descending` partition with `-k` I guess (unstable sorting but it is).

seberg:
I don't have a very strong opinion `mode=` seems fine, but the NumPy PR was still inheriting the torch `largest=` so I think there may be a bit of bikeshed potential.
The `mode="smallest` does seem nicer to me in the final result `top_k(arr, largest=False)` seems unnecessarily negated?  If there is a preference for boolean, I think `smallest=` may be nicer (sorts use `descending=False` as well, it's reverse here, but the default is False makes sense to me).
(One thing I don't love about "mode" is that it is super generic `mode="omit_nans"` could also make sense ;).)

EDIT: Or `order="smallest"`, `values="smallest"` (maybe there is a nicer that isn't mode)?

MaanasArora:
> For NumPy I lean strongly that the only good thing is to prefer an "omit" style and sorting should always place NaN at the end (which NumPy does, i.e. also for descending sorts the NaNs are at the end).

+1 on this for NumPy for ease of implementation and consistency, though I also wonder if it is simply easier on the user end as well (or for Array API if we formalize), as users who want to get the top values in common cases seem to naturally intend to either already have no nans, or omit nans. Propagation here seems to guard against error-cases, but we then add complexity in the regular case where a user doesn't want a separate call to exclude nans (which could be conceivably quite a common preprocessing step).

> PyTorch supports returning either the smallest or largest and does so via a `largest` keyword argument. This PR chooses to name the kwarg `mode` in order to be more explicit (what does `largest=False` mean to the lay reader?)

I would also like `smallest`, but I suppose that is at odds with PyTorch. It makes sense to me to say `smallest=True` (default `False`) as it reads naturally as "top k smallest values". Maybe a string kwarg isn't a best fit here because this seems quite naturally a boolean situation (I can't imagine adding another value here, and `mode` especially suggests we could!), if we have to pick one maybe `which="smallest/largest"` (or `order`)?

betatim:
I like the consistency of sorting and top k that you get from the proposal in https://github.com/data-apis/array-api/pull/722#issuecomment-4683587548

But as a user of top k it feels somehow annoying/surprising/tedious that you will have to write code that checks that the k elements (or some of them) that you got aren't all NaN. So I'd prefer the NaNs to be at the bottom (at the opposite end compared to the top k elements).

Maybe thinking about how you'd get NaNs in the first place helps. My bias is that `top_k` gets applied to a sequence that contains distances that you just calculated. Which makes me think that you'd get NaNs because either your distance computation isn't robust enough or because you have "weird" values in the input to the distance computation. Given that there is almost always "weird" stuff in real world data makes me think that NaNs should be at the bottom? They are artefacts?

kgryte:
Re: kwarg bikeshed. I originally used "mode" as a neutral word in order to make the operation explicit: either "smallest" or "largest", rather than require users to perform mental negation that `!largest` means `smallest`. The deviation from PyTorch was intentional at the time, as other libraries had yet to add (and still haven't) a kwarg for controlling whether the min or max `k` values are returned, so we have some green field here, rather than just doing what PyTorch does, especially as we are already changing the name (`top_k` instead of `topk`).

The use of "mode" in particular came from specification language used in other APIs in the spec. Currently, the only API which has a "mode" kwarg is [qr](https://data-apis.org/array-api/latest/extensions/generated/array_api.linalg.qr.html).

As for other relevant APIs,

- [`sort`](https://data-apis.org/array-api/latest/API_specification/generated/array_api.sort.html)/[`argsort`](https://data-apis.org/array-api/latest/API_specification/generated/array_api.argsort.html): both use kwarg negation to mean the opposite (i.e., `descending = false` means "ascending").
- [`fft`](https://data-apis.org/array-api/latest/extensions/generated/array_api.fft.fft.html): uses `norm` for the normalization "mode". This one also supports more than two options, so `false`/`true` would not be appropriate.

src/array_api_stubs/_draft/searching_functions.py
```diff
@@ -177,6 +185,50 @@ def searchsorted(
...
+    x: array,
+    k: int,
+    /,
+    *,
+    axis: Optional[int] = None,
```
ev-br: I'd second Olivier's comment (https://github.com/data-apis/array-api/pull/722#issuecomment-2026954773):

> I don't think users have an need for this: then can flatten the input by themselves if need. 

As a user, I would definitely prefer to flatten myself, as opposed to getting a 1D array as an output for an nD input.

The "axis=None means ravel" default IMO makes sense for reductions which return a scalar: "give me the sum of all elements of this array which happens to be nD". 

Returning a ravelled array for an nD input is not intuitive, unexpected, and I don't think it has much precedent even in NumPy? If anything, the default for `np.sort`, `np.partition` and `np.argpartition` is all `axis=-1`, so it seems to make sense to be consistent with them --- which is the what the NumPy PR default, too https://github.com/numpy/numpy/pull/31659/

And in the Array API spec, `sort` defaults to `axis=-1`, too.

src/array_api_stubs/_draft/searching_functions.py
rgommers: A `device` keyword isn't relevant for this kind of function, only for creation functions. No other sorting or statistical functions have it. I'll resolve this comment.

ev-br:
While we bikeshed `largest={True, False}` vs `mode={"largest", "smallest"}`, I'd also consider [dask.array choice](https://docs.dask.org/en/latest/_modules/dask/array/reductions.html#topk) of using negative `k` to count from the end, as it were:

```
dask.array.topk(a, k, axis=-1, split_every=None)
     Extract the k largest elements from a on the given axis, and return them sorted from largest to smallest. If k is 
     negative, extract the -k smallest elements instead, and return them sorted from smallest to largest.
```

src/array_api_stubs/_draft/searching_functions.py
rgommers: I agree that the default should be -1, not None. This is what all array libraries do for their implementation, and `sort`/`partition` default is at least as relevant as `min`/`max`.

A code search shows that it's also somewhat regularly used in practice for `torch.topk`, with `dim=1` or `dim=-1` most often.

rgommers:
For the sorting behavior keyword:

- The Dask "negative `k`" option is clearly the worst; no clarity when reading the code, and also opens the door to harder to catch issues if `k` is computed incorrectly and goes negative.
- There won't be a third `mode` value (even if we'd want different NaN handling in the future, that's orthogonal), so `largest`/`smallest`/`mode` is primarily a readability/taste question:
   - `largest=False`: benefit is that it matches PyTorch
   - `smallest=True`: benefit is that it reads better for `top_k(..., smallest=True)`
       - However, the default value is now a negation, and people look at that as well, so is it really any better than `largest=False`?
   - `mode="largest"` is arguably the most readable of the three choices here, since there's no negation for either of the two options. 
       - The argument against it is that it is a clearly binary choice, wrapped in an extensible-looking keyword.

I think there's something to say for each of those three options, and there is no clear winner. Hence, I don't have a strong preference for any of them.

lucascolley:
I would have a minor preference for following the precedence of https://data-apis.org/array-api/latest/extensions/generated/array_api.linalg.cholesky.html in `{non_default}=False`. So `smallest=False`.

rgommers:
Discussion in today's community meeting:

- `axis` keyword: consensus on changing to a default of `-1`, and no flattening behavior
- largest/smallest/mode: no one was in favor of `largest`, and it was a toss-up between `smallest` and `mode`. Given it's really a bikeshed, we went with the PR authors choice: stay with `mode`.
- NaN sorting behavior: after a very long discussion we concluded that there was no good way to make this uniform, and there were arguments for both what JAX/PyTorch do today, for the "omit" style behavior that would be nicer for scikit-learn, and for "do whatever `sort` does" (which is explicitly unspecified). 
    - Keeping it unspecified is consistent with what we do in all other functions that rely on sorting, like `sort` and `max`
    - Introducing a solution for `nan*` functions is still pending (xref gh-621), and the argument that when we do that, it should be consistently done, including for `top_k`, carried some weight.
    - Individual libraries are free to go beyond the standard and add an extra keyword (or a separate function) to implement if they really feel strongly that they need both propagate and omit style behavior.
    - So for now, keeping it explicitly unspecified with a request to keep the behavior consistent with the underlying primitive (`sort`) seemed like the way to go.

src/array_api_stubs/_draft/searching_functions.py
```diff
@@ -177,6 +185,50 @@ def searchsorted(
...
+) -> Tuple[array, array]:
+    """
+    Returns the values and indices of the ``k`` largest (or smallest) elements of an input array ``x`` along a specified dimension.
+
+    Parameters
+    ----------
+    x: array
+        input array. Should have a real-valued data type.
+    k: int
+        number of elements to find. Must be a positive integer value.
```
seberg: I think "positive" here lead to 0 being rejected in the NumPy PR.  I don't really see a good reason to reject it?

(I admit, it's a bit annoying, because zero doesn't map to `partitioning` nothing, but it feels weird to not allow it.)

EDIT: But I guess the solution is probably to just ignore it unless everyone allows zero?

kgryte: I can remove the restriction.

seberg: FWIW, both torch and JAX support it and I would prefer it for NumPy (although it is slightly awkward which makes me care less).
I don't care too much about here as the question is about being comfortable to *prescribe* it to "conforming implementations".

kgryte: Changed to "must be a nonnegative integer".

kgryte:
**Updates**:

- Changed the default `axis` value to `-1`. `None` is no longer allowed.
- Added a note stating that, when `x` contains `NaN` values, behavior is left unspecified.
- Added a note stating that the returned indices of values which compare as equal is left unspecified. E.g., if there is a tie for the 5th largest value when `k = 5`, which index should be returned? PyTorch already states that the returned index is unstable and may change between invocations.
- Updated guidance to state that `k` must be `>= 0`, rather than `> 0`.

jakevdp (commented):
The current version looks great!

ev-br (approved):
LGTM!

I also ran smoke tests of this specification against the numpy PR, `torch.topk` and a naive implementation in array-api-strict in https://github.com/data-apis/array-api-tests/pull/438. All seems to pan out.

rgommers (approved):
LGTM too, looks like we're all happy at last, so let's give this a go - thanks everyone!

Issue #616, opened 2023-03-30
Choice of conjugation in complex inner product


lezcano:
[vecdot](https://data-apis.org/array-api/latest/API_specification/generated/array_api.vecdot.html#array_api.vecdot) is defined for complex vectors as $\langle x, y \rangle = \sum_i x_i \overline{y_i}$. This can be written as $\langle x, y \rangle = y^H x$ (note how the order of the parameters is swapped).

In PyTorch we define it as $\langle x, y \rangle = x^H y$. This notation is more common in physics, while the previous notation is more common in [complex geometry](https://en.wikipedia.org/wiki/Hermitian_manifold#Formal_definition). For a short discussion see this [wikipedia page](https://en.wikipedia.org/wiki/Sesquilinear_form#Convention).

I would argue that numerical libraries are more targeted to physics people than to pure math people, so perhaps it'd be better to stick with the physics notation?

asmeurer:
I've transferred this issue to the array-api repo.

kgryte:
What is the ask here? Is this just a question about updating the docs or something else?

lezcano:
It's a question about the definition of `vecdot` for complex inputs in the array API (should it conjugate the first or the second input?)

kgryte:
This issue has been closed by https://github.com/data-apis/array-api/pull/723. During the workgroup meeting on 30 Nov, we discussed this issue and consensus was to align the definition of the complex inner product with PyTorch and NumPy, as described in the OP. The definition previously in the spec was an oversight, and the changes have been backported to the 2022 revision of the API standard.

PR #723, opened 2023-12-14
Fix definition of complex inner product in `vecdot`

mtsokol:
Hi!

As `torch.vdot`, `torch.linalg.vecdot` and `numpy.vdot` perform conjugate on the first parameter it appears that the definition of `vecdot` in the Array API is incorrect: It performs conjugate on the second parameter.

This PR updates `vecdot` definition. WDYT?


kgryte (approved):

kgryte:
@mtsokol Thanks for creating this PR. This relates to https://github.com/data-apis/array-api/issues/616.

In the workgroup meeting on 30 Nov, we discussed https://github.com/data-apis/array-api/issues/616 and consensus from that meeting was to fix the definition of `vecdot` in the specification, as done in this PR.

The proposed change should be considered a bug fix stemming from an oversight when originally adding `vecdot` to the specification.

PR #740, opened 2024-02-02
Only require axis to be negative in vecdot and cross

### LINKED ISSUE #617: Broadcasting for linalg functions that accept an axis
Two linear algebra functions allow contraction over an arbitrary axis, [cross](https://data-apis.org/array-api/latest/extensions/generated/array_api.linalg.cross.html#array_api.linalg.cross) and [vecdot](https://data-apis.org/array-api/latest/API_specification/generated/array_api.vecdot.html#array_api.vecdot). 

These APIs currently specify: 

> x2: Must be compatible with x1 for all non-compute axes (see [Broadcasting](https://data-apis.org/array-api/latest/API_specification/broadcasting.html#broadcasting)).

as well as

> The compute axis (dimension) must not be broadcasted.

This is ambiguous however when the contracted axis is a dimension that is created by broadcasting, for instance

```
vecdot(zeros((1, 4, 5)), zeros((4, 5)), axis=0)
```

Here axis=0 applied to the first dimension would be size `1`. 

I think this case should be disallowed.

Additionally, `axis` is ambiguous. It isn't clear if it should refer to the axis before or after broadcasting:

> axis (int) – the axis (dimension) of x1 and x2 containing the vectors for which to compute the cross product. Must be an integer on the interval [-N, N), where N is the rank (number of dimensions) of the shape determined according to [Broadcasting](https://data-apis.org/array-api/latest/API_specification/broadcasting.html#broadcasting). If specified as a negative integer, the function must determine the axis along which to compute the cross product by counting backward from the last dimension (where -1 refers to the last dimension). By default, the function must compute the cross product over the last axis. Default: -1.

This is of particular interest if `axis >= 0`. 

NumPy appears to refer to the axis before broadcasting:

```py
>>> np.cross(np.zeros((3, 1, 2)), np.zeros((3,)), axis=0)
array([[[0., 0.]],

       [[0., 0.]],

       [[0., 0.]]])
```

In fact, these two arrays aren't strictly broadcast compatible. What NumPy does is move the `axis` dimension of `x1` and `x2` to the end of the arrays, then broadcasts `x1[..., 0]` and `x2[..., 0]`. Effectively:

```py
x1 = moveaxis(x1, axis, -1)
x2 = moveaxis(x2, axis, -1)
if a.shape[-1] != 3 or b.shape[-1] != 3:
        raise ValueError("incompatible dimensions for cross product")

shape = broadcast(x1[..., 0], x2[..., 0]).shape
```

In other words, the arrays should be broadcast compatible after removing `axis` from the shape (and we should have `x1.shape[axis] == x2.shape[axis] == 3`).

NumPy doesn't have `vecdot` yet, but it should obviously work the same (the only difference being the contracted axis can have any size in vecdot, not just 3, and unlike cross, in vecdot the contracted axis is removed from the resulting shape). torch.linalg.cross doesn't appear to support any broadcasting. 

My implementations of `vecdot` in `numpy.array_api` and `array-api-compat` have been using the idea that `axis` refers to the axis broadcasting and allowing an added broadcasted axis. But I think this should be changed to work like `np.cross`. The `numpy.array_api` and `array_api_compat.numpy` `cross` implementations just reuse `np.cross` and therefore use those semantics (I didn't realize til now that we weren't actually testing any broadcasting rules for `cross` in the test suite). 

This was discussed at https://github.com/data-apis/array-api-compat/pull/35#discussion_r1151717215 (CC @lezcano).

Finally, note that [tensordot](https://data-apis.org/array-api/latest/API_specification/generated/array_api.tensordot.html#array_api.tensordot) doesn't have this issue because the axes are specified for each array separately.

### LINKED ISSUE #724: Remove "use axis after broadcasting" and change it or unspecify it
@mhvk noticed [here](https://github.com/numpy/numpy/pull/25411) that the definitions here are incompatible with NumPy gufuncs which I think is an oversight that was never noted/discussed.  The "after broadcasting" looks reasonable on first sight, but TBH, doesn't really seem reasonable to me:

Assume broadcasting does something, you are specifying an axes that by definition has length 1 along the array, and also by definition has little meaning for the array which gets broadcast.

Now, is the alternative of using the non-broadcast axes better?  Maybe not, maybe it is also very unclear and any *positive* axes (i.e. where broadcasting matters) is just too strange.

But, whether you like what NumPy does or not after thinking about it, it does seem if anything at least (maybe moreso) logical than the alternative.

So, I propose to change any place where broadcasting is mentioned w.r.t. to axes handling to define it as **unspecified**.  Negative axis remain allowed, as it doesn't interfere with broadcasting.  Negative axis *beyond* the number of existing dimensions prior to broadcasting *will also* be unspecified.
The last point looks necessary to me and strengthens the argument for NumPy's (or unspecified/error) logic: `matmul([1], [[2]], axis=(-2, -1))` should **not** be valid, since it might as well be a bug that the first array is not 2-D as it should be).

EDIT: The above example should use `vecdot`, matmul doesn't work: `vecdot(1, [2], axis=-1)`

asmeurer:
Nonnegative axes and negative axes less than the smaller of the two arrays are unspecified.

This is because it is ambiguous in these cases whether the dimension should refer to the axis before or after broadcasting. Previously, the spec stated it should refer to the dimension before broadcasting, but this deviates from NumPy gufunc behavior, and results in ambiguous and confusing situations, where, for instance, the result of a the function is different when the inputs are manually broadcasted together.

Also clean up some of the cross text a little bit since the computed dimension must be exactly size 3.

Fixes #724
Fixes #617

See the discussion in those issues for more details.

I don't think this should be backported, since my vecdot implementation in np.array_api and array-api-compat already followed the spec behavior here, and will need to be updated.

### DIFF
--- a/src/array_api_stubs/_draft/linalg.py
+++ b/src/array_api_stubs/_draft/linalg.py
@@ -83,15 +83,15 @@ def cross(x1: array, x2: array, /, *, axis: int = -1) -> array:
     Parameters
     ----------
     x1: array
-        first input array. Must have a numeric data type.
+        first input array. Must have a numeric data type. The size of the axis over which the cross product is to be computed must be equal to 3.
     x2: array
-        second input array. Must be compatible with ``x1`` for all non-compute axes (see :ref:`broadcasting`). The size of the axis over which to compute the cross product must be the same size as the respective axis in ``x1``. Must have a numeric data type.
+        second input array. Must be broadcast compatible with ``x1`` along all axes other than the axis along which the cross-product is computed (see :ref:`broadcasting`). The size of the axis over which the cross product is to be computed must be equal to 3. Must have a numeric data type.
 
         .. note::
            The compute axis (dimension) must not be broadcasted.
 
     axis: int
-        the axis (dimension) of ``x1`` and ``x2`` containing the vectors for which to compute the cross product. Must be an integer on the interval ``[-N, N)``, where ``N`` is the rank (number of dimensions) of the shape determined according to :ref:`broadcasting`. If specified as a negative integer, the function must determine the axis along which to compute the cross product by counting backward from the last dimension (where ``-1`` refers to the last dimension). By default, the function must compute the cross product over the last axis. Default: ``-1``.
+        the axis (dimension) of ``x1`` and ``x2`` containing the vectors for which to compute the cross product. Should be an integer on the interval ``[-N, -1]``, where ``N`` is ``min(x1.ndim, x2.ndim)``. The function must determine the axis along which to compute the cross product by counting backward from the last dimension (where ``-1`` refers to the last dimension). By default, the function must compute the cross product over the last axis. Default: ``-1``.
 
     Returns
     -------
@@ -110,8 +110,7 @@ def cross(x1: array, x2: array, /, *, axis: int = -1) -> array:
 
     **Raises**
 
-    -   if the size of the axis over which to compute the cross product is not equal to ``3``.
-    -   if the size of the axis over which to compute the cross product is not the same (before broadcasting) for both ``x1`` and ``x2``.
+    -   if the size of the axis over which to compute the cross product is not equal to ``3`` (before broadcasting) for both ``x1`` and ``x2``.
     """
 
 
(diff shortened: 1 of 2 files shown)


asmeurer:
We could also allow nonnegative axis when the two input arrays have the same number of dimensions. In that case, there is no ambiguity for something like `axix=0` because it would refer to the same dimensions both before and after broadcasting.

leofang:
@kgryte could you remind me if this will be part of v2023?

kgryte (approved):
This LGTM and is aligned with the discussion in the workgroup meeting on February 8, 2024. Thanks, @asmeurer!

PR #741, opened 2024-02-07
Support `copy` and `device` keywords in `from_dlpack`

### LINKED ISSUE #626: to_device() -- any way to force back to host "portably?"
This came up in https://github.com/data-apis/array-api-compat/pull/40.

My particular use case is something like this:

`x.to_device("cpu")` to allow downstream library test suites to guarantee that an array is returned to host (or no-op if already there) for usage in i.e., `assert_allclose(x, expected)` where `expected` is a NumPy array (always on host) but `x` could be on some other device depending on the array lib/backend.

I think specification to that level of detail is left to the libs at the moment. CuPy doesn't even have `to_device()` at the moment let alone `cpu`, instead requiring usage of i.e., `arr.get()` to return to host, but even if they did, I'm not so sure the current array API spec would mandate that they need to allow `cpu` as a valid argument.

Maybe the decision is that we can't really require that level of detail, in which case I can at least point to this issue when justifying shims to get around it when testing. If it were possible to mandate a common call signature for returning to host someday, hopefully the benefit of that would be fairly obvious re: testing multiple backends.

leofang:
~~Blocked by #602.~~

Close #626. As discussed there, we expand the DLPack data exchange protocol to formalize copies. In particular, we allow two common needs:
1. Explicitly enable/disable data copies (through the new `copy` argument). The default is still `None` ("may copy") to preserve backward compatibility and align with other constructors (ex: `asarray`).
2. Allow copying to CPU (through the new `device` argument, plus setting `copy=True`)

In principle, arbitrary cross-device copies could be allowed too, but the consensus in #626 was that limiting to device-to-host copies is enough for now. This also avoids the needs of ~~actual~~ **exhaustive** library support and clarifying unforeseen semantics issues.

This PR requires an accompanying PR to the DLPack repo (https://github.com/dmlc/dlpack/pull/136), in order for the producer to signal if a copy has been made (through a new bit mask), so please review both together.

### DIFF
--- a/src/array_api_stubs/_draft/array_object.py
+++ b/src/array_api_stubs/_draft/array_object.py
@@ -293,7 +293,7 @@ class _array:
         *,
         stream: Optional[Union[int, Any]] = None,
         max_version: Optional[tuple[int, int]] = None,
-        dl_device: Optional[Tuple[Enum, int]] = None,
+        dl_device: Optional[tuple[Enum, int]] = None,
         copy: Optional[bool] = None,
     ) -> PyCapsule:
         """
@@ -347,7 +347,7 @@ class _array:
             if it does support that), or of a different version.
             This means the consumer must verify the version even when
             `max_version` is passed.
-        dl_device: Optional[Tuple[Enum, int]]
+        dl_device: Optional[tuple[enum.Enum, int]]
             the DLPack device type. Default is ``None``, meaning the exported capsule
             should be on the same device as ``self`` is. When specified, the format
             must be a 2-tuple, following that of the return value of :meth:`array.__dlpack_device__`.
--- a/src/array_api_stubs/_draft/array_object.py
+++ b/src/array_api_stubs/_draft/array_object.py
@@ -326,6 +326,12 @@ class _array:
             - ``> 2``: stream number represented as a Python integer.
             - Using ``1`` and ``2`` is not supported.
 
+            .. note::
+                When ``dl_device`` is provided explicitly, ``stream`` must be a valid
+                construct for the specified device type. In particular, when ``kDLCPU``
+                is in use, ``stream`` must be ``None`` and a synchronization must be
+                performed to ensure data safety.
+
             .. admonition:: Tip
                 :class: important
 
@@ -344,15 +350,21 @@ class _array:
         dl_device: Optional[Tuple[Enum, int]]
             the DLPack device type. Default is ``None``, meaning the exported capsule
             should be on the same device as ``self`` is. When specified, the format
-            must follow that of the return value of :meth:`array.__dlpack_device__`.
+            must be a 2-tuple, following that of the return value of :meth:`array.__dlpack_device__`.
             If the device type cannot be handled by the producer, this function must
-            raise `BufferError`.
+            raise ``BufferError``.
         copy: Optional[bool]
             boolean indicating whether or not to copy the input. If ``True``, the
             function must always copy (paerformed by the producer), potentially allowing
             data movement across the library (and/or device) boundary. If ``False``,
             the function must never copy. If ``None``, the function must reuse existing
             memory buffer if possible and copy otherwise. Default: ``None``.
+
+            When a copy happens, the ``DLPACK_FLAG_BITMASK_IS_COPIED`` flag must be set.
+
+            .. note::
+                If a copy happens, and if the consumer-provided ``stream`` and ``dl_device``
+                can be understood by the producer, the copy must be performed over ``stream``.
 
         Returns
         -------
@@ -413,11 +425,14 @@ class _array:
         .. code:: python
 
             try:
-                x.__dlpack__(max_version=(1, 0))
+                x.__dlpack__(max_version=(1, 0), ...)
                 # if it succeeds, store info from the capsule named "dltensor_versioned",
                 # and need to set the name to "used_dltensor_versioned" when we're done
             except TypeError:
-                x.__dlpack__()
+                x.__dlpack__(...)
+
+        This logic is also applicable to handling of the new ``dl_device`` and ``copy``
+        keywords.
 
         .. versionchanged:: 2022.12
             Added BufferError.
--- a/src/array_api_stubs/_draft/array_object.py
+++ b/src/array_api_stubs/_draft/array_object.py
@@ -293,6 +293,8 @@ class _array:
         *,
         stream: Optional[Union[int, Any]] = None,
         max_version: Optional[tuple[int, int]] = None,
+        dl_device: Optional[Tuple[Enum, int]] = None,
+        copy: Optional[bool] = False
     ) -> PyCapsule:
         """
         Exports the array for consumption by :func:`~array_api.from_dlpack` as a DLPack capsule.
@@ -339,6 +341,17 @@ class _array:
             if it does support that), or of a different version.
             This means the consumer must verify the version even when
             `max_version` is passed.
+        dl_device: Optional[Tuple[Enum, int]]
+            The DLPack device type. Default is ``None``, meaning the exported capsule
+            should be on the same device as ``self`` is. When specified, the format
+            must follow that of the return value of :meth:`array.__dlpack_device__`.
+            If the device type cannot be handled by the producer, this function must
+            raise `BufferError`.
+        copy: Optional[bool]
+            Whether or not a copy should be made. Default is ``False`` to enable
+            zero-copy data exchange. However, a user can request a copy to be made
+            by the producer (through the consumer's :func:`~array_api.from_dlpack`)
+            to move data across the library (and/or device) boundary.
 
         Returns
         -------
@@ -394,7 +407,7 @@ class _array:
                     # here to tell users that the consumer's max_version is too
                     # old to allow the data exchange to happen.
 
-        And this logic for the consumer in ``from_dlpack``:
+        And this logic for the consumer in :func:`~array_api.from_dlpack`:
 
         .. code:: python
 
@@ -409,7 +422,7 @@ class _array:
             Added BufferError.
 
         .. versionchanged:: 2023.12
-            Added the ``max_version`` keyword.
+            Added the ``max_version``, ``dl_device``, and ``copy`` keywords.
         """
 
     def __dlpack_device__(self: array, /) -> Tuple[Enum, int]:
@@ -436,6 +449,8 @@ class _array:
               METAL = 8
               VPI = 9
               ROCM = 10
+              CUDA_MANAGED = 13
+              ONE_API = 14
         """
 
     def __eq__(self: array, other: Union[int, float, bool, array], /) -> array:
--- a/src/array_api_stubs/_draft/array_object.py
+++ b/src/array_api_stubs/_draft/array_object.py
@@ -357,9 +357,8 @@ class _array:
             The v2023.12 standard only mandates that a compliant library should offer a way for
             ``__dlpack__`` to return a capsule referencing an array whose underlying memory is
             accessible to the Python interpreter (represented by the ``kDLCPU`` enumerator in DLPack).
-            If the array library does not support ``kDLCPU`` at all, the function must raise
-            ``BufferError``. If a copy must be made to enable this support but ``copy`` is set to
-            ``False``, the function must raise ``ValueError``.
+            If a copy must be made to enable this support but ``copy`` is set to ``False``, the
+            function must raise ``ValueError``.
 
             Other device kinds will be considered for standardization in a future version of this
             API standard.
--- a/src/array_api_stubs/_draft/array_object.py
+++ b/src/array_api_stubs/_draft/array_object.py
@@ -294,7 +294,7 @@ class _array:
         stream: Optional[Union[int, Any]] = None,
         max_version: Optional[tuple[int, int]] = None,
         dl_device: Optional[Tuple[Enum, int]] = None,
-        copy: Optional[bool] = None
+        copy: Optional[bool] = None,
     ) -> PyCapsule:
         """
         Exports the array for consumption by :func:`~array_api.from_dlpack` as a DLPack capsule.
--- a/src/array_api_stubs/_draft/array_object.py
+++ b/src/array_api_stubs/_draft/array_object.py
@@ -294,7 +294,7 @@ class _array:
         stream: Optional[Union[int, Any]] = None,
         max_version: Optional[tuple[int, int]] = None,
         dl_device: Optional[Tuple[Enum, int]] = None,
-        copy: Optional[bool] = False
+        copy: Optional[bool] = None
     ) -> PyCapsule:
         """
         Exports the array for consumption by :func:`~array_api.from_dlpack` as a DLPack capsule.
@@ -335,23 +335,24 @@ class _array:
                 not want to think about stream handling at all, potentially at the
                 cost of more synchronizations than necessary.
         max_version: Optional[tuple[int, int]]
-            The maximum DLPack version that the *consumer* (i.e., the caller of
+            the maximum DLPack version that the *consumer* (i.e., the caller of
             ``__dlpack__``) supports, in the form of a 2-tuple ``(major, minor)``.
             This method may return a capsule of version ``max_version`` (recommended
             if it does support that), or of a different version.
             This means the consumer must verify the version even when
             `max_version` is passed.
         dl_device: Optional[Tuple[Enum, int]]
-            The DLPack device type. Default is ``None``, meaning the exported capsule
+            the DLPack device type. Default is ``None``, meaning the exported capsule
             should be on the same device as ``self`` is. When specified, the format
             must follow that of the return value of :meth:`array.__dlpack_device__`.
             If the device type cannot be handled by the producer, this function must
             raise `BufferError`.
         copy: Optional[bool]
-            Whether or not a copy should be made. Default is ``False`` to enable
-            zero-copy data exchange. However, a user can request a copy to be made
-            by the producer (through the consumer's :func:`~array_api.from_dlpack`)
-            to move data across the library (and/or device) boundary.
+            boolean indicating whether or not to copy the input. If ``True``, the
+            function must always copy (paerformed by the producer), potentially allowing
+            data movement across the library (and/or device) boundary. If ``False``,
+            the function must never copy. If ``None``, the function must reuse existing
+            memory buffer if possible and copy otherwise. Default: ``None``.
 
         Returns
         -------
--- a/src/array_api_stubs/_draft/array_object.py
+++ b/src/array_api_stubs/_draft/array_object.py
@@ -355,10 +355,11 @@ class _array:
             raise ``BufferError``.
         copy: Optional[bool]
             boolean indicating whether or not to copy the input. If ``True``, the
-            function must always copy (paerformed by the producer), potentially allowing
-            data movement across the library (and/or device) boundary. If ``False``,
-            the function must never copy. If ``None``, the function must reuse existing
-            memory buffer if possible and copy otherwise. Default: ``None``.
+            function must always copy (performed by the producer). If ``False``, the
+            function must never copy, and raise a ``BufferError`` in case a copy is
+            deemed necessary (e.g. if a cross-device data movement is requested, and
+            it is not possible without a copy). If ``None``, the function must reuse
+            the existing memory buffer if possible and copy otherwise. Default: ``None``.
 
             When a copy happens, the ``DLPACK_FLAG_BITMASK_IS_COPIED`` flag must be set.
 
--- a/src/array_api_stubs/_draft/array_object.py
+++ b/src/array_api_stubs/_draft/array_object.py
@@ -353,6 +353,16 @@ class _array:
             must be a 2-tuple, following that of the return value of :meth:`array.__dlpack_device__`.
             If the device type cannot be handled by the producer, this function must
             raise ``BufferError``.
+
+            The v2023.12 standard only mandates that a compliant library should offer a way for
+            ``__dlpack__`` to return a capsule referencing an array whose underlying memory is
+            accessible to the Python interpreter (represented by the ``kDLCPU`` enumerator in DLPack).
+            If the array library does not support ``kDLCPU`` at all, the function must raise
+            ``BufferError``. If a copy must be made to enable this support but ``copy`` is set to
+            ``False``, the function must raise ``ValueError``.
+
+            Other device kinds will be considered for standardization in a future version of this
+            API standard.
         copy: Optional[bool]
             boolean indicating whether or not to copy the input. If ``True``, the
             function must always copy (performed by the producer). If ``False``, the
--- a/src/array_api_stubs/_draft/creation_functions.py
+++ b/src/array_api_stubs/_draft/creation_functions.py
@@ -230,8 +230,9 @@ def from_dlpack(
     device: Optional[device]
         device on which to place the created array. If ``device`` is ``None`` and ``x`` supports DLPack, the output array must be on the same device as ``x``. Default: ``None``.
 
-        The v2023.12 standard only mandates that a compliant library must offer a way for ``from_dlpack`` to create an array on CPU (using
-        a library-specific way to represent the CPU device (``kDLCPU`` in DLPack) e.g. a ``"CPU"`` string or a ``Device("CPU")`` object).
+        The v2023.12 standard only mandates that a compliant library must offer a way for ``from_dlpack`` to create an array
+        whose underlying memory is accessible to the Python interpreter (using a library-specific way to represent the CPU
+        device (``kDLCPU`` in DLPack) e.g. a ``"CPU"`` string or a ``Device("CPU")`` object).
         If the array library does not support the CPU device and needs to outsource to another (compliant) array library, it may do so
         with a clear user documentation and/or run-time warning. If a copy must be made to enable this, and ``copy`` is set to ``False``,
         the function must raise ``ValueError``.
--- a/src/array_api_stubs/_draft/creation_functions.py
+++ b/src/array_api_stubs/_draft/creation_functions.py
@@ -218,7 +218,7 @@ def eye(
 def from_dlpack(
     x: object, /, *,
     device: Optional[device] = None,
-    copy: Optional[bool] = False,
+    copy: Optional[bool] = None
 ) -> Union[array, Any]:
     """
     Returns a new array containing the data from another (array) object with a ``__dlpack__`` method.
@@ -238,7 +238,7 @@ def from_dlpack(
 
         Other kinds of devices will be considered for standardization in a future version.
     copy: Optional[bool]
-        boolean indicating whether or not to copy the input. If ``True``, the function must always copy. If ``False``, the function must never copy and must raise a ``BufferError`` in case a copy would be necessary (e.g. the producer disallows views). Default: ``False``.
+        boolean indicating whether or not to copy the input. If ``True``, the function must always copy. If ``False``, the function must never copy and must raise a ``BufferError`` in case a copy would be necessary (e.g. the producer disallows views). If ``None``, the function must reuse existing memory buffer if possible and copy otherwise. Default: ``None``.
 
         If a copy is needed, the stream over which the copy is performed must be taken from the consumer, following the DLPack protocol (see :meth:`array.__dlpack__`).
 
--- a/src/array_api_stubs/_draft/creation_functions.py
+++ b/src/array_api_stubs/_draft/creation_functions.py
@@ -231,12 +231,12 @@ def from_dlpack(
     device: Optional[device]
         device on which to place the created array. If ``device`` is ``None`` and ``x`` supports DLPack, the output array must be on the same device as ``x``. Default: ``None``.
 
-        The v2023.12 standard only mandates that a compliant library should offer a way for ``from_dlpack`` to create an array
-        whose underlying memory is accessible to the Python interpreter (represented by the ``kDLCPU`` enumerator in DLPack).
-        If the array library does not support such case at all, the function must raise ``BufferError``. If a copy must be made
-        to enable this but ``copy`` is set to ``False``, the function must raise ``ValueError``.
+        The v2023.12 standard only mandates that a compliant library should offer a way for ``from_dlpack`` to return an array
+        whose underlying memory is accessible to the Python interpreter, when the corresponding ``device`` is provided. If the
+        array library does not support such cases at all, the function must raise ``BufferError``. If a copy must be made to
+        enable this support but ``copy`` is set to ``False``, the function must raise ``ValueError``.
 
-        Other kinds of devices will be considered for standardization in a future version of this API standard.
+        Other device kinds will be considered for standardization in a future version of this API standard.
     copy: Optional[bool]
         boolean indicating whether or not to copy the input. If ``True``, the function must always copy. If ``False``, the function must never copy, and raise ``BufferError`` in case a copy is deemed necessary (e.g.  if a cross-device data movement is requested, and it is not possible without a copy). If ``None``, the function must reuse the existing memory buffer if possible and copy otherwise. Default: ``None``.
 
--- a/src/array_api_stubs/_draft/creation_functions.py
+++ b/src/array_api_stubs/_draft/creation_functions.py
@@ -240,6 +240,8 @@ def from_dlpack(
     copy: Optional[bool]
         boolean indicating whether or not to copy the input. If ``True``, the function must always copy. If ``False``, the function must never copy and must raise a ``BufferError`` in case a copy would be necessary (e.g. the producer disallows views). Default: ``False``.
 
+        If a copy is needed, the stream over which the copy is performed must be taken from the consumer, following the DLPack protocol (see :meth:`array.__dlpack__`).
+
     Returns
     -------
     out: Union[array, Any]
--- a/src/array_api_stubs/_draft/creation_functions.py
+++ b/src/array_api_stubs/_draft/creation_functions.py
@@ -228,19 +228,17 @@ def from_dlpack(
     x: object
         input (array) object.
     device: Optional[device]
-        device on which to place the created array. If ``device`` is ``None`` and ``x`` supports DLPack, the output array device must be inferred from ``x``. Default: ``None``.
+        device on which to place the created array. If ``device`` is ``None`` and ``x`` supports DLPack, the output array device must be inferred from ``x.device``. Default: ``None``.
 
         The v2023.12 standard only mandates that a compliant library must offer a way for ``from_dlpack`` to create an array on CPU (using
-        the library-chosen way to represent the CPU device - ``kDLCPU`` in DLPack - e.g. a ``"CPU"`` string or a ``Device("CPU")`` object).
-        If the compliant library does not support the CPU device and needs to outsource to another (compliant) array library, it may do so
+        a library-specifc way to represent the CPU device (``kDLCPU`` in DLPack) e.g. a ``"CPU"`` string or a ``Device("CPU")`` object).
+        If the array library does not support the CPU device and needs to outsource to another (compliant) array library, it may do so
         with a clear user documentation and/or run-time warning. If a copy must be made to enable this, and ``copy`` is set to ``False``,
         the function must raise ``ValueError``.
 
-        Other kinds of devices will be considered for standardization in a future version.
+        Other kinds of devices will be considered for standardization in a future version of this API standard.
     copy: Optional[bool]
-        boolean indicating whether or not to copy the input. If ``True``, the function must always copy. If ``False``, the function must never copy and must raise a ``BufferError`` in case a copy would be necessary (e.g. the producer disallows views). If ``None``, the function must reuse existing memory buffer if possible and copy otherwise. Default: ``None``.
-
-        If a copy is needed, the stream over which the copy is performed must be taken from the consumer, following the DLPack protocol (see :meth:`array.__dlpack__`).
+        boolean indicating whether or not to copy the input. If ``True``, the function must always copy. If ``False``, the function must never copy and must raise a ``BufferError`` in case a copy is deemed necessary (e.g. the producer disallows views). If ``None``, the function must reuse the existing memory buffer if possible and copy otherwise. Default: ``None``.
 
     Returns
     -------
--- a/src/array_api_stubs/_draft/creation_functions.py
+++ b/src/array_api_stubs/_draft/creation_functions.py
@@ -238,7 +238,7 @@ def from_dlpack(
 
         Other kinds of devices will be considered for standardization in a future version of this API standard.
     copy: Optional[bool]
-        boolean indicating whether or not to copy the input. If ``True``, the function must always copy. If ``False``, the function must never copy and must raise a ``BufferError`` in case a copy is deemed necessary (e.g. the producer disallows views). If ``None``, the function must reuse the existing memory buffer if possible and copy otherwise. Default: ``None``.
+        boolean indicating whether or not to copy the input. If ``True``, the function must always copy. If ``False``, the function must never copy, and raise ``BufferError`` in case a copy is deemed necessary (e.g.  if a cross-device data movement is requested, and it is not possible without a copy). If ``None``, the function must reuse the existing memory buffer if possible and copy otherwise. Default: ``None``.
 
     Returns
     -------
--- a/src/array_api_stubs/_draft/creation_functions.py
+++ b/src/array_api_stubs/_draft/creation_functions.py
@@ -19,6 +19,7 @@ __all__ = [
 
 
 from ._types import (
+    Any,
     List,
     NestedSequence,
     Optional,
@@ -214,7 +215,11 @@ def eye(
     """
 
 
-def from_dlpack(x: object, /) -> array:
+def from_dlpack(
+    x: object, /, *,
+    device: Optional[device] = None,
+    copy: Optional[bool] = False,
+) -> Union[array, Any]:
     """
     Returns a new array containing the data from another (array) object with a ``__dlpack__`` method.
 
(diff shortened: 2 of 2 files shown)


leofang:
Note: only commit [0083b7a](https://github.com/data-apis/array-api/pull/741/commits/0083b7a3413fe882261d932adbde97275a82481c) is new; the rest was from #602.

leofang:
cc: @rgommers @seberg @tqchen @oleksandr-pavlyk

src/array_api_stubs/_draft/array_object.py
```diff
@@ -288,7 +288,11 @@ def __complex__(self: array, /) -> complex:
         """
 
     def __dlpack__(
-        self: array, /, *, stream: Optional[Union[int, Any]] = None
+        self: array, /, *,
+        stream: Optional[Union[int, Any]] = None,
+        max_version: Optional[tuple[int, int]] = None,
+        dl_device: Optional[Tuple[Enum, int]] = None,
+        copy: Optional[bool] = False
```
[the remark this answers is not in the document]
rgommers: resolving this as a duplicate comment for now

rgommers:
Thanks Leo! Should I take over the first two commits in gh-602?

leofang:
Let's just review/merge #602 first, this branch was based on the one there.

src/array_api_stubs/_draft/array_object.py
```diff
@@ -298,45 +302,52 @@ def __dlpack__(
...
+            The DLPack device type. Default is ``None``, meaning the exported capsule
+            should be on the same device as ``self`` is. When specified, the format
+            must follow that of the return value of :meth:`array.__dlpack_device__`.
+            If the device type cannot be handled by the producer, this function must
+            raise `BufferError`.
+        copy: Optional[bool]
+            Whether or not a copy should be made. Default is ``False`` to enable
+            zero-copy data exchange. However, a user can request a copy to be made
+            by the producer (through the consumer's :func:`~array_api.from_dlpack`)
+            to move data across the library (and/or device) boundary.
```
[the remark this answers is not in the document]
leofang: The current default is "may copy", so changing to 3-way is fine, though in practice I haven't seen any copy being done in the wild, as it'd defeat the original intent of using DLPack to facilitate zero-copy exchange. Of course I can't possibly check out all use cases, so to stay on the safe side I'm fine to change.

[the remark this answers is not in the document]
leofang: Done in [8bc321c](https://github.com/data-apis/array-api/pull/741/commits/8bc321cb350d1c539f4a5e12c02d91a461939f11).

src/array_api_stubs/_draft/creation_functions.py
```diff
@@ -214,19 +215,38 @@ def eye(
...
+        the library-chosen way to represent the CPU device - ``kDLCPU`` in DLPack - e.g. a ``"CPU"`` string or a ``Device("CPU")`` object).
+        If the compliant library does not support the CPU device and needs to outsource to another (compliant) array library, it may do so
+        with a clear user documentation and/or run-time warning. If a copy must be made to enable this, and ``copy`` is set to ``False``,
+        the function must raise ``ValueError``.
+
+        Other kinds of devices will be considered for standardization in a future version.
+    copy: Optional[bool]
+        boolean indicating whether or not to copy the input. If ``True``, the function must always copy. If ``False``, the function must never copy and must raise a ``BufferError`` in case a copy would be necessary (e.g. the producer disallows views). If ``None``, the function must reuse existing memory buffer if possible and copy otherwise. Default: ``None``.
+
+        If a copy is needed, the stream over which the copy is performed must be taken from the consumer, following the DLPack protocol (see :meth:`array.__dlpack__`).
```
rgommers, line 243: `from_dlpack` is a function from the consuming library, so this sentence looks a little confusing. There is no `stream` keyword in this function, so how does this help either the user or the implementer?

rgommers (commented):
Is there a timeline issue here as well? `from_dlpack` should do nothing with the new `copy` and `device` keywords beyond forwarding them to the producer. So if the producer doesn't support that functionality yet, what happens?

src/array_api_stubs/_draft/creation_functions.py
```diff
@@ -214,19 +215,38 @@ def eye(
...
 
     Returns
     -------
-    out: array
-        an array containing the data in `x`.
+    out: Union[array, Any]
+        an array containing the data in ``x``. In the case that the compliant library does not support the given ``device`` out of box
+        and must oursource to another (compliant) library, the output will be that library's compliant array object.
```
leofang, line 249: Thinking out loud: I would like to highlight this treatment. When I wrote this, I had this use case in mind:
```python
xp = a. __array_namespace__()
# to avoid unnecessary distraction, I don't use a "cpu" string here...
b = xp.from_dlpack(a, copy=True, device=<some_object_representing_cpu>)
```
because the original issue asked if a portable `.to_device()` method is possible. This is doable, at the expense of 
1. crossing two array namespaces (since `b` may not be of the same type `xp.array` as `a`, ex: CuPy -> NumPy) 
2. the extra burden for the namespace `xp` to maintain an device object representing `kDLCPU`, which may or may not be feasible. 

Do people feel better if we insist namespace leak is not allowed? So instead we would require users (and libraries!) to explicitly use up to two namespaces, e.g.
```python
# xp1 and xp2 may or may not be the same, e.g.
#   1. xp1 = cp, xp2 = np
#   2. xp1 = xp2 = torch 
a = xp1.random.random(...)
b = xp2.from_dlpack(a, copy=True, device=another_xp2_array.device)
```
which would bring us back to the explicit N-to-N library conversion hell (ex: will we have to write this code in SciPy?!), but potentially reduce confusion at run time and avoid the two costs noted above (no namespace leak, and no redundant device object).

leofang (commented):
Just leaving two notes below.

First, in the call today we discussed one issue in this PR. When two libraries exchange data via copying, potentially across devices, there might not be a common `stream` object that can be recognized and used by both libraries. The discussed solutions involve different ways of requesting synchronous copies (see [HackMD](https://hackmd.io/EiGt1U6RQbWFR6nIznDSgw?view#Support-copy-and-device-kwargs-in-from_dlpack-Leo)).

See the 2nd note below.

src/array_api_stubs/_draft/creation_functions.py
seberg: So for the `from_dlpack` I am not sure I am too worried about `device=` ending up specific to `xp2`, there should be a mapping from an `xp2` device to a dlpack device, but the user doesn't need to know/understand it.

The important part is that `xp2` translates it to a dlpack device (or a dlpack device tuple `(device_id, device_num)`, depending on what we specify right now).

I suspect the first usage is probably `numpy.from_dlpack(obj)` or `cupy.from_dlpack(obj)` to work.  I.e. we have a reference library to use because:
* We know speed is irrelevant and we need additional capabilities (say some other library that only supports numpy/cupy). 
* This is testing code.

In those common cases the author already knows how to hardcode what they want.  (It is a bit funny that the user has to pass `device="cpu"` to allow a cross-device copy, but it seems fair.)

---

Thinking about synchronization again, I think we have two obvious choices for what `stream` means: It can be in terms of originating device or in terms of target device.

I am starting to think that defining synchronization based on the *target* device (i.e. the one passed in) may have some advantages:

* If the target is CPU, we don't have a synchronization mechanism, so it is obvious that we synchronize.
* It is at least somewhat nice to know that `dldevice=kCUDA, stream=CudaStream` always makes sense!

That means I can't specify synchronized to-device copy if the target is `CPU`.  But, I am not sure if that matters too much for Cuda because we do have the cuda managed and cuda pinned memory device types?  And those _could_ specify to use the same semantics as the CUDA one (with the caveat of maybe defaulting to full synchronization).

rgommers: > So for the `from_dlpack` I am not sure I am too worried about `device=` ending up specific to `xp2`, there should be a mapping from an `xp2` device to a dlpack device, but the user doesn't need to know/understand it.

I agree. Neither (1) nor (2) seem to be a concern, and I don't quite understand the "which would bring us back to the explicit N-to-N library conversion hell". The `device=` keyword is already specific to a particular library everywhere, e.g. all creation functions and `to_device`. So nothing changes here regarding how device specifiers work. The only way to obtain them is with `some_array.device`, and hence the two code examples given above are already essentially the same.

I think there is nothing that can be changed or is of particular concern for this topic. So I suggest we resolve this comment, and copy the synchronization-related part of @seberg's comment to a non-inline comment (it's a different topic).

rgommers:
> When two libraries exchange data via copying, potentially across devices, there might not be a common `stream` object that can be recognized and used by both libraries. The discussed solutions involve different ways of requesting synchronous copies (see [HackMD](https://hackmd.io/EiGt1U6RQbWFR6nIznDSgw?view#Support-copy-and-device-kwargs-in-from_dlpack-Leo)).

In that discussion I was perhaps missing a more precise problem statement. Also note that that HackMD link isn't public, so here is what it says for potential solutions:

@seberg suggested: _The important thing is that `stream=None` (or not passing) synchronizes if you specify device="cpu". What happens if a stream is passed, seems not to important to me the consumer passed it, so they must expect the right thing._

@leofang suggested: _piggyback on `stream=None` to request sync copy, when device is not the same. Solution 2: add `stream=-2` to signal the two devices do not have a common language (stream/queue)._

I had a look at what PyTorch does, and it's basically the same as @seberg's suggestion I believe. From https://pytorch.org/docs/master/notes/cuda.html#asynchronous-execution: _In general, the effect of asynchronous computation is invisible to the caller, because (1) each device executes operations in the order they are queued, and (2) PyTorch automatically performs necessary synchronization when copying data between CPU and GPU or between two GPUs. Hence, computation will proceed as if every operation was executed synchronously._ It goes on to explain a `non_blocking` keyword, but that's not relevant to the `kDLCPU` device (only pinned memory).

> I am starting to think that defining synchronization based on the target device (i.e. the one passed in) may have some advantages:

Since we only want to enable copying to host (where there is no stream concept) at this moment, do we really need to go into this for the purposes of getting this PR in? It looks to me like this can be deferred.

seberg:
> do we really need to go into this for the purposes of getting this PR in? It looks to me like this can be deferred.

Well, there are two things we need, I guess:
* We need to specify that `device="cpu"` without passing a stream *must* synchronize fully.
* We shouldn't have random decisions about what a passed stream and device means.  So arguably should say that a producer **must** ignore the stream object for the time being *unless* they are 100% sure that UB is impossible for some reason.  I.e. as of now it is unclear if `device="cpu", stream=123123` actually means a cuda stream and cupy must not hope it does.

Streams really should always have been fully type safe, but OK (which would resolve this, we could just say: one of them, figure it out based on type).

Maybe I should change my proposal to, this:  Streams belong to the *target* device unless the producer knows it cannot possibly belong to it (the stream is typed to be a `CudaStream`, say an integer subclass).  The important part being here that the *default* must be in terms of the target!

rgommers:
> So arguably should say that a producer **must** ignore the stream object for the time being _unless_ they are 100% sure that UB is impossible for some reason.  I.e. as of now it is unclear if `device="cpu", stream=123123` actually means a cuda stream and cupy must not hope it does.

I'd say that:
1. The description of the `stream` parameter in [`array_api.array.__dlpack__.html`](https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__dlpack__.html) seems pretty clear that the stream number is that of the consumer. 
2. Providing a stream number for a "to host" operation is nonsensical input I'd say, which we tend not to specify. If anything, let's make it even clearer that (1) is the case. However, I don't think this can be much clearer: _The ownership of the stream stays with the consumer. On CPU and other device types without streams, only `None` is accepted._

Since only `None` is accepted, changing that to now require something else (like -2) looks like an unnecessary backwards compat break. 

> We need to specify that `device="cpu"` without passing a stream _must_ synchronize fully.

That sounds good to add. Needs to avoid the string `"cpu"` and use text like "when the `device=` keyword receives an argument representing the CPU/host device".

seberg:
Well, the more I think about it the more I think it makes sense to say:

If `dl_device=` is passed the stream must refer to the specified target devices definition for streams.  For example if the `dl_device=CPU` only `stream=None` is valid and full synchronization must occur even if the data is currently on a CUDA device.
A producer may ignore the stream passed and opt to the safe default synchronization of the target device.

That covers what we need, and in principle allows a `cpu -> gpu` exporter to use a stream.  A `gpu -> cpu` can also use stream if the CPU is specified as `CUDAManaged` or `CUDAHost`.
The main thing is: It doesn't leave an odd gap in the spec.

Now there might be a point for more complex synchronization schemes in the future.  But maybe that just needs a future extension.  (The annoyance is that it would be nice to not need a try/except for such a future extension, but beggars can't be choosers)

leofang:
I see that you are continuing the discussion we had during the meeting on the subject
> When two libraries exchange data via copying, potentially across devices, there might not be a common `stream` object that can be recognized and used by both libraries.

I think Sebastian is right to conclude that 
> If `dl_device=` is passed the stream must refer to the specified target devices definition for streams.

The `stream` object and the `dl_device` object must go in as a valid pair. One subtlety is that while it is in theory possible for a D2H copy to be asynchronously performed on a stream, with the current "one-way" DLPack protocol there is no safe way to allow this. DLPack is one-way in the sense that during handshaking only the consumer would offer `stream` (and now `dl_device` too, with this PR) to the producer, not the other way around.[^1]

[^1]: It would be a two-way handshaking if both parties offer their `dl_device` and `stream` for negotiation -- then even D2H copies can be done asynchronously, by a bunch of complex logic checking device compatibility and establishing ordering. (By the way, when I developed the draft and exchanged with Sebastian offline, I somehow mistakenly assumed we had two-way handshaking, or at least a part of it -- that a consumer can query what devices the producer supports.) But the ship has sailed and I guess it'd make the implementation very complicated that it's probably not worth it, even if we could start DLPack over from scratch.

Given no major concern with this PR based on the discussions in the original issue, in the call, and above, let me proceed to make necessary changes to refine this PR.

> Also note that that HackMD link isn't public, so here is what it says ...

Interesting, I didn't know that. I thought we just need to log in HackMD to view it?

rgommers:
> or at least a part of it -- that a consumer can query what devices the producer supports.

The new introspection API should allow that, but I agree that it's probably not worth it to go there.

> I thought we just need to log in HackMD to view it?

No, HackMD has permissions. It's also transient; for any issue/PR with a relevant discussion we should be posting a summary on GitHub.

> If `dl_device=` is passed the stream must refer to the specified target devices definition for streams.

Sounds right. One thing that stood out to me related to this in this PR is that the "only CPU/host is supported" is part of the `from_dlpack` changes, but the `__dlpack__` changes are more general. Is that on purpose (if so, I don't mind)?

src/array_api_stubs/_draft/array_object.py
```diff
@@ -333,12 +341,30 @@ def __dlpack__(
...
             `max_version` is passed.
+        dl_device: Optional[Tuple[Enum, int]]
+            the DLPack device type. Default is ``None``, meaning the exported capsule
+            should be on the same device as ``self`` is. When specified, the format
+            must be a 2-tuple, following that of the return value of :meth:`array.__dlpack_device__`.
+            If the device type cannot be handled by the producer, this function must
+            raise ``BufferError``.
+        copy: Optional[bool]
+            boolean indicating whether or not to copy the input. If ``True``, the
+            function must always copy (paerformed by the producer), potentially allowing
```
rgommers, line 358: typo: performed

src/array_api_stubs/_draft/array_object.py
```diff
@@ -333,12 +341,30 @@ def __dlpack__(
...
+            data movement across the library (and/or device) boundary. If ``False``,
```
rgommers, line 359: this may be slightly misleading, because it's hard to imagine that a copy on the same device is needed to enable crossing a library boundary. Rather, it allows (a) to use in-place operations safely and (b) device transfer.

src/array_api_stubs/_draft/creation_functions.py
```diff
@@ -214,19 +215,36 @@ def eye(
...
+) -> Union[array, Any]:
     """
     Returns a new array containing the data from another (array) object with a ``__dlpack__`` method.
 
     Parameters
     ----------
     x: object
         input (array) object.
+    device: Optional[device]
+        device on which to place the created array. If ``device`` is ``None`` and ``x`` supports DLPack, the output array device must be inferred from ``x.device``. Default: ``None``.
```
rgommers, line 231: The consumer will not be able to understand the producer's device object, so `x.device` is a bit misleading here. I suggest something like: "the output array will be on the same device as `x`".

src/array_api_stubs/_draft/creation_functions.py
```diff
@@ -214,19 +215,36 @@ def eye(
...
+
+        The v2023.12 standard only mandates that a compliant library must offer a way for ``from_dlpack`` to create an array on CPU (using
+        a library-specifc way to represent the CPU device (``kDLCPU`` in DLPack) e.g. a ``"CPU"`` string or a ``Device("CPU")`` object).
```
rgommers, line 234: typo: specific

src/array_api_stubs/_draft/creation_functions.py
```diff
@@ -214,19 +215,36 @@ def eye(
...
+        If the array library does not support the CPU device and needs to outsource to another (compliant) array library, it may do so
+        with a clear user documentation and/or run-time warning. If a copy must be made to enable this, and ``copy`` is set to ``False``,
+        the function must raise ``ValueError``.
+
+        Other kinds of devices will be considered for standardization in a future version of this API standard.
+    copy: Optional[bool]
+        boolean indicating whether or not to copy the input. If ``True``, the function must always copy. If ``False``, the function must never copy and must raise a ``BufferError`` in case a copy is deemed necessary (e.g. the producer disallows views). If ``None``, the function must reuse the existing memory buffer if possible and copy otherwise. Default: ``None``.
```
rgommers, line 241: > (e.g. the producer disallows views)

Views are not a thing in the standard at all, and this can encourage a library like JAX (which does the right thing now) to start raising errors. I suggest to change this text to: _"(e.g., the producer's data doesn't natively fit DLPack's strided memory description)"_

src/array_api_stubs/_draft/creation_functions.py
```diff
@@ -214,19 +215,36 @@ def eye(
...
 
     Returns
     -------
-    out: array
-        an array containing the data in `x`.
+    out: Union[array, Any]
+        an array containing the data in ``x``. In the case that the compliant library does not support the given ``device`` out of box
```
rgommers, line 246: The "In the case ..." sentence is out of place here I think, and should be deleted. Returning a device from another library as an intermediate object from `__dlpack__` can be fine, but `from_dlpack` should only return its own array type. If it does not understand the input to `device=`, it should raise an appropriate error like `ValueError` or `TypeError` (doesn't need to be specified here).

rgommers (changes requested):
Overall this is starting to look close to ready - a few more comments.

src/array_api_stubs/_draft/array_object.py
leofang: I think copy is copy, it can do whatever this keyword indicates, especially we also are adding a `device` keyword. If 
> copy on the same device is needed to enable crossing a library boundary

is the impression that this text gives, we need to fix it. I'll come up with a longer version (I hate long text if a shorter version does an equally good job) for us to trim/revise.

src/array_api_stubs/_draft/creation_functions.py
leofang: I think this is the kind of note that is (array) library developer facing, not end user facing. During the revision of this PR, my feeling is sometimes it is hard to separate them, and sometimes we need to do a better job. Let's apply your suggestion and see what it looks like.

src/array_api_stubs/_draft/creation_functions.py
leofang: Jax shouldn't raise errors because it is not necessary. They would get an influx of user complaints if they just flip the flag out of blue. Let me come up with something better (related to `device`).

src/array_api_stubs/_draft/creation_functions.py
leofang: I am sorry, but I thought I [highlighted this issue and raised a discussion above](https://github.com/data-apis/array-api/pull/741#discussion_r1483777111), and both of you agreed it is not a problem? Let us continue in that thread, and focus on the first snippet there. 

> but `from_dlpack` should only return its own array type. If it does not understand the input to `device=`, it should raise an appropriate error like `ValueError` or `TypeError` (doesn't need to be specified here).

This is simply impossible when someone wants `cupy.from_dlpack(arr_cupy, device="cpu", copy=True)` to work, which defeats the purpose of this PR (have a way to get the array on host that is portable). In this case, the returned object has to be a `numpy.ndarray`.

src/array_api_stubs/_draft/creation_functions.py
leofang: > The only way to obtain them is with some_array.device, and hence the two code examples given above are already essentially the same.

Reading [the thread](https://github.com/data-apis/array-api/pull/741#discussion_r1485213049) below, I am not sure I agree. In order for `from_dlpack()` to allow a way to copy data to host, each library needs its own CPU identifier (which we leave unspecified, except noting in this PR that it is allowed).

> and I don't quite understand the "which would bring us back to the explicit N-to-N library conversion hell".

When I wrote this, I had in mind the downstream libraries like SciPy and scikit-learn relying on this to work. They would have to detect `xp1` namespace (which they already did), and come up with a suitable `xp2` choice that allows them to move data to CPU.

rgommers: > and come up with a suitable `xp2` choice that allows them to move data to CPU.

that's not a problem: for scipy/sklearn that is always going to be numpy. 

> each library needs its own CPU identifier (which we leave unspecified

Sorry, I meant "obtain portably". You can use `device=mylib.cpu` indeed, but that's outside the standard. Using features provided by the standard, one can write `x.device`. E.g. in test cases for SciPy:
```python
def test_func(xp):  # test fixture xp, can be any library 
    x = xp.asarray(...)
    actual = func(x)
    expected = np.asarray([0.51, 3.54])
    assert_allcllose(actual, expected)   # problem due to mismatching namespaces
```
gets modified to:
```python
def new_assert_allclose(actual, expected, ...):
    actual = np.from_dlpack(actual, expected.device)
    assert_allcllose(actual, expected)
```

I'm still not sure why your second code example would be a problem. It looks the way I expect it, and should be usable enough. The first one works fine too though.

src/array_api_stubs/_draft/creation_functions.py
rgommers: Ah okay. I think it's useful to explicitly say "implementors may choose ..." or "implementers need to ...". Or use a `.. note::` if it really needs highlighting. I'm not sure that is done 100% consistently everywhere, but in principle all API docs are for user-observable behavior and should be usable as user-facing docs.

src/array_api_stubs/_draft/creation_functions.py
leofang: fixed in b855568

src/array_api_stubs/_draft/array_object.py
leofang: fixed in b855568

src/array_api_stubs/_draft/creation_functions.py
leofang: Updated in commit 60f74396. Let me know if it is OK.

src/array_api_stubs/_draft/creation_functions.py
rgommers: LGTM now after commit b855568 - resolving.

src/array_api_stubs/_draft/array_object.py
rgommers: New version in commit b855568ae LGTM, thanks. Resolving.

rgommers:
pre-commit and Sphinx have some very minor complaints, other than that this looks pretty good to me now. I guess it'll stay in draft until there is movement on https://github.com/dmlc/dlpack/pull/136 (since this PR explicitly mentions setting a to-be-added C level flag)?

oleksandr-pavlyk:
May I suggest a change:

```
The v2023.12 standard only mandates that a compliant library must offer a \
    way for ``from_dlpack`` to create an array on CPU (using ...
```

to 

```
The v2023.12 standard only mandates that a compliant library must offer a \
    way for ``from_dlpack`` to create an array whose underlying memory is \ 
    accessible to the Python interpreter (using ...
```

I want to reiterate that while this means transferring data to colloquial "CPU", the notion of CPU device is overloaded. In oneAPI, a CPU device refers to SYCL device which adheres to SYCL execution model, memory model and programming model, and is different from host device, which refers to physical CPU without these restrictions. 

I want the verbiage used in the spec to reflect the essential requirement mandated by the standard to make data available to Python interpreter process. Opinions are welcome.

rgommers:
That looks like a good suggestion to me, thanks @oleksandr-pavlyk

src/array_api_stubs/_draft/creation_functions.py
leofang: My feeling is each of us applied some assumptions regarding portability. Let me sketch my perception.  

I left the 2nd code example (which would certainly work, I am not debating whether it would, but that it looks uglier) because for the 1st one to work it requires the highlighted escape clause (regarding cross namespaces), not because they are separate, independent considerations. 

Think of `xp` being CuPy, the 1st snippet
```python
# a is a cupy.ndarray
b = cp.from_dlpack(a, copy=True, device=cp.CPU)
# b is a numpy.ndarray
```
means `b` must be `np.ndarray`. There is no way that `b` can still be `cp.ndarray` while residing on CPU. CuPy simply disallows that.

But this is my definition of portability: When #626 was raised, it was proposed to use a method (either modifying `.to_device()` or adding `.to_numpy()`) that should always work. This PR simply says that instead of modifying a method we modify a free function. If we say the 1st snippet would not work for CuPy, it doesn't look good as far as portability is concerned, although this use case does mean that a non-portable identifier `cupy.CPU` has to be defined and used.

The other definition of portability emerges as "the `device=` argument must always be the `.device` attribute from an existing consumer array, but `from_dlpack` doesn't have to always work". Then, the slightly modified 1st snippet becomes
```python
# a is a cupy.ndarray
b = np.from_dlpack(a, copy=True, device=another_np_arr.device)
# b is a numpy.ndarray
```
and certainly it'd work without any non-portable device identifier. My complaint about its ugliness is that in a production, downstream code it means we have the N-to-N conversion to consider. I did not expect at all 
> that's not a problem: for scipy/sklearn that is always going to be numpy.

because I assumed what's discussed here could be applicable to actual array-api compliant implementation as well, not just a simple test helper. That "`xp2` is always numpy" is nothing portable either. I think we agreed NumPy should not enjoy a special status. (Also, if a torch GPU tensor is in use, in this example we should expect a torch cpu tensor is returned, not a numpy array.)

If it's all about test helpers that are aimed after in this PR and the original issue, we should not need any standardization, just ask each library to duplicate a few lines of code (from your above illustration). It's not a big deal, I am sure all libraries have done much worse hacks 😄 

@rgommers WDYT?

src/array_api_stubs/_draft/creation_functions.py
leofang: Removed in commit 4513339.

leofang:
> Is there a timeline issue here as well? `from_dlpack` should do nothing with the new `copy` and `device` keywords beyond forwarding them to the producer. So if the producer doesn't support that functionality yet, what happens?

In the later revision, this was clarified that a `try-except` should be used to check if `copy` and `device` keywords are supported.

> One thing that stood out to me related to this in this PR is that the "only CPU/host is supported" is part of the `from_dlpack` changes, but the `__dlpack__` changes are more general. Is that on purpose (if so, I don't mind)?

Indeed, yes. I was thinking we want to limit user-visible effects. If array library maintainers think they can get creative in `__dlpack__` in a way that wouldn't interfere with future evolutions of the array API standard, they can do so.

leofang:
> I guess it'll stay in draft until there is movement on [dmlc/dlpack#136](https://github.com/dmlc/dlpack/pull/136) (since this PR explicitly mentions setting a to-be-added C level flag)?

That's the part I am not sure... The contents in both the dlpack PR and this PR are cross referenced and intertwined. I am fine to go either way, once @tqchen has time to take a look and offer us feedbacks. I can also break the dlpack PR into two (one is to only add the new flag, another is updating the C/Python docs) if it helps, but I wouldn't do so unless Tianqi agrees that it would help.  

> I want the verbiage used in the spec to reflect the essential requirement mandated by the standard to make data available to Python interpreter process.

Applied in commit 338742a.

leofang:
(Requested Aaron's review from the array-api-compat perspective.)

src/array_api_stubs/_draft/creation_functions.py
rgommers: > means `b` must be `np.ndarray`. There is no way that `b` can still be `cp.ndarray` while residing on CPU. CuPy simply disallows that.

Yes, so CuPy must raise, there is no other reasonable choice. Returning a non-native array type from a user-facing API seems like an anti-pattern that has no precedent and would be very surprising. 

> But this is my definition of portability: When #626 was raised, it was proposed to use a method (either modifying `.to_device()` or adding `.to_numpy()`) that should always work. [...]

The goal in gh-626, which came from SciPy, is "make it possible for any array to be turned into a numpy array". For this, I'd expect `np.from_dlpack(...)`. Which is your second code example in the comment right above.

> My complaint about its ugliness is that in a production, downstream code it means we have the N-to-N conversion to consider.

Why do you return to N-to-N? I think the arguments you gave don't apply here. In `np.from_dlpack(a, copy=True, device=another_np_arr.device)`, the `from_dlpack` function and the `device=...` argument must be from the same library, that is all. That is how `device` keyword work in all other functions too. And it's not numpy-specific, because:

> That "`xp2` is always numpy" is nothing portable either. I think we agreed NumPy should not enjoy a special status. (Also, if a torch GPU tensor is in use, in this example we should expect a torch cpu tensor is returned, not a numpy array.)

This doesn't sound right to me. It can equally be this:
```python
# a is a cupy.ndarray
b = torch.from_dlpack(a, copy=True, device=torch.device('cpu'))
# b is a torch.Tensor on cpu
```

To me, the point of this exercise is that:
1. the library implementer `xp` accepts its own device specifier in `from_dlpack` (for now CPU-only, but can be extended later as we discussed)
2. translates that to `kDLCPU` under the hood and passes it to the producer library
3. that producer library returns a DLPack capsule with data in host memory
4. hence, if `xp` supports CPU arrays, there is _always_ a way to get that from an arbitrary library, _even if_ that library doesn't support CPU arrays natively.

There is no N-to-N problem, because what is exchanged between libraries is only the portable `kDLCPU` enum value.

rgommers: Edited my above comment for clarity. We focused on CPU, but there is even a way to go from (for example) a CuPy array on a CUDA device to a OneAPI array on a SYCL device, _via_ host memory.

leofang: > The goal in [gh-626](https://github.com/data-apis/array-api/issues/626), which came from SciPy, is "make it possible for any array to be turned into a numpy array". For this, I'd expect `np.from_dlpack(...)`. Which is your second code example in the comment right above.
> 

I have read #626 multiple times while preparing this PR, and my take was that "turned into a **numpy array**" was never the ask. As highlighted in that issue title and also noted by you later, the ask was to have something that's accessible from host (or, using Sasha's words, to the Python interpreter). Whether or not it's a NumPy array is not a  big deal. 

> Why do you return to N-to-N? 

I insist that N-to-N is a problem because we're forcing (or "encouraging", if this choice of wording helps) a downstream library code to decide what is their favorite `xp2` pick (`xp1` is from user input so they don't get to choose). This question is still unanswered. 

I have no objection/concern in how `xp1` and `xp2` in the 2nd snippet should interact and pan out in an actual implementation, I think implementation-wise and semantics-wise it is crystal clear so I don't think I have anything extra to say. My objection is why we're thinking about it at all. Let me ask this way using your example.
> This doesn't sound right to me. It can equally be this:
> ```python
> # a is a cupy.ndarray
> b = torch.from_dlpack(a, copy=True, device=torch.device('cpu'))
> # b is a torch.Tensor on cpu
> ```

 If `xp1` is cupy, how would a downstream library know that it can use torch as `xp2`? Either they insist on pre-selecting a default array library for `xp2`, or they come up with a bunch of if-else branches to encode this logic (try torch, then try jax, they try numpy, then ...). How is this justifying that the standard is array-library agnostic? 

If this question does not get across, I am afraid we will need a call to resolve this. I do not have a better, alternative way to phrase this question...

rgommers: > Whether or not it's a NumPy array is not a big deal.

agreed, it works the same for any library - as long as it supports exchanging via host memory.

> downstream library code to decide what is their favorite `xp2` pick (`xp1` is from user input so they don't get to choose

In practice this is almost always the case, but you can even avoid it (xref your `# xp1 and xp2 may or may not be the same, e.g.` example above).

> My objection is why we're thinking about it at all

Okay. The direct reason for gh-626 was that, as of now, it was not possible at all to work with two libraries that don't have a common device (e.g., NumPy and CuPy) without special-casing the library. After this PR, it will be. 

> If `xp1` is cupy, how would a downstream library know that it can use torch as `xp2`? 

It will always work with any library, on any device, as long as they can both interact with host memory. 

> Either they insist on pre-selecting a default array library for `xp2`, or they come up with a bunch of if-else branches to encode this logic (try torch, then try jax, they try numpy, then ...). How is this justifying that the standard is array-library agnostic?

No, no pre-selecting is needed, nor is an if-else block. Your code from your first comment just works:
```python
a = xp1.random.random(...)
b = xp2.from_dlpack(a, copy=True, device=another_xp2_array.device)
```

I'll make it even more explicit. Say you have no preferred library and want to write super-generic code - some function that takes two arrays that are from different libraries. Right now, you just can't do that. After this PR, you can:
```python
def func(x, y):
    xp = x.__array_namespace__()
    xp_other = y.__array_namespace__()

    # Other functions than `from_dlpack` only work if both arrays are from the same library. So if
    # `y` is from a different one than `x`, let's convert `y` into an array of the same type as `x`:
    if not xp == xp_other:
        # this always works; note that if `x.device` isn't CPU, it may incur two data copies
        # (from y.device to host on the producer side, then from host to x.device on consumer side).
        y = xp.from_dlpack(y, copy=True, device=x.device)

    # From now on use `xp.xxx` functions
```

> If this question does not get across, I am afraid we will need a call to resolve this

If the above still isn't addressing your point, let's do a quick call later today or tomorrow so we can talk through it?

tqchen:
Thanks for looping me in. Personally, I find there are two goals in here which are slightly intermingled

- G0: The ability to do zero copy exchange
- G1: The ability to cross device copy

My main thought is that we should keep API support for G0 to be simple and ensure `from_dlpack` to come with clear intent as in zero copy.

Because G1 might indeed opens up significant implementation overhead, as well as different possible ways of synchronization that I think would merit a different API(`to_device`?). 

My main question is how to phase out the API, so most libraries can start with G0. But happy to hear about others thoughts as well

seberg:
> My main question is how to phase out the API, so most libraries can start with G0

Nobody is required to implement the cross device API, we could bump minor API (so consumers can know whether or not the flag should be set).  Flagging `IS_COPY` (and a code change) is thus only needed if the library is atypical and doesn't use zero-copy exchange.

For the Array-API, this is a bit different: because I think there the intention is to *additionally* require `dl_device="cpu"` to work.  A maybe slightly overpowered way to allow comparing two libraries in tests (e.g. by converting both to the same CPU library, like NumPy).

rgommers:
> Personally, I find there are two goals in here which are slightly intermingled
> 
> * G0: The ability to do zero copy exchange
> * G1: The ability to cross device copy

That is the right framing I think indeed. 

> I think would merit a different API(to_device?).

I think a separate API would be more problematic to do, because:

1. If two libraries don't support the same devices, anything like `to_device` will not work. Asking a library like CuPy to return a DLPack with host memory for under-the-hood usage is one thing; asking it to do so in a public/user-facing API looks like a nonstarter.
2. DLPack is the only candidate for implementing this under the hood; the device support is needed and related issues like streams have been worked out. API-wise a new function not called `from_dlpack` but still doing everything through DLPack would lead to a strange asymmetry.

> so most libraries can start with G0

I think one should always start there if that's an option. `from_dlpack(y, device=x.device)` will be zero-copy if the data is already on the same device. If we really wanted to ensure always-zero-copy by default, then the `copy` keyword could default to `False` rather than `None`; that way `from_dlpack(y, device=x.device)` would never be useful and authors would have to write `from_dlpack(y, copy=None, device=x.device)` (or `copy=True`). 

Does it really matter though if this is using `from_dlpack` with the `device` keyword or some other function like `xp.from_dlpack_with_devicetransfer`? The former seems preferable, and I don't think it detracts from the "prefer zero-copy" nature of DLPack.

tqchen:
Thanks @rgommers for the comments. I think as long as packages can phase their implementation status this would be fine.

src/array_api_stubs/_draft/creation_functions.py
leofang: Thanks @rgommers. Your super-generic code snippet looks nice, I included it in the docstring. Though I think it just kick the answer to "how to pick `xp2`?" down the road. Somewhere, at some point, the choice has to be made to provide the array `y`.

Nevertheless, I think if people are happy with the PR as-is, I shouldn't be blocking for two reasons:
1. Making it strict (no namespace leak) in the beginning is better than relaxed at first and tightened later.
2. If I want to make `cupy.from_dlpack(arr_cp, copy=True, device=cupy.CPU)` work and not raise an exception (not necessary a good idea, ofc), I still can, it's just beyond what's standardized.

One minor note on the code comment:
>         # this always works; note that if `x.device` isn't CPU, it may incur two data copies
>         # (from y.device to host on the producer side, then from host to x.device on consumer side).

I think so far we've not mentioned anything about double copy, so I didn't include this comment, though it's not disallowed either. If a library wants to pick a complex route, they can.

leofang:
Thanks @tqchen @rgommers @seberg for the comments. Indeed, G0 (zero copy) has been the goal and we've achieved that since the first version of the standard. What's new with this PR is 
1. we also enable G1 (cross-device copy) using DLPack (for which Ralf gave the rationale; tl;dr is DLPack offers all the info we need to facilitate this, so no reason not use it), and 
2. use the kw arg `copy` to enforce G0 if needed (setting `copy=False` will ensure we never copy, a guarantee we couldn't have before).

leofang:
Q: Does anyone know why the doc build has a warning?
```
/home/circleci/repo/src/array_api_stubs/_draft/array_object.py:docstring of array_api_stubs._draft.array_object._array.__dlpack__:1: WARNING: py:class reference target not found: Enum
```
I thought `Enum` has been used in the return value type before and it worked just fine?

rgommers:
> I thought `Enum` has been used in the return value type before and it worked just fine?

Those warnings are a real pain. I pushed a fix. For why it was needed here and not in `__dlpack_device__` right below it, I have no explanation beyond "autodoc is buggy"

src/array_api_stubs/_draft/creation_functions.py
```diff
@@ -214,19 +214,36 @@ def eye(
...
 
     Parameters
     ----------
     x: object
         input (array) object.
+    device: Optional[device]
+        device on which to place the created array. If ``device`` is ``None`` and ``x`` supports DLPack, the output array must be on the same device as ``x``. Default: ``None``.
+
+        The v2023.12 standard only mandates that a compliant library should offer a way for ``from_dlpack`` to create an array
+        whose underlying memory is accessible to the Python interpreter (represented by the ``kDLCPU`` enumerator in DLPack).
```
rgommers, line 235: I think this is still in the wrong place. The requirement should be for `__dlpack__` to offer a way to export a DLPack capsule with data on host memory. `from_dlpack` doesn't have to allow this (we cannot mandate CuPy to return a CPU array to the user).

The sentence is fine as is, as long as `from_dlpack` is replaced by `__dlpack__` and moved to the `__dlpack__` docstring.

leofang: I think the array API users should know about this limitation too, no?

rgommers: Ah, I see what you mean. It makes sense to add something here too if you're intending to leave the option open to not support non-CPU devices as inputs. Which is one of two ways to go here.

So then:
- `__dlpack__` should say "must support kDLCPU"
- `from_dlpack` should say "must support returning an array whose underlying memory is accessible to the Python interpreter _if the library has such an array type_" and "may support non-host-memory devices in this API (if it does, data transfer may go via host memory)".

It does make it more complicated to explain, but if you don't want to make everything unconditionally work in `from_dlpack` to avoid mandating double copies in some cases, that's the price to pay.

seberg: > I think the array API users should know about this limitation too, no?

True, but maybe we need to be more specific maybe:
> As of the v2023.12 standard, a successful copy to `dl_device` is only guaranteed to be implemented when the target `dl_device` represents host (CPU) memory accessible to the Python interpreter (i.e. `np.from_dlpack(x, device="cpu")` must work; however, this doesn't apply `cp.from_dlpack(x, device="cpu")` as cupy does not have a "cpu" device)

I dunno about the host (CPU), the "accessible from the Python interpreter" is unfortunately a pretty abstract mouthful :).

seberg: Nvm. Ralfs version is shorter and better :).  I might change "must" -> "is guaranteed to".

leofang: Let me know if commit 24cb4db works for you. I don't wanna use the wording "if the library has such an array type" because it's a bit confusing. Also, I don't mention "non-host-memory" to keep things simple.

rgommers: That looks reasonable. One question:

> If the array library does not support ``kDLCPU`` at all,

Should that phrasing be removed from the `__dlpack__` description? Is there a realistic chance that there will be a library that really cannot copy data to host memory? If not, I don't think it is necessary to hedge there.

leofang: It was my attempt to mirror the `device` treatment in `from_dlpack`, but I see your point. A library does not necessarily have a `device` object representing CPU, but supporting it at the C level is no problem. Removed in commit 43263ce.

rgommers (approved):
All looks good to me now. It looks like all comments were addressed; I plan to merge this within the next day or two unless there are more comments.

leofang:
Reminder to all: Remember to review https://github.com/dmlc/dlpack/pull/136 too, which goes hand-in-hand with this PR.

Issue #645, opened 2023-07-05
to_device should not copy when the operation is a no op


asmeurer:
The spec for `to_device()` [says](https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.to_device.html#array_api.array.to_device):

> Copy the array from the device on which it currently resides to the specified device.

It doesn't say anything about the case where the given array is already on the specified device. 

Note that PyTorch's `to` has a `copy` flag, which is `False` by default, https://pytorch.org/docs/stable/generated/torch.Tensor.to.html, which prevents copying when the device is the same. to_device should probably behave the same. We may or may not want to add a `copy` flag to `to_device` as well (but note that you can control this explicitly using `asarray`).

This came up thinking about this code in scikit-learn https://github.com/scikit-learn/scikit-learn/pull/26315/files/42524bd42900d8ea5f4a334780387a72c6f9580d#diff-86c94a3ca33490c6190f488f5d40b01bf0fd29be36da0b4497ef0da1fda4148a.

rgommers:
I think this requires a clarification but not a new keyword. The case of "what if the given device to copy to is the current device" isn't covered in the current text, however I'd expect it to be done with the natural implementation (return self). In the whole design of the standard we do not allow modifying aliased memory, so it's fully equivalent to return self vs. making a data copy in memory. And making a data copy is an execution detail, so it's clear we should never specify such a thing.

leofang:
https://github.com/data-apis/array-api/issues/626#issuecomment-1855049430 discussed the consideration of adding `copy` to `from_dlpack()`. Would be nice to apply the same consideration here (to `to_device()`) too.

kgryte:
@leofang Does https://github.com/data-apis/array-api/pull/741 have any bearing here (e.g., adding a new `copy` kwarg to `to_device`)?

I think at a minimum we still want to add a clarification that the default behavior is zero-copy when the specified device matches the array's current device.

PR #742, opened 2024-02-08
Add note to `to_device` regarding copy behavior when moving to same device

kgryte:
This PR

- closes https://github.com/data-apis/array-api/issues/645 by adding a note concerning expected behavior when a specified device corresponds to the device on which an array instance resides. Namely, copy behavior is left implementation-dependent, with conforming array libraries being allowed to either return `self` or perform an explicit copy, as discussed in https://github.com/data-apis/array-api/issues/645#issuecomment-1634419029.

cc @leofang @rgommers

### DIFF
--- a/src/array_api_stubs/_draft/array_object.py
+++ b/src/array_api_stubs/_draft/array_object.py
@@ -1086,8 +1086,11 @@ class _array:
             an array with the same data and data type as ``self`` and located on the specified ``device``.
 
 
-        .. note::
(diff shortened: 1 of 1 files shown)


rgommers (approved):
LGTM, thanks @kgryte. This is a straightforward clarification, so can be merged as is I believe. A potential new `copy=` keyword can always be added later, this PR doesn't change that.

leofang (approved):

rgommers:
In it goes then, thanks both!

PR #743, opened 2024-02-08
nit: fix typo

leofang:


kgryte (approved):
Thanks for catching, @leofang!

PR #752, opened 2024-02-22
docs: clarify shape of `values` and order of `counts` in `unique` APIs

kgryte:
This PR

- closes https://github.com/data-apis/array-api/issues/588.
- is a follow-up from https://github.com/data-apis/array-api/pull/700.
- explicitly states that `values` in all `unique_*` APIs must be a one-dimensional array. Previously, the output shape was implied, but not made explicit.
- explicitly states that `indices` in `unique_all` must be the indices of a flattened `x` (ref: https://github.com/data-apis/array-api/issues/588).
- adds a sentence requiring that the order of `counts` match the order of `values`.


rgommers (approved):
LGTM. I verified that this matches the actual behavior of NumPy, and aligns with previous comments on this topic.

Issue #761, opened 2024-03-17
Autogenerated page for cholesky does not render some math symbols correctly


oleksandr-pavlyk:
This page https://data-apis.org/array-api/draft/extensions/generated/array_api.linalg.cholesky.html#array_api.linalg.cholesky renders as follows for me:

![image](https://github.com/data-apis/array-api/assets/21087696/e85a080b-4169-44ff-b7b6-e9a7ae34c889)

Notice that only the first math expression `$\mathbb{K}$` is rendered correctly, while subsequent `$\mathbb{R}$`, `$\mathbb{K}$`, and `$\mathbb{C}$` did not. 

Not sure how to fix this though.

asmeurer:
The source is using `$` instead of `:math:` for these expressions. If we want to use `$` in RST we need to add the extension sphinx-math-dollar.

PR #762, opened 2024-03-21
docs: fix equation rendering in `linalg.cholesky`

kgryte:
This PR:

- closes: https://github.com/data-apis/array-api/issues/761
- addresses an oversight when migrating from Markdown to RST by fixing equation rendering in `linalg.cholesky`.


rgommers (approved):
Thanks @kgryte

PR #763, opened 2024-03-21
feat: add `max rank` to inspection API capabilities

kgryte:
This PR

- adds support for querying the maximum number of supported dimensions (i.e., the maximum array rank).
- requires that conforming array libraries supporting an infinite number of dimensions specify the max rank as `None` (i.e., there is not maximum array rank), and, for those unable to support infinite dimensions, those libraries must specify a finite integer value.
- closes: https://github.com/data-apis/array-api/issues/694

### DIFF
--- a/src/array_api_stubs/_draft/_types.py
+++ b/src/array_api_stubs/_draft/_types.py
@@ -140,5 +140,10 @@ DataTypes = TypedDict(
     total=False,
 )
 Capabilities = TypedDict(
-    "Capabilities", {"boolean indexing": bool, "data-dependent shapes": bool}
+    "Capabilities",
+    {
+        "boolean indexing": bool,
+        "data-dependent shapes": bool,
+        "max rank": Optional[int],
+    },
 )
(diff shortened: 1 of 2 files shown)


rgommers (approved):
This LGTM, thanks @kgryte. 

I considered whether this needed to be device-specific, but unless someone has a concrete example of that being the case in an existing library, I don't think it's needed now. It's always possible to do this in the future by adding a `device` keyword to `capabilities` in the future.

leofang (approved):
Q: Is this for v2024 only, or would we backport it?

asmeurer:
New APIs shouldn't be backported. Only things like typo fixes and clarifications that don't affect implementations should be backported to previous versions.

kgryte:
As this PR has received 2 approvals and no objections, I'll go ahead and merge. Any further changes can be addressed in follow-up PRs. Thanks all!

asmeurer:
Since this isn't yet in a released version of the standard, I would like to propose that we rename "max rank" to something like "max dimensions". The term "rank" is ambiguous when referring to an array, since it can also refer to the linear algebra definition of rank (or even the [tensor definition](https://en.wikipedia.org/wiki/Tensor_(intrinsic_definition)#Tensor_rank), which is yet another thing). I know we also use the term "rank" throughout the standard, but I would also propose that we replace this with something like "dimensionality".

rgommers:
I think that's a fair comment. "max ndim" would be another option if name length matters.

rgommers:
Briefly discussed today: seemed fine to everyone to change name - @asmeurer will open a new PR.

asmeurer:
#809

Issue #383, opened 2022-01-25
Specify that dunder get/set item accepts `SupportsIndex`


honno:
Currently [`__getitem__()`](https://data-apis.org/array-api/latest/API_specification/generated/signatures.array_object.array.__getitem__.html#signatures.array_object.array.__getitem__)/[`__setitem__()`](https://data-apis.org/array-api/latest/API_specification/generated/signatures.array_object.array.__setitem__.html#signatures.array_object.array.__setitem__) accept `array` as keys, which implicitly has to be boolean arrays, or 0d integer arrays i.e. has a working [`__index__()`](https://data-apis.org/array-api/latest/API_specification/generated/signatures.array_object.array.__index__.html#signatures.array_object.array.__index__). The latter is not an obvious concept to grasp, and maybe not even quite specified anyway.

I propose we add the stdlib [`typing.SupportsIndex`](https://docs.python.org/3/library/typing.html#typing.SupportsIndex) to the get/set item signatures, specifically before `array`. This does add support for any indexable type, but that seems okay, and was the intention anyway when `__index__()` was proposed in #231. I also think a note would be helpful in regards to what arrays are accepted, e.g.:

> `key` can only be an `array` if it is valid for [boolean array indexing](https://data-apis.org/array-api/latest/API_specification/indexing.html#boolean-array-indexing), or supports [`__index__()`](https://data-apis.org/array-api/latest/API_specification/generated/signatures.array_object.array.__index__.html#signatures.array_object.array.__index__) i.e. is a 0-dimensional integer array.

cc @kgryte @asmeurer

asmeurer:
Just to be clear, the point of `__index__` isn't just to support 0d integer arrays as indices on arrays, but to let them be indices on any type that supports integer indexing like `str` and `list`. And conversely, for arrays to accept as integer indices any type that implements `__index__` (with the exception of `bool`, which unfortunately inherits it from `int` but usually has a different indexing meaning for arrays). 

>  i.e. is a 0-dimensional integer array.

I would omit this part. Any type can implement `__index__`. That's the whole point. In fact, we might link to [PEP 357](https://www.python.org/dev/peps/pep-0357/) instead of the spec `__index__` definition.

kgryte:
@honno Would you be willing to open a PR addressing this issue for the v2024 revision?

NeilGirdhar:
[...]

PR #766, opened 2024-03-21
Replace `int` with `SupportsIndex` in indexing methods hints

honno:
Resolves https://github.com/data-apis/array-api/issues/383


src/array_api_stubs/_draft/array_object.py
```diff
@@ -1089,11 +1098,13 @@ def __setitem__(
...
         self: array
             array instance.
-        key: Union[int, slice, ellipsis, Tuple[Union[int, slice, ellipsis], ...], array]
+        key: Union[SupportsIndex, slice, ellipsis, Tuple[Union[SupportsIndex, slice, ellipsis], ...], array]
             index key.
         value: Union[int, float, bool, array]
             value(s) to set. Must be compatible with ``self[key]`` (see :ref:`broadcasting`).
 
+        .. note::
+           ``key`` can only be an array if it is valid for boolean array indexing, or supports ``__index__()``.
```
honno, line 1107: I really can't figure out why this gives a warning—any playing around with notes for `__setitem__()` gives me the same thing  :thinking:

```
/home/honno/PycharmProjects/array-api/src/array_api_stubs/_draft/array_object.py:docstring of array_api_stubs._draft.array_object._array.__setitem__:1: WARNING: py:class reference target not found: ..
```

rgommers:
Thanks @honno.

Adding a quick clarifying example:
```python
>>> import numpy as np
>>> import torch
>>> 
>>> x = np.arange(1, 4)
>>> y = torch.arange(1, 4)
>>> 
>>> class Ix:
...     def __index__(self):
...         return 1
... 
>>> x[Ix()]
2
>>> y[Ix()]
tensor(2)
```

src/array_api_stubs/_draft/array_object.py
rgommers: There is `nitpick_ignore` list in `src/_array_api_conf.py` that may be needed here.

honno: Can ignore the re-review-request still need to figure this one out :sweat_smile:

rgommers:
@honno would you be able to open a PR with a test for this to `array-api-tests`? Doesn't have to be merged before this will be merged, but that will help smoke out if any known/tested library does not yet support this feature.

src/array_api_stubs/_draft/array_object.py
```diff
@@ -620,9 +622,13 @@ def __getitem__(
         ----------
         self: array
             array instance.
-        key: Union[int, slice, ellipsis, None, Tuple[Union[int, slice, ellipsis, None], ...], array]
+        key: Union[int, slice, ellipsis, None, SupportsIndex, Tuple[Union[int, slice, ellipsis, None, SupportsIndex], ...], array]
             index key.
 
+
+        .. note::
+           ``key`` can only be an array if it is valid for boolean array indexing, or supports ``__index__()``.
```
asmeurer, line 630: Presumably somewhere else we say it should support `__index__` if and only if it is a 0-D integer array. Maybe it would be clearer to just say that directly here.

honno:
> @honno would you be able to open a PR with a test for this to `array-api-tests`? Doesn't have to be merged before this will be merged, but that will help smoke out if any known/tested library does not yet support this feature.

Good shout, I opened https://github.com/data-apis/array-api-tests/pull/247 to check this all out. From the looks of it:

* NumPy (+ `array_api_strict`), PyTorch
* JAX doesn't support, has a general note to open a feature requests on indexing modes (... that's prob refering to advance integer indexing).
* cupy seems to support advance integer indexing, but doesn't play nice with 0-D integer arrays as indexes.arrays used as indices must be of integer (or boolean) type
* Dask at least accepts 0-D integer input, but `array_api_compat.dask` not playing nice with the test suite so I'll have to triage that and explore further.

src/array_api_stubs/_draft/array_object.py
honno: Yep that'd help, let me know on the wording from my latest commit.

asmeurer:
> but array_api_compat.dask not playing nice with the test suite so I'll have to triage that and explore further.

Let me know what you find out. We do run it on CI with some [skips](https://github.com/data-apis/array-api-compat/blob/main/dask-skips.txt) and [xfails](https://github.com/data-apis/array-api-compat/blob/main/dask-xfails.txt), and also the [max-examples is set to 5](https://github.com/data-apis/array-api-compat/blob/main/.github/workflows/array-api-tests-dask.yml). I haven't looked at the Dask xfails too closely, and obviously if we can remove any of those that would be great. CC @lithomas1

kgryte:
@honno Were you able to triage the Dask issues with the test suite?

asmeurer:
By the way, for implementers, the generally correct behavior is to `operator.index()` to normalize index objects that aren't one of the other supported index types like Ellipsis, slice, or array. My guess is that dask.array isn't doing that.

kgryte:
@honno Is there anything more that we need to do with this PR?

honno:
> @honno Is there anything more that we need to do with this PR?

PR I think I'm happy with, just be mindful that last I checked in March it was only NumPy and PyTorch that supported "indexables", whereas JAX/CuPy/Dask didn't, so they'd need updating to support this.

Example of what I mean by an indexable:

```python
class AwkwardIndexable:
    def __init__(self, value: int):
        self._value = value

    def __int__(self):
        raise TypeError("__int__() should not be called")

    def __index__(self):
        return self._value
``

Issue #784, opened 2024-04-04
RFC: add `diff` for computing the n-th forward difference


kgryte:
This RFC proposes the addition of a new API in the array API specification for computing the n-th forward difference along a specified dimension.

## Overview

Based on array comparison [data](https://github.com/data-apis/array-api-comparison/blob/c8bac90bc684ad5e83bf92725c4cc239f36ebe4a/signatures/mathematical-functions-elementwise/diff.md), the API is available across all major array libraries in the PyData ecosystem.

`diff` was originally discussed in https://github.com/data-apis/array-api/issues/187 as a potential standardization candidate and has been requested by downstream libraries, such as [xarray](https://github.com/data-apis/array-api/issues/187#issuecomment-1553615779).

## Prior art

- NumPy: https://numpy.org/doc/stable/reference/generated/numpy.diff.html
    - `append` and `prepend` defaults to `<no_value>`.
- CuPy: https://docs.cupy.dev/en/stable/reference/generated/cupy.diff.html
- Dask: https://docs.dask.org/en/latest/generated/dask.array.diff.html
- PyTorch: https://pytorch.org/docs/stable/generated/torch.diff.html
    - `prepend` and `append` must be tensors.
- JAX: https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.diff.html
- TensorFlow: https://www.tensorflow.org/api_docs/python/tf/experimental/numpy/diff
    - Does not support `prepend` and `append` kwargs.

## Proposal

```python
def diff(x: array, /, *, n: int = 1, axis: int = -1, prepend = Optional[array], append = Optional[array]) -> array
```

## Questions

- NumPy supports `prepend` and `append` as scalar values and subsequently wraps using `asarray`. Are we okay limiting to only arrays within the specification? Libraries, such as NumPy, would be free to accept scalars; this would just not be considered portable behavior.
- Apart from scalars, NumPy requires that `prepend` and `append` arrays match the shape of `x` except along the specified axis, thus precluding broadcasting. Is there ever a situation in which broadcasting would make sense?
- The output array must have the same dtype as the input array. Consequently, when `x` has a boolean dtype, the output array must also have a boolean dtype. Similarly, unsigned integer input arrays result in unsigned integer output arrays. Are we okay requiring that `diff` support boolean and unsigned integer dtypes? Or should we limit portable behavior to floating-point (real and complex) and signed integers?

rgommers:
> Are we okay limiting to only arrays within the specification?

[...]

PR #791, opened 2024-04-18
feat: add `diff` to specification

kgryte:
This PR

- resolves https://github.com/data-apis/array-api/issues/784 by adding `diff` to the specification.
- restricts portable behavior for `prepend` and `append` to arrays. While NumPy and others support scalars (a common use case), PyTorch does not support scalars.
- supports all numeric dtypes (real and complex floats, signed and unsigned integers).
- allows some wiggle room for array libraries supporting "exotic" dtypes, such as datetimes, by stating that the output array "should" have the same data type as the input array `x`. E.g., NumPy supports datetimes and returns timedeltas. For numeric dtypes, the output array should always have the same dtype.
- restricts `prepend` and `append` to the same data type as the input array. If we more generally allow any combination of dtypes among `x`, `prepend`, and `append`, we'd need to consider type promotion semantics. We don't have much, if any, precedent in the specification for array kwargs causing an input array to upcast, and I don't think we should establish precedent here.
- set the defaults for `prepend` and `append` to `None`. This is in contrast to NumPy which doesn't set a default value.

### DIFF
--- a/spec/draft/API_specification/utility_functions.rst
+++ b/spec/draft/API_specification/utility_functions.rst
@@ -20,3 +20,4 @@ Objects in API
 
    all
    any
+   diff
--- a/src/array_api_stubs/_draft/utility_functions.py
+++ b/src/array_api_stubs/_draft/utility_functions.py
@@ -1,4 +1,4 @@
-__all__ = ["all", "any"]
+__all__ = ["all", "any", "diff"]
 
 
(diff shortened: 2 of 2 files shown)


kgryte:
As this PR has previously been discussed during workgroup meetings, does not have any open questions, and has not received any objections, I'll go ahead and merge.

Issue #664, opened 2023-07-26
RFC: add `nextafter`


steff456:
This RFC proposes adding `nextafter` function

## Overview

Based on array comparison [data](https://github.com/data-apis/array-api-comparison/blob/main/signatures/mathematical-functions-elementwise/nextafter.md), the API is available in most array libraries. The main exception is MXNet which doesn't implement it.

## Prior art

- NumPy: https://numpy.org/doc/stable/reference/generated/numpy.nextafter.html#numpy.nextafter
- PyTorch: https://pytorch.org/docs/stable/generated/torch.nextafter.html
- TensorFlow: https://www.tensorflow.org/api_docs/python/tf/math/nextafter
- CuPy: https://docs.cupy.dev/en/stable/reference/generated/cupy.nextafter.html
- Dask: https://docs.dask.org/en/stable/generated/dask.array.nextafter.html
- JAX: https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.nextafter.html

* MXNet doesn't have an implementation for `nextafter`

## Proposal

This proposal follows similar element-wise APIs,

```
def nextafter(x1: array, x2: array, /) -> array
```

## Related

- https://github.com/data-apis/array-api/issues/187#issuecomment-1638963751
- `nextafter` is available as part of [C99](https://en.cppreference.com/w/c/numeric/math/nextafter).
- `nextafter` has equivalent APIs in the [IEEE 754](http://www.dsc.ufcg.edu.br/~cnum/modulos/Modulo2/IEEE754_2008.pdf) specification as `nextUp`/`nextDown`.
- Julia includes [`nextfloat`](https://docs.julialang.org/en/v1/base/numbers/#Base.nextfloat) and [`prevfloat`](https://docs.julialang.org/en/v1/base/numbers/#Base.prevfloat).

kgryte:
A few questions come to mind:

- Is C99 preferred here (`nextafter`/`nexttoward`) or would following IEEE 754 (`nextUp`/`nextDown`) be better?
  - In SciPy use case, IEEE 754 `nextDown` is what is desired: <https://github.com/scipy/scipy/blob/422ae7150123feb11b1deb83c6bad61ed3055251/scipy/stats/_axis_nan_policy.py#L203>
- how should type promotion behave? Should `x1` and `x2` participate in type promotion? Should `x2` be converted to `x1`'s type? If not converted, then `nextafter` is actually more like C99's `nexttoward`.

rgommers:
[...]

PR #792, opened 2024-04-18
feat: add `nextafter` to specification

kgryte:
This PR

- resolves https://github.com/data-apis/array-api/issues/664 by adding `nextafter` to the specification.
- restricts `x2` to an array for portable behavior. While NumPy and others support scalars (and this is a common use case), PyTorch only supports tensors for `x2`. In general, we are consistent throughout the specification in requiring that positional array arguments be strictly arrays. Were we to make an exception here, I'd imagine that the exception would apply equally to other arithmetic operations (e.g., `add`, `multiply`, etc) for which we have standardized functional APIs.
- requires ("should") that `x2` have the same data type as `x1` for portable behavior. Libraries are free to support other data types and subsequently type promotion (e.g., NumPy), but this should not be guaranteed across conforming array libraries, as the general expectation is that one typically wants the next representable value in the **same** precision as `x1`.
- follows [C99](https://en.cppreference.com/w/c/numeric/math/nextafter) (checked with NumPy) in ensuring consistent behavior around zero. Namely, when `x1` and `x2` are equal and both zero, the result should be `x2`.

### DIFF
--- a/spec/draft/API_specification/elementwise_functions.rst
+++ b/spec/draft/API_specification/elementwise_functions.rst
@@ -66,6 +66,7 @@ Objects in API
    minimum
    multiply
    negative
+   nextafter
    not_equal
    positive
    pow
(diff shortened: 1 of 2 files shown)


kgryte:
As this PR has received approval and no objections have been raised, will go ahead and merge. Should changes need to be made, we have time to correct before we cut the 2024 release.

Issue #598, opened 2023-02-14
RFC: add support for computing the cumulative product to the standard


steff456:
This RFC requests to include a new API in the array API specification for the purpose of computing the cumulative product.

## Overview

Based on array comparison [data](https://github.com/data-apis/array-api-comparison/blob/7ecfa9867ec43a28e914b35868c041443d31d772/data/common_apis.csv#L42), the API is available in the majority of the libraries in the PyData ecosystem.

## Prior art

- NumPy: <https://numpy.org/doc/stable/reference/generated/numpy.cumprod.html>
- PyTorch: https://pytorch.org/docs/stable/generated/torch.cumprod.html
- MXNet: <https://mxnet.apache.org/versions/master/api/python/docs/api/np/generated/mxnet.np.cumprod.html>
- TensorFlow: <https://www.tensorflow.org/api_docs/python/tf/math/cumprod>

## Proposal:

```python
def cumprod(x: array, /, *, axis: Optional[int] = None, dtype: Optional[dtype] = None) -> array
```

- `dtype` kwarg is for consistency with `sum` et al

cc @oleksandr-pavlyk

kgryte:
@steff456 for the `axis` kwarg, this should be limited to just an `int` and not a `Tuple`. Based on the various docs, libraries seem to only support an `int`.

kgryte:
One thing to note is that PyTorch, e.g., requires an axis argument and does not support NumPy's default behavior of computing the cumulative sum over the flattened array.

PR #793, opened 2024-04-18
feat: add `cumulative_prod` specification

kgryte:
This PR

- resolves https://github.com/data-apis/array-api/issues/598 by adding `cumulative_prod` to the specification for computing the cumulative product.
- follows `cumulative_sum` in matching `axis`, `dtype`, and `include_initial` semantics.
- as with `cumulative_sum`, this specification differs from NumPy in that flattening is not the default behavior. Similar to PyTorch, an `axis` argument is required if the input array has more than one dimension.

### DIFF
--- a/spec/draft/API_specification/statistical_functions.rst
+++ b/spec/draft/API_specification/statistical_functions.rst
@@ -18,6 +18,7 @@ Objects in API
    :toctree: generated
    :template: method.rst
 
+   cumulative_prod
    cumulative_sum
    max
    mean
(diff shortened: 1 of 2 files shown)


asmeurer:
#797 is relevant here too.

kgryte:
**Update**: I've added a note similar to what's suggested for `cumulative_sum` in https://github.com/data-apis/array-api/pull/851.

kgryte:
As this PR has not received any objections since April 18, is consistent with `cumulative_sum`, and has been previously discussed during workgroup meeting, I'll go ahead and merge. Any further changes can be addressed in follow-up PRs.

PR #798, opened 2024-04-24
Add missing return type hint to tile()

asmeurer:


kgryte:
@asmeurer Thanks for catching. Would you mind also adding to the draft specification?

PR #800, opened 2024-04-30
Fix doc-requirements.txt to work with conda

asmeurer:
pip treats dashes and underscores the same but conda does not. The package names here are actually dashes (pip just replaces an underscore with a dash) so this is more correct anyways.


kgryte (approved):
LGTM. Thanks, @asmeurer.

Issue #796, opened 2024-04-22
Being clearer about use of RFC 2119


asmeurer:
We use [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119) for words like "should" and "must" in the standard. But many people are confused by this and think that "should" restrictions are actually "must" restrictions. This is most common for dtype restrictions, but it appears in other places in the standard as well.

We need to be clearer about this somehow. Some suggestions

- Capitalize (as is the common convention) or bold these words (https://github.com/data-apis/array-api/issues/397)
- Add a section somewhere making it clear that we are following the RFC. There is a link to the RFC at the bottom of https://data-apis.org/array-api/latest/purpose_and_scope.html but that's it. The RFC itself says:
  > Authors who follow these guidelines
   should incorporate this phrase near the beginning of their document:
  >
  >>   The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL
      NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and
      "OPTIONAL" in this document are to be interpreted as described in
      RFC 2119.
- The RFC also lists "required", "shall", "recommended", "may", and "optional". We should review our use of those terms to make sure they match the RFC.
- We might be using "should" in places where we should be using "may" or "optional". From [the RFC](https://datatracker.ietf.org/doc/html/rfc2119): SHOULD "means that there may exist valid reasons in particular circumstances to ignore a particular item, but the full implications must be understood and carefully weighed before choosing a different course", but MAY "means that an item is truly optional". Things like data type restrictions are probably correct in using SHOULD, but things like indexing restrictions, like out-of-bounds slices" perhaps ought to use MAY, since the only reason they are limited is so the the standard can support libraries that can't implement them. 

- We should consider if there are other ways things could be made clearer, for instance, explicitly stating "other behavior is unspecified".

Micky774:
> Add a section somewhere making it clear that we are following the RFC. 

IMO this is probably the most impactful change that could be made immediately. I failed to discover the reference at the bottom of the document.

Other than that, I think clarifying changelog entries a bit would help, e.g. for the v2023.12 API, one of the entries is:
> `fft.fft`: require the input array to have a complex-valued floating-point data type and require that the output array have the same data type as the input array

The term "require" implies less freedom than the term "should" is meant to represent, and would perhaps an alternative would be "clarify that the input array should have a complex..."

seberg:
[...]

PR #801, opened 2024-04-30
Making RFC 2119 usage clearer

asmeurer:
I've

- Added text to the end of the term definitions indicating that RFC 2119 is used.
- Add a Sphinx extension that automatically bolds RFC 2119 words.

However, note that our usage of some of the RFC terms is not always consistent, especially for less used terms like "required", "may", "optional", and "recommended". So I would suggest that we either

- Audit our use of these terms and fix them to align with RFC 2119 usage.
- Add some text to clarify that certain terms are not used in accordance with RFC 2119.

For example, we use the term "optional" to refer to "optional parameters". 

Fixes https://github.com/data-apis/array-api/issues/796

### DIFF
--- a/spec/2023.12/API_specification/data_types.rst
+++ b/spec/2023.12/API_specification/data_types.rst
@@ -9,35 +9,38 @@ A conforming implementation of the array API standard must provide and support
 the following data types ("dtypes") in its array object, and as data type
 objects in its main namespace under the specified names:
 
-+--------------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
-| dtype object | description                                                                                                                                                                                |
-+==============+============================================================================================================================================================================================+
-| bool         | Boolean (``True`` or ``False``).                                                                                                                                                           |
-+--------------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
-| int8         | An 8-bit signed integer whose values exist on the interval ``[-128, +127]``.                                                                                                               |
-+--------------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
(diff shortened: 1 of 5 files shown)


kgryte:
Thanks, @asmeurer, for opening the draft PR. If we are going to try to auto-bold, this should only apply to the documents under "API Specification" and "Extensions". Much of the other content is narrative text and should not fall under RFC 2119.

kgryte:
Also, swapping out the rst table for a list table seems unrelated to the changes described in the OP. That may be better left to a small separate PR to keep things clean.

kgryte:
One last addendum: the only words we apply from RFC 2119 are **must** and **should**. I don't believe we have a written record of this, but those were the only two words which we sought to apply when originally considering RFC 2119.

asmeurer:
> Also, swapping out the rst table for a list table seems unrelated to the changes described in the OP. That may be better left to a small separate PR to keep things clean.

There was a build error related to this table. I think adding the bold messed up the formatting of the table.

asmeurer:
We should consider that if we say that we follow RFC 2119, then it would be better to follow it consistently, meaning in all documents, and also follow it completely, meaning the full set of words. If we don't do this, then we need to make it very clear that we aren't, but it would be clearer to people familiar with the RFC conventions if we did.

kgryte:
> We should consider that if we say that we follow RFC 2119, then it would be better to follow it consistently, meaning in all documents, and also follow it completely, meaning the full set of words. If we don't do this, then we need to make it very clear that we aren't, but it would be clearer to people familiar with the RFC conventions if we did.

We'll need to update quite a bit of copy then, as a fair amount of copy when discussing, e.g., the test suite or methodology or various design topics is not actually normative.

kgryte:
I think, as a first pass, if we can limit to only the specification directories ("API Specification" and "Extensions"), that would be preferable. Otherwise, the non-normative content becomes confusing in its own right.

Issue #790, opened 2024-04-18
RFC: add `reciprocal` for computing the element-wise reciprocal


kgryte:
This RFC proposes the addition of a new API in the array API specification for computing the element-wise reciprocal.

## Overview

Based on array comparison [data](https://github.com/data-apis/array-api-comparison/blob/763ee61e4ce8d85c5431eee28d20b46a74626699/signatures/mathematical-functions-elementwise/reciprocal.md), the API is available across all major array libraries in the PyData ecosystem.

`reciprocal` was originally identified in https://github.com/data-apis/array-api/issues/187 as a potential standardization candidate and has been mentioned by downstream libraries (see [Intheon](https://github.com/data-apis/array-api/discussions/769) case study).

## Prior art

- NumPy: <https://numpy.org/doc/stable/reference/generated/numpy.reciprocal.html>
  - Intended for floating-point data types.
- CuPy: <https://docs.cupy.dev/en/stable/reference/generated/cupy.reciprocal.html>
- Dask: <https://docs.dask.org/en/stable/generated/dask.array.reciprocal.html>
- PyTorch: <https://pytorch.org/docs/stable/generated/torch.reciprocal.html>
  - Supports integral dtypes with internal promotion.
- TensorFlow: <https://www.tensorflow.org/api_docs/python/tf/math/reciprocal>
- JAX: <https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.reciprocal.html>

## Proposal

```python
def reciprocal(x: array, /) -> array
```

## Questions

- While `1/x` may be appear to be a trivial operation not warranting a dedicated API, there can be benefits should the API be mapped to dedicated [hardware instructions](https://ieeexplore.ieee.org/document/5438691). One can argue that this is best dealt with at the compiler level and end users should not be concerned with potential micro-optimizations. However, for larger datasets, there can be real-world benefits to explicit avoidance of division operations and array libraries may choose to avoid relying on compiler magic to implement polynomial approximations to the reciprocal. Given that this is implemented by all considered array libraries and straightforward to implement (similar to the standard already including `negative` and `positive`), are there any reasons to **not** include this API?
- Given that the output data type will be a floating-point data type, are we okay limiting (i.e., via the use of "should") input dtypes to floating-point? Libraries would be free, of course, to explicitly support other dtypes (e.g., as in PyTorch and TensorFlow), but that would not be guaranteed to be portable.

rgommers:
[...]

PR #802, opened 2024-05-02
feat: add `reciprocal` to the specification

kgryte:
This PR

- resolves https://github.com/data-apis/array-api/issues/790 by adding support for computing the reciprocal to the specification.
- requires that special cases be handled as if implemented as `1.0 / x`.
- adopts the same "type promotion" language as other unary element-wise APIs, which is based on the premise that an input array promotes to itself. If this language should be changed, we should do as part of a follow-up PR across all relevant element-wise APIs.
- requires ("should") that the input array have a floating-point data type, as the output array should have a floating-point data type and promotion of mixed kinds is left unspecified.

### DIFF
--- a/spec/draft/API_specification/elementwise_functions.rst
+++ b/spec/draft/API_specification/elementwise_functions.rst
@@ -70,6 +70,7 @@ Objects in API
    positive
    pow
    real
+   reciprocal
    remainder
    round
    sign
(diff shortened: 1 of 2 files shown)


kgryte:
As this addition was previously discussed during workgroup meetings, consensus was in favor of its conclusion--the API having been implemented across array libraries--and no objections have been made, I will go ahead and merge. This will allow test suite and compat work to move forward to allow us to snuff out any potential edge cases before 2024 revision finalization.

Issue #794, opened 2024-04-18
RFC: add `count_nonzero` for counting the number of "non-zero" values


kgryte:
This RFC proposes a new addition to the array API specification for counting the number of "non-zero" (i.e., truthy) values in an array.

## Overview

Based on array comparison [data](https://github.com/data-apis/array-api-comparison/blob/01d47bde30f19ea4b2afa54f4c58ad62ab56e778/signatures/statistics/count_nonzero.md), the API is available across all major array libraries in the PyData ecosystem.

`count_nonzero` was originally identified in https://github.com/data-apis/array-api/issues/187 as a potential standardization candidate and has usage within downstream libraries (e.g., [sklearn](https://github.com/search?q=repo%3Ascikit-learn%2Fscikit-learn%20count_nonzero&type=code), [SciPy](https://github.com/search?q=repo%3Ascipy%2Fscipy%20count_nonzero&type=code)).

## Prior art

- NumPy: <https://numpy.org/doc/stable/reference/generated/numpy.count_nonzero.html>
- CuPy: <https://docs.cupy.dev/en/stable/reference/generated/cupy.count_nonzero.html>
  - does not support `keepdims`.
- Dask: <https://docs.dask.org/en/stable/generated/dask.array.count_nonzero.html>
  - does not support `keepdims`.
- JAX: <https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.count_nonzero.html>
- PyTorch: <https://pytorch.org/docs/stable/generated/torch.count_nonzero.html>
  - does not support `keepdims` and was not discussed upon initial addition ([ref](https://github.com/pytorch/pytorch/issues/23907)).
- TensorFlow: <https://www.tensorflow.org/api_docs/python/tf/math/count_nonzero>

## Proposal

```python
def count_nonzero(x: array, /, *, axis: Optional[Union[int, Tuple[int, ...]]] = None, keepdims: bool = False) -> array
```

- When `axis` is `None`, the function should count the number of non-zero elements along a flattened array.
- The function should return an array having the default index data type.

## Questions

- In contrast to `sum` and other reductions, support for `keepdims` is less common among array libraries. Why this is the case is not clear. Are there any reasons why `keepdims` should not be standardized?

rgommers:
@asmeurer can you tell us if it's easy to work around a missing `keepdims` keyword in `array-api-compat`?

rgommers:
[...]

PR #803, opened 2024-05-02
feat: add `count_nonzero` to specification

kgryte:
This PR

- resolves https://github.com/data-apis/array-api/issues/794 by adding support for counting the number of non-zero values to the specification.
- is consistent with statistical functions (such as `max`, `mean`, `sum`, etc) in supporting one or more axes over which to perform the reduction and in supporting `keepdims` to ensure the returned result is broadcast compatible.
- ~~follows `nonzeros` (ref: https://github.com/data-apis/array-api/pull/23#issuecomment-680784657) in disallowing zero-dimensional input arrays. Note, however, that the statistical function APIs do not contain any explicit guidance regarding minimum rank.~~ **Update**: based on discussions during the May 2, 2024 workgroup meeting, consensus was to allow 0d input arrays. The proposed specification has been updated accordingly.
- updates `nonzeros` by moving the notes to a dedicated notes section (i.e., below the "fold").

### DIFF
--- a/spec/draft/API_specification/searching_functions.rst
+++ b/spec/draft/API_specification/searching_functions.rst
@@ -22,6 +22,7 @@ Objects in API
 
    argmax
    argmin
+   count_nonzero
    nonzero
    searchsorted
    where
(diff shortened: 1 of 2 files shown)


kgryte:
As this PR has not received any objections since May 1 and has previously been discussed in workgroup meetings, I'll go ahead and merge so that we can begin test suite and compat work. Any further changes can be addressed in follow-up PRs.

PR #806, opened 2024-05-15
Fixes to the array-api stubs

asmeurer:
These are necessary for the test suite.


rgommers (approved):
This all seems very reasonable to me, and doesn't change the meaning of any spec. So I don't think it needs discussion - going ahead and merge. Thanks @asmeurer

PR #809, opened 2024-05-30
feat!: rename "max rank" to "max dimensions" in `capabilities()`

asmeurer:
See https://github.com/data-apis/array-api/pull/763#issuecomment-2138355341. "Rank" is an ambiguous term and "dimensions" is clearer. Although other names like "max ndims" would also be OK. 

This is an unreleased API, and it's new in all implementations, so there should not be a concern with renaming it.


kgryte (approved):
LGTM. This change was discussed in workgroup meetings and was considered any improvement over the current state. And as no other suggestions for the rename were proposed, `max dimensions` wins out by default. Thanks, @asmeurer!

Issue #807, opened 2024-05-15
Python scalars in elementwise functions


shoyer:
The array API supports Python scalars in arithmetic only, i.e., operations like `x + 1`.

For the same readability reasons that supporting scalars in arithmetic is valuable, it would nice to also support Python scalars in other elementwise functions, at least those that take multiple arguments like `maximum(x, 0)` or `where(y, x, 0)`.

rgommers:
Hmm,I am in two minds about reconsidering this choice.

On the con side: non-array input to functions is going against the design we have had from the start, it makes static typing a bit harder (we'd need both an `Array` protocol and an `ArrayOrScalar` union), and not all libraries support it yet - PyTorch in particular. E.g.:
```python
>>> import torch
>>> t = torch.ones(3)
>>> torch.maximum(t, 1.5)
...
TypeError: maximum(): argument 'other' (position 2) must be Tensor, not float
```
In principle PyTorch is fine with adding this it looks like, but it's a nontrivial amount of work and no one is working on it as far as I know: https://github.com/pytorch/pytorch/issues/110636. PyTorch does support it in functions matching operators (e.g., `torch.add`) and in `torch.where`.

TensorFlow also doesn't support it (except for in their `experimental.numpy` namespace IIRC), but that's less relevant now since it doesn't look like they're going to implement anything.

> For the same readability reasons that supporting scalars in arithmetic is valuable

The readability argument is less prominent for functions that for operators though. Both because `x + 1` is very short so the relative increase in characters is worse than for function calls (since `modname.funcname` is already long). Plus scalars are less commonly used in function calls.

* * *

On the pro side: I agree that it is pretty annoying to get right in a completely portable and generic way. In the cases where one does need it, the natural choice of `asarray(scalar)` often doesn't work, it should also use the dtype and device. So `xp.maximum(x, 1)` becomes:
```python
xp.maximum(x, xp.asarray(1, dtype=x.dtype, device=x.device))
```
Hence if this is a pattern that a project happens to need a lot, it will probably create a utility function like:
```python
def as_zerodim(value, x, /, xp=None):
    if xp is None:
        xp = array_namespace(x)
    return xp.asarray(value, dtype=x.dtype, device=x.device)


# Usage:
xp.maximum(x, as_zerodim(1, x))
```

*** 

[...]

PR #811, opened 2024-06-04
feat!: allow clip to have int min or max when x is floating-point

asmeurer:
This is for consistency with operators, which allow combining an int with an array that has a floating-point data type.

See the discussion at #807. 

I also mentioned this at the meeting last week.

### DIFF
--- a/src/array_api_stubs/_draft/elementwise_functions.py
+++ b/src/array_api_stubs/_draft/elementwise_functions.py
@@ -806,7 +806,7 @@ def clip(
 
     - If both ``min`` and ``max`` are ``None``, the elements of the returned array must equal the respective elements in ``x``.
     - If a broadcasted element in ``min`` is greater than a corresponding broadcasted element in ``max``, behavior is unspecified and thus implementation-dependent.
-    - If ``x`` and either ``min`` or ``max`` have different data type kinds (e.g., integer versus floating-point), behavior is unspecified and thus implementation-dependent.
+    - If ``x`` has an integer data type and either ``min`` or ``max`` is floating-point, behavior is unspecified and thus implementation-dependent.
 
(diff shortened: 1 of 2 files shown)


asmeurer:
I also made it clearer that if all three arguments are arrays then there is a three-way broadcast.

src/array_api_stubs/_draft/elementwise_functions.py
```diff
@@ -806,7 +806,7 @@ def clip(
 
     - If both ``min`` and ``max`` are ``None``, the elements of the returned array must equal the respective elements in ``x``.
     - If a broadcasted element in ``min`` is greater than a corresponding broadcasted element in ``max``, behavior is unspecified and thus implementation-dependent.
-    - If ``x`` and either ``min`` or ``max`` have different data type kinds (e.g., integer versus floating-point), behavior is unspecified and thus implementation-dependent.
+    - If ``x`` has an integer data type and either ``min`` or ``max`` is floating-point, behavior is unspecified and thus implementation-dependent.
```
rgommers, line 809: This isn't just about scalars, as written. What if `x` is a floating-point array, and `min` an integer one? Seems ill-defined, and in `numpy` it upcasts right now:
```python
>>> np.clip(np.ones((3,), dtype=np.float32), a_min=np.zeros((2, 1), dtype=np.int32), a_max=1.5).dtype
dtype('float64')
```

rgommers (commented):
Thanks @asmeurer. A question about the dtype change inline.

> I also made it clearer that if all three arguments are arrays then there is a three-way broadcast.

This change seems reasonable to me, and I verified that `numpy` behaves that way. It may be worth including this in the test suite to ensure this actually is the case for all tested implementations?

PR #813, opened 2024-06-05
docs: specify the behavior of `clip()` when one of the operands is NaN

asmeurer:
This follows the NumPy behavior, where the result is nan if any operand is nan. I checked PyTorch and it seems to do this as well.


rgommers (approved):
That looks very reasonable to me, thanks Aaron.

kgryte (approved):
LGTM.

PR #814, opened 2024-06-13
Clarify that `clip` behavior is undefined when `min` or `max` is outside the bounds of `x`

asmeurer:
As discussed in today's consortium meeting.

See the discussion at numpy/numpy#24976.

### DIFF
--- a/src/array_api_stubs/_2023_12/elementwise_functions.py
+++ b/src/array_api_stubs/_2023_12/elementwise_functions.py
@@ -805,6 +805,7 @@ def clip(
 
     - If both ``min`` and ``max`` are ``None``, the elements of the returned array must equal the respective elements in ``x``.
     - If a broadcasted element in ``min`` is greater than a corresponding broadcasted element in ``max``, behavior is unspecified and thus implementation-dependent.
+    - If ``x`` has an integral data type and a broadcasted element in ``min`` or ``max`` is outside the bounds of the data type of ``x``, the behavior is unspecified and thus implementation-dependent.
     - If ``x`` and either ``min`` or ``max`` have different data type kinds (e.g., integer versus floating-point), behavior is unspecified and thus implementation-dependent.
 
     .. versionadded:: 2023.12
--- a/src/array_api_stubs/_2023_12/elementwise_functions.py
+++ b/src/array_api_stubs/_2023_12/elementwise_functions.py
@@ -805,7 +805,7 @@ def clip(
 
     - If both ``min`` and ``max`` are ``None``, the elements of the returned array must equal the respective elements in ``x``.
     - If a broadcasted element in ``min`` is greater than a corresponding broadcasted element in ``max``, behavior is unspecified and thus implementation-dependent.
-    - If ``x`` has an integral data type and a broadcasted element in ``min`` or ``max`` is outside the bounds of the data type of ``x``, the behavior is unspecified and thus implementation-dependent.
+    - If ``x`` has an integral data type and a broadcasted element in ``min`` or ``max`` is outside the bounds of the data type of ``x``, behavior is unspecified and thus implementation-dependent.
     - If ``x`` and either ``min`` or ``max`` have different data type kinds (e.g., integer versus floating-point), behavior is unspecified and thus implementation-dependent.
 
     .. versionadded:: 2023.12
--- a/src/array_api_stubs/_draft/elementwise_functions.py
+++ b/src/array_api_stubs/_draft/elementwise_functions.py
@@ -806,7 +806,7 @@ def clip(
 
     - If both ``min`` and ``max`` are ``None``, the elements of the returned array must equal the respective elements in ``x``.
     - If a broadcasted element in ``min`` is greater than a corresponding broadcasted element in ``max``, behavior is unspecified and thus implementation-dependent.
-    - If ``x`` has an integral data type and a broadcasted element in ``min`` or ``max`` is outside the bounds of the data type of ``x``, the behavior is unspecified and thus implementation-dependent.
+    - If ``x`` has an integral data type and a broadcasted element in ``min`` or ``max`` is outside the bounds of the data type of ``x``, behavior is unspecified and thus implementation-dependent.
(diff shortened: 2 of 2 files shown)


kgryte:
@asmeurer You opened this PR, but, based on your comments in https://github.com/numpy/numpy/issues/24976#issuecomment-2201078537, it is not clear whether you still support making this change. Are you okay with this PR moving forward, assuming that the current standardized behavior remains the same (i.e., no type promotion)?

asmeurer:
Yes, I think this text is especially important *given* that there is no promotion. If there were promotion this text would serve no purpose, because x would just promote to the common data type and there would be no "out-of-bounds" values.

fancidev:
Separately, there might be a few other issues   with `clip`.

- In the parameter doc of `min` and `max`, `x1` should be `x` if I read correctly.

- Similar in spirit to the second Note, what if x is float32, min and max are float64, min < max, but min and max are so close together that they are between two adjacent float32 numbers? (i.e. there exists no float32 number between min and max)

- The additional Note proposed by this PR addresses the case where x has an integral type. But the param doc of x above says x should have a real-valued data type. Which one should prevail?

These ambiguities seem to root from the fact that what exactly is the “clip” operation? Is it supposed to be equivalent to a shorthand for maximum(minimum(…))? Or is it supposed to be some sort of “saturation” that retains the input type? This seems not clear.

asmeurer:
The more general issue with a min/max float64 is that rounding might work against you when downcasting to float32. This does not require `min` and `max` to be near each other, but rather just near `x`. See https://github.com/numpy/numpy/issues/24976#issuecomment-2195598454. (this is why I personally believe the non-promoting rule in `clip` was a bad idea. This sort of thing is exactly why type promotion behavior exists)

> ”If min and max have the same data type as x and min <= max, then clip(x, min, max) is equivalent to maximum(minimum(x, max), min). Otherwise, the behavior is implementation defined.”
 
Only defining `clip` for the case where `x`, `min`, and `max` have the same data type would be a significant change from the existing text, and probably not what people would want.

fancidev:
The referenced numpy issue is an interesting one. It shows that Array API conformant semantics *cannot* be implemented by promotion followed by downcasting.

A general definition of clip that is consistent with the currently proposed semantics could be the following:

> Let T be the data type of x. If there exists a value in T that is greater than or equal to min and less than or equal to max, then clip(x, min, max) returns (1) the smallest element in T that is greater than or equal to min if x < min, (2) the largest element in T that is less than or equal to max if x > max, or (3) x.

This definition consistently accounts for any mixture of data type. It is certainly implementable, but since some implementations already impose a stricter requirement on the input types, it may not be suitable for a standard. That’s why I proposed two more restrictive versions in the previous comment, one assuming min and max are representable in T, and one assuming min and max have the type T.

src/array_api_stubs/_2023_12/elementwise_functions.py
```diff
@@ -805,6 +805,7 @@ def clip(
 
     - If both ``min`` and ``max`` are ``None``, the elements of the returned array must equal the respective elements in ``x``.
     - If a broadcasted element in ``min`` is greater than a corresponding broadcasted element in ``max``, behavior is unspecified and thus implementation-dependent.
+    - If ``x`` has an integral data type and a broadcasted element in ``min`` or ``max`` is outside the bounds of the data type of ``x``, the behavior is unspecified and thus implementation-dependent.
```
kgryte, line 808:
```suggestion
    - If ``x`` has an integral data type and a broadcasted element in ``min`` or ``max`` is outside the bounds of the data type of ``x``, behavior is unspecified and thus implementation-dependent.
```

src/array_api_stubs/_draft/elementwise_functions.py
```diff
@@ -806,6 +806,7 @@ def clip(
 
     - If both ``min`` and ``max`` are ``None``, the elements of the returned array must equal the respective elements in ``x``.
     - If a broadcasted element in ``min`` is greater than a corresponding broadcasted element in ``max``, behavior is unspecified and thus implementation-dependent.
+    - If ``x`` has an integral data type and a broadcasted element in ``min`` or ``max`` is outside the bounds of the data type of ``x``, the behavior is unspecified and thus implementation-dependent.
```
kgryte, line 809:
```suggestion
    - If ``x`` has an integral data type and a broadcasted element in ``min`` or ``max`` is outside the bounds of the data type of ``x``, behavior is unspecified and thus implementation-dependent.
```

kgryte:
> this is why I personally believe the non-promoting rule in clip was a bad idea. This sort of thing is exactly why type promotion behavior exists

As a general principle, yes, but `clip` doesn't fall neatly into the same category as, e.g., `add`, `mul`, etc, where type promotion makes more intuitive sense.

If we wanted to allow type promotion wiggle room for some libraries while also allowing other libraries to make the reasonable choice of raising an exception, we could limit portable `clip` behavior to only the scenario in which `x`, `min`, and `max` all have the same data type. We'd need to tweak the language of the spec a bit to open the door for implementing libraries to implement non-portable behavior without running afoul of the standard.

Were we to circumscribe portable behavior, this would require downstream libraries building on the Array API to be explicit about what behavior they want (e.g., by wrapping `x` in `astype` or `min` and `max` in `astype`). E.g., if they always want type promotion semantics, then use `result_type` to determine the desired output array dtype. If they want `min` and `max` to always be downcasted, then use `astype` on `x`. Et cetera.

kgryte:
A brief update. I went through SciPy and scikit-learn to see their usage of `xp.clip` (`np.clip`) and overwhelmingly the dominant use case is `clip(x, min_scalar, max_scalar)`. In the few instances where `min` and/or `max` were arrays, there was no expectation/intent to have `x` type promote, the data types all being the same.

Accordingly, I think it is likely fine to make the standard more restricted here in which we expect all three arguments to have the same data types.

kgryte (approved):
LGTM. The discussion regarding type promotion behavior (or the lack thereof) will be addressed in a follow-up PR.

Issue #808, opened 2024-05-30
RFC: add `take_along_axis` to take values along a specified dimension


kgryte:
This RFC proposes the addition of a new API in the array API specification for taking values from an input array by matching one-dimensional index and data slices.

## Overview

Based on array comparison [data](https://github.com/data-apis/array-api-comparison/blob/f61360fccc8b1b5fba7cf6d64ee450c7d0eddd5d/signatures/indexing/take_along_axis.md), the API is available across most major array libraries in the PyData ecosystem.

`take_along_axis` was previously discussed in https://github.com/data-apis/array-api/issues/177 as a potential standardization candidate and has been mentioned in downstream [usage](https://github.com/data-apis/array-api/issues/177#issuecomment-2093630595). As indexing with multidimensional integer arrays (see https://github.com/data-apis/array-api/issues/669) is not yet supported in the specification, the specification lacks a means to concisely select multiple values along multiple one-dimensional slices. This RFC aims to fill this gap.

Additionally, even with advanced indexing, replicating `take_along_axis` is more verbose and trickier to get right. For example, consider

```python
In [1]: import numpy as np

In [2]: a = np.array([[10,30,20], [60,40,50]])

In [3]: a
Out[3]:
array([[10, 30, 20],
       [60, 40, 50]])

In [4]: indices = np.array([[2,0,1],[1,2,0]])

In [5]: indices
Out[5]:
array([[2, 0, 1],
       [1, 2, 0]])

In [6]: np.take_along_axis(a, indices, axis=1)
Out[6]:
array([[20, 10, 30],
       [40, 50, 60]])
```

To replicate with advanced indexing,

```python
In [7]: i0 = np.arange(a.shape[0])[:, np.newaxis]

In [8]: a[i0, indices]
Out[8]:
array([[20, 10, 30],
       [40, 50, 60]])
```

where we need to create an integer index (with expanded dimensions) for the first dimension, which can then be broadcast against the integer index `indices`. Especially for higher order dimensions, replication of `take_along_axis` becomes even more verbose. E.g., for a 3-dimensional array,

```python
a = np.random.rand(3, 4, 5)
indices = np.random.randint(5, size=(3, 4, 2))

i0 = np.arange(a.shape[0])[:, np.newaxis, np.newaxis]
i1 = np.arange(a.shape[1])[np.newaxis, :, np.newaxis]

result = a[i0, i1, indices]
```

[...]

PR #816, opened 2024-06-27
feat: add `take_long_axis` to specifiation

kgryte:
This PR

- resolves https://github.com/data-apis/array-api/issues/808 by adding support for selecting elements at one-dimensional indices along a specified axis.
- makes the `axis` kwarg keyword-only, which differs from the OP (see https://github.com/data-apis/array-api/issues/808). This is for consistency with `axis` kwargs elsewhere in the specification. The proposal seemed to follow PyTorch in allowing both positional and kwarg, which could be considered allowed behavior, but positional usage won't be portable.
- leaves out-of-bounds behavior unspecified.

### DIFF
--- a/spec/draft/API_specification/indexing_functions.rst
+++ b/spec/draft/API_specification/indexing_functions.rst
@@ -21,3 +21,4 @@ Objects in API
    :template: method.rst
 
    take
+   take_along_axis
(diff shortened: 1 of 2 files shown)


kgryte:
As this PR has been discussed during workgroup meetings and has not received any objections, I'll go ahead and merge.

mdhaber:
Thank you! This will be very helpful.

PR #821, opened 2024-07-15
docs: add note that iteration is defined for 1-D arrays

### LINKED ISSUE #818: Iteration on 1-D arrays
Recently in array-api-strict, I [accidentally disabled iteration on 1-D arrays](https://github.com/data-apis/array-api-strict/pull/48). This [broke a lot of code in SciPy](https://github.com/scipy/scipy/pull/21074). I've since reverted the change (array-api-strict disallows iteration on >1-D arrays but allows it for 1-D arrays). 

There have been discussions in the past about now allowing iteration on arrays https://github.com/data-apis/array-api/issues/188. Disallowing it for higher dimensional arrays is probably fine, but it's unclear whether a library like array-api-strict should disallow it for 1-D arrays. The reason is that technically speaking, an array object that implements on the methods defined in the standard would allow iteration on 1-D arrays. This is because by default if `__iter__` is not defined but `__getitem__` is, Python defines iteration as `a[0]`, `a[1]`, etc. 

Given how painful this can be for upstream code, I wonder if we should make it explicit in the standard that iteration is defined for 1-D arrays. 

A possible counterargument is that the new [`unstack`](https://data-apis.org/array-api/latest/API_specification/generated/array_api.unstack.html#unstack) function can be used to iterate on an array of any dimension. `unstack(x)` is the same as `iter(x)` in the NumPy sense of iteration (it iterates the elements if `x` is 1-dimensional and along the first axis if it is n-dimensional).

asmeurer:
Fixes #818

I also added a cross-reference to the indexing page to the `__getitem__` and `__setitem__` docs.

### DIFF
--- a/src/array_api_stubs/_draft/array_object.py
+++ b/src/array_api_stubs/_draft/array_object.py
@@ -631,7 +631,7 @@ class _array:
             an array containing the accessed value(s). The returned array must have the same data type as ``self``.
 
         .. note::
-           When ``__getitem__`` is defined on an object, Python will automatically define iteration (i.e., the behavior from ``iter(x)``) as  ``x[0]``, ``x[1]``, .... This can also be implemented directly by defining ``__iter__``. Therefore, for 1-D arrays, iteration should produce the sequence ``x[0]``, ``x[1]``, ... of 0-D arrays. Iteration behavior for arrays with more than one dimension is unspecified and thus implementation-defined.
+           When ``__getitem__`` is defined on an object, Python will automatically define iteration (i.e., the behavior from ``iter(x)``) as  ``x[0]``, ``x[1]``, and so forth. This can also be implemented directly by defining ``__iter__``. Therefore, for one-dimensional arrays, iteration should produce the sequence ``x[0]``, ``x[1]``, ... of zero-dimensional arrays. Iteration behavior for arrays having more than one dimension is unspecified and thus implementation-defined.
 
         """
 
--- a/src/array_api_stubs/_draft/array_object.py
+++ b/src/array_api_stubs/_draft/array_object.py
@@ -631,7 +631,7 @@ class _array:
             an array containing the accessed value(s). The returned array must have the same data type as ``self``.
 
         .. note::
-           When ``__getitem__`` is defined on an object, Python will automatically define iteration (i.e., the behavior from ``iter(x)``) as  ``x[0]``, ``x[1]``, ..., ``x[N-1]``. This can also be implemented directly by defining ``__iter__``. Therefore, for a one-dimensional array ``x``, iteration should produce a sequence of zero-dimensional arrays ``x[0]``, ``x[1]``, ..., ``x[N-1]``, where ``N`` is the number of elements in the array. Iteration behavior for arrays having more than one dimension is unspecified and thus implementation-defined.
+           When ``__getitem__`` is defined on an object, Python will automatically define iteration (i.e., the behavior from ``iter(x)``) as  ``x[0]``, ``x[1]``, ..., ``x[N-1]``. This can also be implemented directly by defining ``__iter__``. Therefore, for a one-dimensional array ``x``, iteration should produce a sequence of zero-dimensional arrays ``x[0]``, ``x[1]``, ..., ``x[N-1]``, where ``N`` is the number of elements in the array. Iteration behavior for arrays having zero dimensions or more than one dimension is unspecified and thus implementation-defined.
 
         """
 
--- a/src/array_api_stubs/_draft/array_object.py
+++ b/src/array_api_stubs/_draft/array_object.py
@@ -629,6 +629,10 @@ class _array:
         -------
         out: array
             an array containing the accessed value(s). The returned array must have the same data type as ``self``.
+
+        .. note::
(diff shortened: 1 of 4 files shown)


src/array_api_stubs/_draft/array_object.py
```diff
@@ -627,6 +629,10 @@ def __getitem__(
         -------
         out: array
             an array containing the accessed value(s). The returned array must have the same data type as ``self``.
+
+        .. note::
+           When ``__getitem__`` is defined on an object, Python will automatically define iteration (i.e., the behavior from ``iter(x)``) as  ``x[0]``, ``x[1]``, .... This can also be implemented directly by defining ``__iter__``. Therefore, for 1-D arrays, iteration should produce the sequence ``x[0]``, ``x[1]``, ... of 0-D arrays. Iteration behavior for arrays with more than one dimension is unspecified and thus implementation-defined.
```
kgryte, line 634:
```suggestion
           When ``__getitem__`` is defined on an object, Python will automatically define iteration (i.e., the behavior from ``iter(x)``) as  ``x[0]``, ``x[1]``, et cetera. This can also be implemented directly by defining ``__iter__``. Therefore, for one-dimensional arrays, iteration should produce the sequence ``x[0]``, ``x[1]``, ... of zero-dimensional arrays. Iteration behavior for arrays having more than one dimension is unspecified and thus implementation-defined.
```

src/array_api_stubs/_draft/array_object.py
```diff
@@ -627,6 +629,10 @@ def __getitem__(
         -------
         out: array
             an array containing the accessed value(s). The returned array must have the same data type as ``self``.
+
+        .. note::
+           When ``__getitem__`` is defined on an object, Python will automatically define iteration (i.e., the behavior from ``iter(x)``) as  ``x[0]``, ``x[1]``, and so forth. This can also be implemented directly by defining ``__iter__``. Therefore, for one-dimensional arrays, iteration should produce the sequence ``x[0]``, ``x[1]``, ... of zero-dimensional arrays. Iteration behavior for arrays having more than one dimension is unspecified and thus implementation-defined.
```
kgryte, line 634:
```suggestion
           When ``__getitem__`` is defined on an object, Python will automatically define iteration (i.e., the behavior from ``iter(x)``) as  ``x[0]``, ``x[1]``, and so forth. This can also be implemented directly by defining ``__iter__``. Therefore, for one-dimensional arrays, iteration should produce a sequence of zero-dimensional arrays ``x[0]``, ``x[1]``, and so forth. Iteration behavior for arrays having more than one dimension is unspecified and thus implementation-defined.
```

src/array_api_stubs/_draft/array_object.py
```diff
@@ -627,6 +629,10 @@ def __getitem__(
         -------
         out: array
             an array containing the accessed value(s). The returned array must have the same data type as ``self``.
+
+        .. note::
+           When ``__getitem__`` is defined on an object, Python will automatically define iteration (i.e., the behavior from ``iter(x)``) as  ``x[0]``, ``x[1]``, and so forth. This can also be implemented directly by defining ``__iter__``. Therefore, for one-dimensional arrays, iteration should produce a sequence of zero-dimensional arrays ``x[0]``, ``x[1]``, and so forth. Iteration behavior for arrays having more than one dimension is unspecified and thus implementation-defined.
```
kgryte, line 634:
```suggestion
           When ``__getitem__`` is defined on an object, Python will automatically define iteration (i.e., the behavior from ``iter(x)``) as  ``x[0]``, ``x[1]``, ..., ``x[N-1]``. This can also be implemented directly by defining ``__iter__``. Therefore, for a one-dimensional array ``x``, iteration should produce a sequence of zero-dimensional arrays ``x[0]``, ``x[1]``, ..., ``x[N-1]``, where ``N`` is the number of elements in the array. Iteration behavior for arrays having more than one dimension is unspecified and thus implementation-defined.
```

src/array_api_stubs/_draft/array_object.py
```diff
@@ -627,6 +629,10 @@ def __getitem__(
         -------
         out: array
             an array containing the accessed value(s). The returned array must have the same data type as ``self``.
+
+        .. note::
+           When ``__getitem__`` is defined on an object, Python will automatically define iteration (i.e., the behavior from ``iter(x)``) as  ``x[0]``, ``x[1]``, ..., ``x[N-1]``. This can also be implemented directly by defining ``__iter__``. Therefore, for a one-dimensional array ``x``, iteration should produce a sequence of zero-dimensional arrays ``x[0]``, ``x[1]``, ..., ``x[N-1]``, where ``N`` is the number of elements in the array. Iteration behavior for arrays having more than one dimension is unspecified and thus implementation-defined.
```
kgryte, line 634: @asmeurer What should happen for zero-dimensional arrays?

asmeurer: On a 0-D array `x[i]` is an IndexError, so this doesn't really apply (the Python auto-`__iter__` would never work). Ideally calling `list()` on a 0-D should always be an error, but we can also just change this to say "arrays not having one dimension".

kgryte, line 634:
```suggestion
           When ``__getitem__`` is defined on an object, Python will automatically define iteration (i.e., the behavior from ``iter(x)``) as  ``x[0]``, ``x[1]``, ..., ``x[N-1]``. This can also be implemented directly by defining ``__iter__``. Therefore, for a one-dimensional array ``x``, iteration should produce a sequence of zero-dimensional arrays ``x[0]``, ``x[1]``, ..., ``x[N-1]``, where ``N`` is the number of elements in the array. Iteration behavior for arrays having zero dimensions or more than one dimension is unspecified and thus implementation-defined.
```

kgryte (approved):
LGTM. Thanks, @asmeurer!

PR #822, opened 2024-07-15
docs: add note that cross-kind comparisons are undefined

### LINKED ISSUE #819: Comparing arrays of different dtype kinds
In array-api-strict, I've had this behavior

```py
>>> import array_api_strict as xp
>>> xp.asarray(0, dtype=xp.int64) < xp.asarray(1, dtype=xp.uint64)
TypeError: array_api_strict.int64 and array_api_strict.uint64 cannot be type promoted together
```

However, this is very annoying, as it makes it basically impossible to compare uint64 and int64. However, I noticed that the standard doesn't actually say anywhere that the inputs to `greater`/`__gt__` need to be type-promotable, only that they should be real numeric. So from my reading of the standard, even comparing integers and floats should be allowed (but not complex numbers or booleans). 

That's all good and fine, and I'm going to update array-api-strict. The only question there is whether the standard should be more explicit about this, or if the current wording is OK. It might also be a good idea to update the wording to explicitly state that the comparison should be based on the full values of the operands. There was an issue in NumPy pre-2.0 where `<` sometimes gave wrong answers because it would downcast float64 to float32, leading to a wrong answer if the rounding went wrong. In other words, the types should promote (or virtually promote) internally before comparing. I actually don't know if all libraries handle the uint64/int64 case correctly since it isn't tested in the test suite. I know NumPy has historically promoted those types to float64, which is dumb since float64 can't even represent the maximum int64. I suppose one could also consider cases of Python int scalars that are very large which should probably compare without doing an overflowing cast, but this could also be treated as unspecified. 
 
The main question I had is with `equal`/`==` and `not_equal`/`!=`. For those functions, the standard only says the inputs can be "any data type" https://data-apis.org/array-api/latest/API_specification/generated/array_api.equal.html#equal. This implies, by the above reasoning, that comparing any two arrays of any two data types should be allowed. But this implies things like comparing complex arrays to real arrays, and comparing booleans to integers. It might be a good idea to clarify either whether this should not be allowed (or implementation defined), or if it is allowed, how these different types compare to one another. We can probably reuse some other text, for instance, the text at https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__bool__.html#bool for comparing with a bool array and the text at https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__complex__.html#complex for comparing with a complex array.

asmeurer:
Does the language here look OK? Should we add anything along the lines of "comparisons with promotable types should promote first" (so that the result is numerically correct)? 

Fixes #819.

### DIFF
--- a/src/array_api_stubs/_draft/array_object.py
+++ b/src/array_api_stubs/_draft/array_object.py
@@ -513,6 +513,9 @@ class _array:
 
         .. note::
            Element-wise results, including special cases, must equal the results returned by the equivalent element-wise function :func:`~array_api.equal`.
+
+        .. note::
+           Comparisons of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
         """
 
     def __float__(self: array, /) -> float:
@@ -599,6 +602,9 @@ class _array:
 
         .. note::
            Element-wise results must equal the results returned by the equivalent element-wise function :func:`~array_api.greater_equal`.
+
+        .. note::
+           Comparisons of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
         """
 
     def __getitem__(
@@ -651,6 +657,9 @@ class _array:
 
         .. note::
            Element-wise results must equal the results returned by the equivalent element-wise function :func:`~array_api.greater`.
+
+        .. note::
+           Comparisons of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
         """
 
     def __index__(self: array, /) -> int:
@@ -748,6 +757,9 @@ class _array:
 
         .. note::
            Element-wise results must equal the results returned by the equivalent element-wise function :func:`~array_api.bitwise_invert`.
+
+        .. note::
+           Comparisons of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
         """
 
     def __le__(self: array, other: Union[int, float, array], /) -> array:
@@ -772,6 +784,9 @@ class _array:
 
         .. note::
            Element-wise results must equal the results returned by the equivalent element-wise function :func:`~array_api.less_equal`.
+
+        .. note::
+           Comparisons of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
         """
 
     def __lshift__(self: array, other: Union[int, array], /) -> array:
@@ -942,6 +957,9 @@ class _array:
 
         .. note::
            Element-wise results, including special cases, must equal the results returned by the equivalent element-wise function :func:`~array_api.not_equal`.
+
+        .. note::
+           Comparisons of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
 
         .. versionchanged:: 2022.12
             Added complex data type support.
--- a/src/array_api_stubs/_draft/array_object.py
+++ b/src/array_api_stubs/_draft/array_object.py
@@ -829,6 +829,9 @@ class _array:
 
         .. note::
            Element-wise results must equal the results returned by the equivalent element-wise function :func:`~array_api.less`.
+
+        .. note::
+           Comparison of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
         """
 
     def __matmul__(self: array, other: array, /) -> array:
--- a/src/array_api_stubs/_draft/array_object.py
+++ b/src/array_api_stubs/_draft/array_object.py
@@ -515,7 +515,7 @@ class _array:
            Element-wise results, including special cases, must equal the results returned by the equivalent element-wise function :func:`~array_api.equal`.
 
         .. note::
-           Comparisons of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
+           Comparison of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
         """
 
     def __float__(self: array, /) -> float:
@@ -604,7 +604,7 @@ class _array:
            Element-wise results must equal the results returned by the equivalent element-wise function :func:`~array_api.greater_equal`.
 
         .. note::
-           Comparisons of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
+           Comparison of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
         """
 
     def __getitem__(
@@ -659,7 +659,7 @@ class _array:
            Element-wise results must equal the results returned by the equivalent element-wise function :func:`~array_api.greater`.
 
         .. note::
-           Comparisons of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
+           Comparison of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
         """
 
     def __index__(self: array, /) -> int:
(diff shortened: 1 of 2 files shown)


src/array_api_stubs/_draft/array_object.py
```diff
@@ -513,6 +513,9 @@ def __eq__(self: array, other: Union[int, float, bool, array], /) -> array:
 
         .. note::
            Element-wise results, including special cases, must equal the results returned by the equivalent element-wise function :func:`~array_api.equal`.
+
+        .. note::
+           Comparisons of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
```
kgryte, line 518:
```suggestion
           Comparisons of arrays without a corresponding promotable data type (see :ref:`type-promotion`) are undefined and thus implementation-dependent.
```

src/array_api_stubs/_draft/array_object.py
```diff
@@ -599,6 +602,9 @@ def __ge__(self: array, other: Union[int, float, array], /) -> array:
 
         .. note::
            Element-wise results must equal the results returned by the equivalent element-wise function :func:`~array_api.greater_equal`.
+
+        .. note::
+           Comparisons of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
```
kgryte, line 607:
```suggestion
           Comparisons of arrays without a corresponding promotable data type (see :ref:`type-promotion`) are undefined and thus implementation-dependent.
```

src/array_api_stubs/_draft/array_object.py
```diff
@@ -651,6 +657,9 @@ def __gt__(self: array, other: Union[int, float, array], /) -> array:
 
         .. note::
            Element-wise results must equal the results returned by the equivalent element-wise function :func:`~array_api.greater`.
+
+        .. note::
+           Comparisons of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
```
kgryte, line 662:
```suggestion
           Comparisons of arrays without a corresponding promotable data type (see :ref:`type-promotion`) are undefined and thus implementation-dependent.
```

src/array_api_stubs/_draft/array_object.py
```diff
@@ -943,6 +958,9 @@ def __ne__(self: array, other: Union[int, float, bool, array], /) -> array:
         .. note::
            Element-wise results, including special cases, must equal the results returned by the equivalent element-wise function :func:`~array_api.not_equal`.
 
+        .. note::
+           Comparisons of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
```
kgryte, line 962:
```suggestion
           Comparison of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
```

src/array_api_stubs/_draft/array_object.py
```diff
@@ -748,6 +757,9 @@ def __invert__(self: array, /) -> array:
 
         .. note::
            Element-wise results must equal the results returned by the equivalent element-wise function :func:`~array_api.bitwise_invert`.
+
+        .. note::
+           Comparisons of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
```
kgryte, line 762:
```suggestion
```

I don't think we need this note for `invert`, as it is not a comparison operation.

src/array_api_stubs/_draft/array_object.py
```diff
@@ -772,6 +784,9 @@ def __le__(self: array, other: Union[int, float, array], /) -> array:
 
         .. note::
            Element-wise results must equal the results returned by the equivalent element-wise function :func:`~array_api.less_equal`.
+
+        .. note::
+           Comparisons of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
```
kgryte, line 789:
```suggestion
           Comparison of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
```

src/array_api_stubs/_draft/array_object.py
```diff
@@ -772,6 +784,9 @@ def __le__(self: array, other: Union[int, float, array], /) -> array:
 
         .. note::
            Element-wise results must equal the results returned by the equivalent element-wise function :func:`~array_api.less_equal`.
+
+        .. note::
```
kgryte, line 788: This comment should be added to `__lt__` below.

src/array_api_stubs/_draft/elementwise_functions.py
```diff
@@ -1118,6 +1118,9 @@ def equal(x1: array, x2: array, /) -> array:
     .. note::
        For discussion of complex number equality, see :ref:`complex-numbers`.
 
+    .. note::
+       Comparisons of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
```
kgryte, line 1122:
```suggestion
       Comparison of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
```

src/array_api_stubs/_draft/elementwise_functions.py
```diff
@@ -1347,6 +1350,10 @@ def greater(x1: array, x2: array, /) -> array:
     -------
     out: array
         an array containing the element-wise results. The returned array must have a data type of ``bool``.
+
+    .. note::
+       Comparisons of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
```
kgryte, line 1355:
```suggestion
       Comparison of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
```

src/array_api_stubs/_draft/elementwise_functions.py
```diff
@@ -1368,6 +1375,9 @@ def greater_equal(x1: array, x2: array, /) -> array:
...
```
kgryte, line 1380:
```suggestion
       Comparison of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
```

src/array_api_stubs/_draft/elementwise_functions.py
```diff
@@ -1563,6 +1573,9 @@ def less(x1: array, x2: array, /) -> array:
...
```
kgryte, line 1578:
```suggestion
       Comparison of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
```

src/array_api_stubs/_draft/elementwise_functions.py
```diff
@@ -1584,6 +1597,9 @@ def less_equal(x1: array, x2: array, /) -> array:
...
```
kgryte, line 1602:
```suggestion
       Comparison of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
```

src/array_api_stubs/_draft/elementwise_functions.py
```diff
@@ -2134,6 +2150,9 @@ def not_equal(x1: array, x2: array, /) -> array:
     .. note::
        For discussion of complex number equality, see :ref:`complex-numbers`.
 
+    .. note::
+       Comparisons of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
```
kgryte, line 2154:
```suggestion
       Comparison of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
```

kgryte (changes requested):
The language looks okay to me. Left comments for a small grammar correction and the need to add a missing note. Once addressed, LGTM.

asmeurer:
Done

fancidev:
For my understanding, by “comparison of arrays without a corresponding promotable data type”, do you mean that there does not exist a type where both array are promotable to, or do you mean the array has a type that can no longer be promoted “further”? I’m asking because the reference issue involves an example of int64 vs uint64, for which both interpretations apply, so it doesn’t seem entirely clear to me.

kgryte (approved):
LGTM. Thanks, @asmeurer!

kgryte:
> Should we add anything along the lines of "comparisons with promotable types should promote first" (so that the result is numerically correct)?

I believe this is already implied, as, practically speaking, in order to perform comparison, promotion is necessary (e.g., as automatically happens when comparing a float and a double in C). If we later find that we need to make this more explicit, we can address in a follow-up PR.

PR #838, opened 2024-08-26
docs: fix reference to eigh/eigvalsh

mdhaber:
The linear algebra extension specification refers to `eig`/`eigvals` as an example of dedicated interfaces rather than a single interface with polymorphic return type. 

`eig`/`eigvals` are not yet in the standard, so this PR replaces these references with `eigh`/`eigvalsh`.

(I went ahead and changed the `202x.12` versions in addition to `draft`; if that's not right, it should be easy to revert them.)


kgryte (approved):
LGTM. Thanks, @mdhaber!

PR #844, opened 2024-09-21
docs: add missing subscripts in `sign` definition

mdhaber:
Looked like these subscripts were omitted.

LMK if this should be adjusted in 2012/2013 standards, but I think the intent is obvious enough that it doesn't need to be fixed in old versions.


kgryte:
Thanks for flagging this. Yes, if you wouldn't mind backporting to previous spec versions, that would be helpful.

oleksandr-pavlyk (approved):

mdhaber:
Thanks @kgryte. I've added the subscripts to past versions, too.

kgryte (approved):
LGTM. Thanks, @mdhaber!

Issue #841, opened 2024-09-11
Arithmetic between real arrays and Python complex scalars (revisited)


mdhaber:
The array API standard is now clear (in [Type Promotion Rules](https://data-apis.org/array-api/latest/API_specification/type_promotion.html)):

![image](https://github.com/user-attachments/assets/815ab116-9a68-4f60-ad01-5be101f885e3)

This means that following does not have defined behavior:
```python3
x = xp.asarray([1., 2., 3.])
x + 1j
```

However, the behavior is defined consistently in NumPy, CuPy, PyTorch, jax.numpy[^1], Dask array, and Tensorflow[^2] in the intuitive way: the real and imaginary component dtypes of the complex output array match the dtype of the input array.

This issue was discussed explicitly in gh-478 beginning with https://github.com/data-apis/array-api/issues/478#issuecomment-1271205727. I see several comments that seem supportive of allowing this operation, e.g.
> I would hope it at least works for Python scalars. `x + 0j` seems like the simplest and most obvious way to convert a real array to a complex one.

The comment I see against it (https://github.com/data-apis/array-api/issues/478#issuecomment-1271226040) is 
> I know it's very common to do this (also * `1.0` to convert integer to real). However, it doesn't seem like great practice irrespective of the decision here - there's an explicit `astype` function for converting to other dtypes.

It looked like the summary comments suggested that this would be allowed, but it was not specifically addressed in the PR that closed the issue. (See postscript for more information.)

I thought it might help to add some perspective on how this impacts a developer translating code to the standard: when tests including an operation like
```
x + 1j
```
begin to fail with `array_api_strict` only, the developer is faced with the choice of skipping the `array_api_strict` tests or figuring out what is wrong and changing the code to something like:
```python3
xp.astype(x, xp.result_type(x.dtype, xp.complex64)) + 1j
```
If they also want it to be able to preserve the lower precision types supported by some libraries (e.g. for NumPy), there would be additional hurdles. I tested for a few libraries, and array API standard compatible version of the code also increases the execution time notably for small arrays. Individually, the inconvenience, complexity, and overhead are small, but they add up. I am excited about adding array API support to SciPy, but not everyone is supportive, and performance regressions and complicated-looking diffs can make it difficult to garner support.

I would suggest that the operation be defined, even if it means adding an exception to the simply stated rules for array/Python arithmetic.

---

The more specific language about this not being defined was added in gh-513, which provided the justification:

[...]

PR #848, opened 2024-10-24
feat: add support for specifying a data type "kind" in `astype`

lucascolley:
This PR:

- Addresses a comment made in https://github.com/data-apis/array-api/issues/841#issuecomment-2392032433 and should close https://github.com/data-apis/array-api/issues/859 by adding data type "kind" support in `astype`.


src/array_api_stubs/_draft/data_type_functions.py
```diff
@@ -40,8 +40,14 @@ def astype(
...
-    dtype: dtype
-        desired data type.
+    dtype_or_kind: Union[dtype, str]
+        desired data type or kind of data type. Supported kinds are:
+        -   ``'bool'``: boolean data types (e.g., ``bool``).
+        -   ``'signed integer'``: signed integer data types (e.g., ``int8``, ``int16``, ``int32``, ``int64``).
+        -   ``'unsigned integer'``: unsigned integer data types (e.g., ``uint8``, ``uint16``, ``uint32``, ``uint64``).
+        -   ``'integral'``: integer data types. Shorthand for ``('signed integer', 'unsigned integer')``.
+        -   ``'real floating'``: real-valued floating-point data types (e.g., ``float32``, ``float64``).
+        -   ``'complex floating'``: complex floating-point data types (e.g., ``complex64``, ``complex128``).
```
lucascolley, line 50: I included `integral` here, but left out `numeric`. Would be happy to have both or neither if that is preferred.

src/array_api_stubs/_draft/data_type_functions.py
```diff
@@ -13,10 +13,15 @@
 
 
 def astype(
-    x: array, dtype: dtype, /, *, copy: bool = True, device: Optional[device] = None
+    x: array,
+    dtype_or_kind: Union[dtype, str],
```
rgommers, line 17: I'd prefer to keep the name as `dtype`, the longer one is quite unwieldy. Also if anyone implemented this without using positional-only, then they may not be able to change the name.

src/array_api_stubs/_draft/data_type_functions.py
rgommers: I don't think `integral` can be included, since it's unclear whether to select signed or unsigned.

src/array_api_stubs/_draft/data_type_functions.py
```diff
@@ -50,7 +61,15 @@ def astype(
     Returns
     -------
     out: array
-        an array having the specified data type. The returned array must have the same shape as ``x``.
+        For ``dtype_or_kind`` a data type, an array having the specified data type.
+        For ``dtype_or_kind`` a kind of data type:
+        -   If ``x.dtype`` is already of that kind, the data type is maintained.
+        -   Otherwise, an attempt is made to convert to the specified kind, according to the type promotion rules (see :ref:`type-promotion`).
```
rgommers, line 67: Why "an attempt"? That seems ambiguous. We have to be clear about what must work. Which I think is:
- float to complex 
- unsigned to signed integer 

Anything else doesn't I think? There's no point allowing `'bool'` I think, since there is only one boolean dtype so `dtype=xp.bool` will be cleaner. 

For `'signed integer' and `'real floating-point'` there are also no promotion rules to follow, so they can be left out - or do you see a use case?

rgommers (commented):
Thanks for getting the ball rolling on this @lucascolley! Some comments inline.

src/array_api_stubs/_draft/data_type_functions.py
lucascolley: I've reduced this down to just `'complex floating'` (use-case: mixed float/complex to complex) and `'signed integer'` (use-case: mixed signed/unsigned to signed).

I think "an attempt" would still be accurate for an implementation of this? `xp.astype(some_int8_array, 'complex floating')` would attempt a conversion, whose success will depend on the implementation-specific type promotion rules, right?

Unless you think that this function should always error when the type promotion is not defined by the standard?

rgommers: > I think "an attempt" would still be accurate for an implementation of this?

I think you have the right idea in mind here, it's just a "language we use to specify things" thing. We specify which behavior has to be supported - `'complex floating'` has type promotion rules defined in the standard, so it's expected to always work for a compliant implementation. Then, if we expect other input types to raise, then we specify that by "must raise ..." or "input type must be ...". In this case there's no reason to do that (implementors are free to suppport more types, it's just not standardized), so we then say "input type should be ...". 

Your "attempt to ..." seems to be the same as "should be ...", it's just language we want to write in a uniform way.

src/array_api_stubs/_draft/data_type_functions.py
```diff
@@ -50,7 +59,15 @@ def astype(
...
     -------
     out: array
-        an array having the specified data type. The returned array must have the same shape as ``x``.
+        For ``dtype`` a data type, an array having the specified data type.
+        For ``dtype`` a kind of data type:
+
+        -   If ``x.dtype`` is already of that kind, the data type must be maintained.
+        -   Otherwise, ``x`` should be cast to a data type of that kind, according to the type promotion rules (see :ref:`type-promotion`) and the above notes.
+        -   Kinds must be interpreted as the lowest-precision standard data type of that kind for the purposes of type promotion. For example, ``astype(x, 'complex floating')`` will return an array with the data type ``complex64`` when ``x.dtype`` is ``float32``, since ``complex64`` is the result of promoting ``float32`` with the lowest-precision standard complex data type, ``complex64``.
+        -   Where type promotion is unspecified and thus implementation-specific, the result is also unspecified. For example, ``astype(x, 'complex floating')``, where ``x`` has data type ``int32``.
```
asmeurer, line 68: This is overly restrictive, and conflicts with the behavior specified above. The whole point of astype is to be able to do casts that aren't in the promotion table, like int -> float casts. In fact, I would say this whole line should be deleted as the above note already clearly talks about what is and isn't defined.

src/array_api_stubs/_draft/data_type_functions.py
```diff
@@ -50,7 +59,15 @@ def astype(
     Returns
     -------
     out: array
-        an array having the specified data type. The returned array must have the same shape as ``x``.
+        For ``dtype`` a data type, an array having the specified data type.
+        For ``dtype`` a kind of data type:
+
+        -   If ``x.dtype`` is already of that kind, the data type must be maintained.
+        -   Otherwise, ``x`` should be cast to a data type of that kind, according to the type promotion rules (see :ref:`type-promotion`) and the above notes.
+        -   Kinds must be interpreted as the lowest-precision standard data type of that kind for the purposes of type promotion. For example, ``astype(x, 'complex floating')`` will return an array with the data type ``complex64`` when ``x.dtype`` is ``float32``, since ``complex64`` is the result of promoting ``float32`` with the lowest-precision standard complex data type, ``complex64``.
```
asmeurer, line 67: The text you've added here seems to only be thinking about the case where the `dtype` argument is a string, but the dtype being an actual dtype object is also still supported.

lucascolley: Line 62 covers the dtype case. Do you have a suggestion to make it more clear?

src/array_api_stubs/_draft/data_type_functions.py
lucascolley: > This is overly restrictive, and conflicts with the behavior specified above. The whole point of astype is to be able to do casts that aren't in the promotion table, like int -> float casts

I think there is a misunderstanding - this point applies only in the case that a kind of data type is provided, as this is under the bullet list started on line 63.

Ralf suggested in https://github.com/data-apis/array-api/pull/848#discussion_r1817689517 that it doesn't make sense to support kinds for which the resulting dtype would always be undefined, given that there are no type promotion rules to follow. Do you disagree?

lucascolley:
Gentle reminder of the 2024 milestone here!

src/array_api_stubs/_draft/data_type_functions.py
kgryte: Quick update: I took the liberty to update the wording. I also made the call to broaden the list of data type kinds. I think there are reasonable arguments for providing any one of the numeric data type kinds (e.g., `int32` and `real floating` => `float64`, etc), and it is possible to delineate a set of clearly defined rules in terms of which data type should be returned. Leaving `bool` out seems somewhat arbitrary, especially when the semantics are clearly specified and all other kinds can be, IMO, reasonably provided.

src/array_api_stubs/_draft/data_type_functions.py
kgryte: I agree with Aaron here. I've updated the wording accordingly. It should be possible, e.g., to be able to do `astype(x, dtype="real floating")` where `x` is `int32` in order to return a `float64` array. This conversion between dtype kinds is not part of type promotion rules, but has well-defined behavior, as `int32` can be represented exactly in `float64`.

src/array_api_stubs/_draft/data_type_functions.py
kgryte: @asmeurer Would you be willing to take a look at the latest text to determine whether this satisfies your concerns?

src/array_api_stubs/_draft/data_type_functions.py
lucascolley: > This conversion between dtype kinds is not part of type promotion rules, but has well-defined behavior, as int32 can be represented exactly in float64.

Okay, yes, that sounds right to me now.

src/array_api_stubs/_draft/data_type_functions.py
```diff
@@ -13,35 +13,43 @@
...
+                -   When type promotion rules are not specified between the data type of ``x`` and the specified data type kind (e.g., ``int16`` and ``'real floating'``) and there exists one or more data types belonging to the specified data kind in which the elements in ``x`` can be represented exactly (e.g., ``int32`` and ``float64``), the function must return an array having the smallest data type (in terms of range of values) capable of precisely representing the elements of ``x`` (e.g., if ``x`` is ``int16`` and the data type kind is ``'complex floating'``, then the returned array must have the data type ``complex64``; if ``x`` is `bool`` and the data type kind is ``integral``, then the returned array must have the data type ``int8``).
+                -   When type promotion rules are not specified between the data type of ``x`` and the specified data type kind and there neither exists a data type belonging to the specified data type in which the elements of ``x`` can be represented exactly (e.g., ``uint64`` and ``'real floating'``) nor are there applicable casting rules documented below, behavior is unspecified and thus implementation-defined.
```
lucascolley, line 41:
```suggestion
                -   When type promotion rules are not specified between the data type of ``x`` and there neither exists a data type belonging to the specified data type in which the elements of ``x`` can be represented exactly (e.g., ``uint64`` and ``'real floating'``) nor are there applicable casting rules documented below, behavior is unspecified and thus implementation-defined.
```

lucascolley (commented):
just one nit, otherwise this new wording LGTM, thanks @kgryte !

src/array_api_stubs/_draft/data_type_functions.py
kgryte: I don't think this suggestion makes sense atm, as the removed phrase is part of

"between the data type of ``x`` and the specified data type kind"

Removing "the specified data type kind" removes the `y` in `between x and y`.

kgryte:
As discussed during the workgroup meeting on January 23, 2025, given that this is new behavior without any downstream implementations, we decided to do two things:

1. Punt to v2025 in order to avoid rushing this into the spec.
2. Incubate the API in `array-api-compat` in order to work out any design issues after having real-world implementations.

@ev-br Would you mind adding to `array-api-compat` and circling back here?

src/array_api_stubs/_draft/data_type_functions.py
lucascolley: Looks like this suggestion was erroneous indeed

Issue #797, opened 2024-04-22
RFC: disallow 0-D inputs in `cumulative_sum`


asmeurer:
The standard is not clear what should happen in `cumulative_sum` for 0-D inputs https://data-apis.org/array-api/latest/API_specification/generated/array_api.cumulative_sum.html#cumulative-sum

Note that NumPy and PyTorch have different conventions here:

```py
>>> import numpy as np
>>> np.cumsum(np.asarray(0))
array([0])
>>> import torch
>>> torch.cumsum(torch.asarray(0), dim=0)
tensor(0)
```

`torch.cumsum` unconditionally requires the `dim` argument, whereas `np.cumsum` defaults to computing over a flattened array if `axis=None`. The standard requires `axis` if the dimensionality is greater than 1. However, `axis=0` doesn't really make sense for a 0-D array. NumPy also allows specifying `axis=0` and gives the same result:

```py
>>> np.cumsum(np.asarray(0), axis=0)
array([0])
```

Furthermore, there is ambiguity here on what should happen for a 0-D input when `include_initial=True`. The standard says:

> if include_initial is True, the returned array must have the same shape as x, except the size of the axis along which to compute the cumulative sum must be N+1.

If the result should be 0-D, then clearly `include_initial` must do nothing, since there is no way to increase the number of elements in the result. 

This doesn't seem to have been discussed in the original pull request https://github.com/data-apis/array-api/pull/653 or issue https://github.com/data-apis/array-api/issues/597, and I don't recall it being brought up at the consortium meetings. 

My suggested behavior would be

- The result of `cumulative_sum` on a 0-D input `x` should be a 0-D output which is the same as `x` (i.e., it would work just like `sum(x)`). This matches the behavior that `cumulative_sum` always returns an output with the same dimensionality as the input. 
- The `include_initial` flag would do nothing when the input is 0-D. One can read the existing text as already supporting this behavior, since "the axis along which to compute the cumulative sum" is vacuous. 
- The `axis` argument must be `None` when the input is 0-D or else the result is an error. This matches the usual "axis must be in the range [-ndim, ndim)" condition, which is not currently spelled out this way for `cumulative_sum` but is for other functions in the standard. 

[...]

PR #851, opened 2024-10-31
docs: add note regarding unspecified behavior for 0d arrays in `cumulative_sum`

kgryte:
This PR

- Closes https://github.com/data-apis/array-api/issues/797.
- updates specification language to limit input arrays to those having one or more dimensions. Conforming implementations are permitted to support 0d arrays for backward-compatibility, but this should not be considered portable behavior due to ambiguities arising from expected output for 0d input (e.g., when providing `include_initial`, what should be the dimensionality of the output? what should happen with the `axis` kwarg? etc).
- aligns the specification based on discussions in #797 and the Consortium workgroup meeting on May 2, 2024.

### DIFF
--- a/src/array_api_stubs/_draft/statistical_functions.py
+++ b/src/array_api_stubs/_draft/statistical_functions.py
@@ -18,7 +18,7 @@ def cumulative_sum(
     Parameters
     ----------
     x: array
-        input array. Should have a numeric data type.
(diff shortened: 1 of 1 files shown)


kgryte:
As no objections have been raised for this PR and the proposed change already discussed during workgroup meetings, I'll go ahead and merge. Any further changes can be addressed in follow-up PRs...

Issue #267, opened 2021-09-19
__array_namespace__ type Hint


gilfree:
I hope this falls under the scope of this repo - 

Is there any a way type-hint `__array_namespace__`? Is it even possible to do so, and it needs to be standardized, or maybe it's not possible with current python type hinting system?

Motivation: 

Suppose I want to write a function that will work in multiple conforming libraries, it will probably start with:

```python
xp = x.__array_namespace__()
```

or, like in NEP-47:
```python
def foo(x,y):
   xp = get_namespace(x,y)
```

In both cases, I want to be able to type check `xp`, and have auto-complete on `xp` in the various IDEs. (At least some auto-complete engines relay on static typing e.g https://github.com/microsoft/pyright)

Is this possible?

BvB93:
Unfortunately you can't return `Literal` modules (which would make this _a lot_ easier), but when implementing the API it would be possible to define and return an auxiliary class that mirrors the content of the namespace.

``` python
from __future__ import annotations

import types
import numpy.array_api
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    # Optionally subclass `types.ModuleType` to closer match runtime behavior
    class _ArrayAPINameSpace(types.ModuleType):
        asarray = numpy.array_api.asarray
        arange = numpy.array_api.arange
        empty = numpy.array_api.empty
        # etc
else:
    import numpy.array_api as _ArrayAPINameSpace


class Foo:
    def __array_namespace__(self, /, *, api_version: None | str = None) -> _ArrayAPINameSpace:
        return numpy.array_api

```

BvB93:
Small update: unfortunately https://github.com/python/mypy/issues/708 was never completelly fixed, so you'd have to return `type[_ArrayAPINameSpace]` in the case of mypy, otherwise it will interpret the namespace functions as normal methods and strip their first argument.
``` python
class Foo:
    def __array_namespace__(self, /, *, api_version: None | str = None) -> type[_ArrayAPINameSpace]:
        return numpy.array_api
```

asmeurer:
This issue is also somewhat related https://github.com/data-apis/array-api/issues/229

gilfree:
Hi @BvB93, 
Thanks for the answer. 

Are there any thoughts about making such a type/class part of the standard (It may be helpful for statically checking conformance)?

BvB93:
[...]

PR #858, opened 2024-11-24
feat: automatically generate Protocol for array-api-namespace

34j:
Inherits #685
Partially fixes. #267
Depends #857

Automatically generated from .pre-commit/update_array_namespace.py using pre-commit
Uses Protocols with TypeVars, therefore finally the correct usage of them can be achieved


kgryte:
@34j Given your recent work and the ongoing work in [`array-api-typing`](https://github.com/data-apis/array-api-typing), are you okay closing out this PR?

PR #862, opened 2024-11-25
feat: add scalar support to element-wise functions

kgryte:
This PR

-   progresses https://github.com/data-apis/array-api/issues/807 in adding scalar support to element-wise functions. The changes are restricted to element-wise functions in order to simplify review.
-   updates the signatures for various dunder methods on the array object which did not indicate `complex` scalar support.
-   reformats and consolidates some of the notes included in affected function signatures to a dedicated "Notes" section, as done for special cases. This moves ancillary specification guidance to "below the fold". A subsequent PR will perform a similar clean-up for APIs outside of those affected by adding scalar support.

### DIFF
--- a/src/array_api_stubs/_draft/array_object.py
+++ b/src/array_api_stubs/_draft/array_object.py
@@ -494,7 +494,7 @@ class _array:
               ONE_API = 14
         """
 
-    def __eq__(self: array, other: Union[int, float, bool, array], /) -> array:
+    def __eq__(self: array, other: Union[int, float, complex, bool, array], /) -> array:
         r"""
         Computes the truth value of ``self_i == other_i`` for each element of an array instance with the respective element of the array ``other``.
 
@@ -502,7 +502,7 @@ class _array:
         ----------
         self: array
             array instance. May have any data type.
-        other: Union[int, float, bool, array]
+        other: Union[int, float, complex, bool, array]
             other array. Must be compatible with ``self`` (see :ref:`broadcasting`). May have any data type.
 
         Returns
@@ -510,12 +510,11 @@ class _array:
         out: array
             an array containing the element-wise results. The returned array must have a data type of ``bool``.
 
+        Notes
+        -----
 
-        .. note::
-           Element-wise results, including special cases, must equal the results returned by the equivalent element-wise function :func:`~array_api.equal`.
-
-        .. note::
-           Comparison of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
+        -   Element-wise results, including special cases, must equal the results returned by the equivalent element-wise function :func:`~array_api.equal`.
+        -   Comparison of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
         """
 
     def __float__(self: array, /) -> float:
@@ -584,9 +583,6 @@ class _array:
         """
         Computes the truth value of ``self_i >= other_i`` for each element of an array instance with the respective element of the array ``other``.
 
-        .. note::
-           For backward compatibility, conforming implementations may support complex numbers; however, inequality comparison of complex numbers is unspecified and thus implementation-dependent (see :ref:`complex-number-ordering`).
-
         Parameters
         ----------
         self: array
@@ -599,12 +595,12 @@ class _array:
         out: array
             an array containing the element-wise results. The returned array must have a data type of ``bool``.
 
+        Notes
+        -----
 
-        .. note::
-           Element-wise results must equal the results returned by the equivalent element-wise function :func:`~array_api.greater_equal`.
-
-        .. note::
-           Comparison of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
+        -   Element-wise results must equal the results returned by the equivalent element-wise function :func:`~array_api.greater_equal`.
+        -   Comparison of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
+        -   For backward compatibility, conforming implementations may support complex numbers; however, inequality comparison of complex numbers is unspecified and thus implementation-dependent (see :ref:`complex-number-ordering`).
         """
 
     def __getitem__(
@@ -645,9 +641,6 @@ class _array:
         """
         Computes the truth value of ``self_i > other_i`` for each element of an array instance with the respective element of the array ``other``.
 
-        .. note::
-           For backward compatibility, conforming implementations may support complex numbers; however, inequality comparison of complex numbers is unspecified and thus implementation-dependent (see :ref:`complex-number-ordering`).
-
         Parameters
         ----------
         self: array
@@ -660,12 +653,12 @@ class _array:
         out: array
             an array containing the element-wise results. The returned array must have a data type of ``bool``.
 
+        Notes
+        -----
 
-        .. note::
-           Element-wise results must equal the results returned by the equivalent element-wise function :func:`~array_api.greater`.
-
-        .. note::
-           Comparison of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
+        -   Element-wise results must equal the results returned by the equivalent element-wise function :func:`~array_api.greater`.
+        -   Comparison of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
+        -   For backward compatibility, conforming implementations may support complex numbers; however, inequality comparison of complex numbers is unspecified and thus implementation-dependent (see :ref:`complex-number-ordering`).
         """
 
     def __index__(self: array, /) -> int:
@@ -769,9 +762,6 @@ class _array:
         """
         Computes the truth value of ``self_i <= other_i`` for each element of an array instance with the respective element of the array ``other``.
 
-        .. note::
-           For backward compatibility, conforming implementations may support complex numbers; however, inequality comparison of complex numbers is unspecified and thus implementation-dependent (see :ref:`complex-number-ordering`).
-
         Parameters
         ----------
         self: array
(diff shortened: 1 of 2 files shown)


asmeurer:
I think we need to update and reference this section for the specification on how scalars convert into arrays. https://data-apis.org/array-api/latest/API_specification/type_promotion.html#mixing-arrays-with-python-scalars

asmeurer:
And https://github.com/data-apis/array-api/issues/841 if fixed will also potentially change how that text reads in relation to how real and complex scalar casting works.

src/array_api_stubs/_draft/elementwise_functions.py
```diff
@@ -1339,56 +1383,60 @@ def floor_divide(x1: array, x2: array, /) -> array:
...
+    -----
+
+    -   At least one of ``x1`` or ``x2`` must be an array.
+    -   Comparison of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
+    -   For backward compatibility, conforming implementations may support complex numbers; however, inequality comparison of complex numbers is unspecified and thus implementation-dependent (see :ref:`complex-number-ordering`).
     """
 
 
-def hypot(x1: array, x2: array, /) -> array:
+def hypot(x1: Union[array, float], x2: Union[array, float], /) -> array:
```
asmeurer, line 1439: Presumably int should also be allowed here to match https://data-apis.org/array-api/latest/API_specification/type_promotion.html#mixing-arrays-with-python-scalars

> a Python int or float for real-valued floating-point array data types.

src/array_api_stubs/_draft/elementwise_functions.py
```diff
@@ -514,7 +518,7 @@ def atan(x: array, /) -> array:
     """
 
 
-def atan2(x1: array, x2: array, /) -> array:
+def atan2(x1: Union[array, float], x2: Union[array, float], /) -> array:
```
asmeurer, line 521: Same here as my `hypot` comment https://github.com/data-apis/array-api/pull/862/files#r1857364664

src/array_api_stubs/_draft/elementwise_functions.py
```diff
@@ -852,15 +883,15 @@ def conj(x: array, /) -> array:
     """
 
 
-def copysign(x1: array, x2: array, /) -> array:
+def copysign(x1: Union[array, float], x2: Union[array, float], /) -> array:
```
asmeurer, line 886: (obviously this is an exception to the "allow int in float functions" rule)

src/array_api_stubs/_draft/elementwise_functions.py
```diff
@@ -1818,15 +1869,15 @@ def log10(x: array, /) -> array:
     """
 
 
-def logaddexp(x1: array, x2: array, /) -> array:
+def logaddexp(x1: Union[array, float], x2: Union[array, float], /) -> array:
```
asmeurer, line 1872: Same comment here as with hypot https://github.com/data-apis/array-api/pull/862/files#r1857364664

src/array_api_stubs/_draft/elementwise_functions.py
```diff
@@ -2091,15 +2163,15 @@ def negative(x: array, /) -> array:
     """
 
 
-def nextafter(x1: array, x2: array, /) -> array:
+def nextafter(x1: Union[array, float], x2: Union[array, float], /) -> array:
```
asmeurer, line 2166: Ditto with copysign. int doesn't make sense in these functions because they specifically manipulate bits of floating-point numbers.

asmeurer:
`clip` is already an elementwise function that supports scalars, but we should ensure the language is consistent.

src/array_api_stubs/_draft/elementwise_functions.py
kgryte: I've matched the text, where we say that `x1` and `x2` should have a real-valued floating-point data type.

src/array_api_stubs/_draft/elementwise_functions.py
kgryte: Same response. If we allow `int`, how that gets converted to a float is undefined. Hence, the types stay true to the parameter descriptions where we specify that `x1` and `x2` should have real-valued floating-point data types.

src/array_api_stubs/_draft/elementwise_functions.py
kgryte: In the text, `x1` should have a real-valued floating-point data type, and `x2` should have the same data type as `x1`.

src/array_api_stubs/_draft/elementwise_functions.py
asmeurer: Well this is something we should decide on. Should functions that operate on floating-point inputs accept `int`? If we are matching what we wrote for scalars with operators, then yes. If we want to be more strict, then no, but then we should probably update that section (which we need to do anyway so it applies to both operators and functions).

rgommers: `int` should be accepted everywhere I think, like the link in the first comment here.

I think this has been decided on for us already by the language; `int` is being made more and more compatible with float, and Mypy et al. have decided that `int` is a subtype of `float`: https://stackoverflow.com/questions/59619778/mypy-why-is-int-a-subtype-of-float. 

In practice it will also "just work" without changes, while trying to reject `int` inputs is cumbersome.

src/array_api_stubs/_draft/elementwise_functions.py
kgryte: I went ahead and added explicit `int` support as we should be consistent across all multi-argument element-wise functions.

kgryte:
In `clip`, which is slightly different in that the scalar arguments are optional, we already added guidance concerning differing data type kinds. However, we will still need to update to point to common documentation concerning scalar casting.

ev-br:
Once python ints make the stage, I'd be ideal to explicitly spell the position on overflows. For both binary operators and elementwise functions. AFAICS, the spec only declares unspecified an attempt to cast a too-large python int to an integer dtype. However,

```
(Pdb) m32 = int(xp.finfo(xp.float32).max)
(Pdb) am32 = xp.asarray(m32, dtype=xp.float32)
(Pdb) am32
Array(3.4028235e+38, dtype=array_api_strict.float32)
(Pdb) am32 + m32
*** RuntimeWarning: overflow encountered in add
(Pdb) zzz = xp.asarray(m32*2, dtype=xp.float32)
*** RuntimeWarning: overflow encountered in cast
```

src/array_api_stubs/_draft/elementwise_functions.py
kgryte: I went ahead and updated to enumerate explicit `int` support; however, two points:

1. If `int` is already considered a `float`, it is not clear to me that saying `Union[int, float]` is even necessary. Simply because a type signature only includes `float` does not mean that downstream libraries would need to "reject" scalar values of type `int`. Meaning, the "just work" aspect should hold true whether or not we explicitly enumerate `int` as an accepted scalar type. Related: https://github.com/python/typing/issues/1746 and https://discuss.python.org/t/clarifying-the-float-int-complex-special-case/54018/30.
2. I think that (1) is symptomatic of a larger issue which is the sometimes conflicting need to have the specification also serve the role of an authoritative type implementation. For several of the functions, the "spirit" of the specification is that, e.g., a function is intended to work with real-valued floating-point numbers. We then muddy the waters a bit by including explicit type annotations which reflect more an implementation detail than the intent, especially when providing an `int` may result in undefined behavior due to casting.

On the one hand having types in the signatures is nice for Sphinx and API docs; on the other hand, I see advantages to more clearly separating specification guidance from a concrete type implementation.

rgommers: > especially when providing an `int` may result in undefined behavior due to casting.

I can't think of a way that this could ever happen given our casting rules which give Python scalars lowest priority.

kgryte: The undefined behavior would stem from cross-kind casting (e.g., an integer value which exceeds the maximum safe floating-point integer).

asmeurer:
Also regarding clip: https://github.com/data-apis/array-api/pull/811

src/array_api_stubs/_draft/array_object.py
```diff
@@ -645,9 +641,6 @@ def __gt__(self: array, other: Union[int, float, array], /) -> array:
         """
         Computes the truth value of ``self_i > other_i`` for each element of an array instance with the respective element of the array ``other``.
 
-        .. note::
-           For backward compatibility, conforming implementations may support complex numbers; however, inequality comparison of complex numbers is unspecified and thus implementation-dependent (see :ref:`complex-number-ordering`).
-
```
ev-br, line 650: I'd suggest to keep the link to https://data-apis.org/array-api/latest/design_topics/complex_numbers.html#complex-number-ordering, here and in other comparison operators.

Rationale:on a first read, I was not sure if ordering for complex value was even defined in the standard.

src/array_api_stubs/_draft/array_object.py
```diff
@@ -934,38 +921,33 @@ def __mul__(self: array, other: Union[int, float, array], /) -> array:
...
         Notes
         -----
 
-        .. note::
-           Element-wise results, including special cases, must equal the results returned by the equivalent element-wise function :func:`~array_api.not_equal`.
-
-        .. note::
-           Comparison of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
+        -   Element-wise results, including special cases, must equal the results returned by the equivalent element-wise function :func:`~array_api.not_equal`.
+        -   Comparison of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
```
ev-br, line 950:
```suggestion
        -   Comparison of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
        -  For backward compatibility, conforming implementations may support complex numbers; however, inequality comparison of complex numbers is unspecified and thus implementation-dependent (see :ref:`complex-number-ordering`).
```

src/array_api_stubs/_draft/elementwise_functions.py
```diff
@@ -2139,6 +2217,8 @@ def not_equal(x1: array, x2: array, /) -> array:
     Notes
     -----
 
+    -   At least one of ``x1`` or ``x2`` must be an array.
+
```
ev-br, line 2221:
```suggestion
    -   Comparison of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is undefined and thus implementation-dependent.
    -   For backward compatibility, conforming implementations may support complex numbers; however, inequality comparison of complex numbers is unspecified and thus implementation-dependent (see :ref:`complex-number-ordering`).
```

(copy-pasted from `less`)

src/array_api_stubs/_draft/array_object.py
kgryte: @ev-br While ordering comparisons are not defined for complex numbers, **equality** comparisons are. Hence, the note from `less` is not applicable for `ne` and `eq`.

src/array_api_stubs/_draft/elementwise_functions.py
kgryte: Same comment. Equality comparison is defined for complex numbers. However, ordering comparisons are not.

ev-br (approved):
LGTM

kgryte:
1. Common documentation. This is being addressed in a separate PR.
2. `clip`. The language is currently consistent. For that function, `x` must always be an array. Further changes are being addressed in separate PRs.

As such, I do not believe this PR has any blockers and has received one approval, I'll go ahead and merge. Any necessary updates can be addressed in follow-up PRs. Thanks everyone!

PR #865, opened 2024-11-30
Small typo in copies and views page

minerharry:
word order typo, should be "either be a [...] or own [...] who may be"


rgommers (approved):
Thanks @minerharry

PR #868, opened 2024-12-06
fix: clarify that `condition` should have a boolean dtype in `where`

lucascolley:
Towards https://github.com/data-apis/array-api-extra/issues/49, https://github.com/scipy/scipy/pull/21783

gh-116 removed this detail because it was contained in a type annotation. Unfortunately, that detail hasn't made its way back into the spec yet, which is causing trouble downstream, with array-api-strict supporting non-boolean input, but `torch` from array-api-compat only supporting boolean input.

cc @ev-br @asmeurer

### DIFF
--- a/src/array_api_stubs/_draft/searching_functions.py
+++ b/src/array_api_stubs/_draft/searching_functions.py
@@ -146,7 +146,7 @@ def where(condition: array, x1: array, x2: array, /) -> array:
     Parameters
     ----------
     condition: array
-        Should have a boolean data type. When ``True``, yield ``x1_i``; otherwise, yield ``x2_i``. Must be compatible with ``x1`` and ``x2`` (see :ref:`broadcasting`).
+        when ``True``, yield ``x1_i``; otherwise, yield ``x2_i``. Should have a boolean data type. Must be compatible with ``x1`` and ``x2`` (see :ref:`broadcasting`).
     x1: array
         first input array. Must be compatible with ``condition`` and ``x2`` (see :ref:`broadcasting`).
     x2: array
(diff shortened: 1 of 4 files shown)


ev-br:
Would be nice to add as an good addition to the agenda of the consortium meeting on Thursday. From the spec I cannot figure out if it is meant to be truly boolean or just truth/falsy (and if so, torch is not conformant)

rgommers:
> And indeed this used to be in the standard, until it was (seemingly accidentally) lost with [gh-116](https://github.com/data-apis/array-api/pull/116), as mentioned above.

That indeed seems to be completely accidental. I marked this for the 2024 milestone and backporting, will merge soon unless there are more comments, since it looks uncontroversial and caused by a minor editing hiccup only.

src/array_api_stubs/_draft/searching_functions.py
```diff
@@ -146,7 +146,7 @@ def where(condition: array, x1: array, x2: array, /) -> array:
     Parameters
     ----------
     condition: array
-        when ``True``, yield ``x1_i``; otherwise, yield ``x2_i``. Must be compatible with ``x1`` and ``x2`` (see :ref:`broadcasting`).
+        Should have a boolean data type. When ``True``, yield ``x1_i``; otherwise, yield ``x2_i``. Must be compatible with ``x1`` and ``x2`` (see :ref:`broadcasting`).
```
kgryte, line 149:
```suggestion
        when ``True``, yield ``x1_i``; otherwise, yield ``x2_i``. Should have a boolean data type. Must be compatible with ``x1`` and ``x2`` (see :ref:`broadcasting`).
```

Minor editorial change, as we discuss details such as expected dtype, shape, and broadcast compatibility after the initial description.

kgryte (approved):
LGTM! Thanks, @lucascolley!

lucascolley:
https://github.com/data-apis/array-api-strict/pull/106 implements this in xp-strict

Issue #478, opened 2022-09-11
Complex type promotion rules are unclear


leofang:
[This table](https://data-apis.org/array-api/draft/API_specification/type_promotion.html#floating-point-type-promotion-table) needs to be updated to include mixed real and complex dtypes (making it a 4x4 table). Currently the complex type promotion is completely unspecified, and the [lattice diagram](https://github.com/data-apis/array-api/blob/main/spec/_static/images/dtype_promotion_complex.png) (which is currently missing, see #477) is not very helpful.

rgommers:
@leofang I opened gh-491 with a table with the rules. It looks like the main thing to discuss on this issue is:

> [This table](https://data-apis.org/array-api/draft/API_specification/type_promotion.html#floating-point-type-promotion-table) needs to be updated to include mixed real and complex dtypes (making it a 4x4 table).

That looks to me like cross-kind promotion, which is in general not mandated to exist. This is analogous to integer and real mixed-kind promotion. Libraries may implement it, but not supporting it is also fine.

Am I missing something here?

seberg:
I would argue that mixed real (floats) and complex should be allowed and defined.  The "kind" issues really pop up with promoting uints, ints, or "inexact" (using inexact as numpy to denote both real and complex).

rgommers:
> The "kind" issues really pop up with promoting uints, ints, or "inexact" (using inexact as numpy to denote both real and complex).

This is only part of the picture. Another major reason is that cross-kind casting may not be implemented in existing libraries. Sometimes by design, for example to avoid unintended type promotions to blow up memory usage silently, or to limit the implementation effort or binary size (for binary ops, you need a lot of kernels for combinations of supported dtypes).

That said, I just checked and JAX and PyTorch fully support mixed real-complex dtype promotion. That leaves only TensorFlow, which is the most limited library in this respect. But I'd say that likely this shouldn't stop us either way.

rgommers:
As I expected, TensorFlow doesn't allow this:
```python
>>> import tensorflow.experimental.numpy as tnp
>>> x = tnp.ones(3, dtype=tnp.float32)
>>> y = tnp.zeros(3, dtype=tnp.complex64)
>>> x + y
Traceback (most recent call last):
  Cell In [21], line 1
    x + y
  File ~/mambaforge/envs/tf/lib/python3.9/site-packages/tensorflow/python/util/traceback_utils.py:153 in error_handler
    raise e.with_traceback(filtered_tb) from None
[...]

PR #871, opened 2024-12-10
feat: define complex scalar <op> fp array

ev-br:
Closes gh-841, which continued from https://github.com/data-apis/array-api/issues/478. If I read the discussion right, there is a general support for allowing binary operations between a python ``complex`` scalar and a floating-point array.

The proposed wording tries to be explicit about preserving the precision: `(1j*single_precision_array).dtype == single_precision_complex` and `(1j*double_precision_array).dtype == double_precision_complex`. As long as we don't have to worry about half- or extended precision, being explicit feels simplest.

### DIFF
--- a/spec/draft/API_specification/type_promotion.rst
+++ b/spec/draft/API_specification/type_promotion.rst
@@ -132,7 +132,7 @@ where ``<op>`` is a built-in operator (including in-place operators, but excludi
 
 -   a Python ``bool`` for a ``bool`` array data type.
 -   a Python ``int`` within the bounds of the given data type for integer array :ref:`data-types`.
--   a Python ``int`` or ``float`` for real-valued floating-point array data types.
+-   a Python ``int``, ``float``, or ``complex`` for real-valued floating-point array data types.
 -   a Python ``int``, ``float``, or ``complex`` for complex floating-point array data types.
 
 Provided the above requirements are met, the expected behavior is equivalent to:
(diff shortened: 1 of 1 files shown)


spec/draft/API_specification/type_promotion.rst
```diff
@@ -143,6 +143,10 @@ Provided the above requirements are met, the expected behavior is equivalent to:
...
 
-   Similarly, behavior is not specified when mixing a Python ``complex`` and an array with a real-valued data type; this may give ``complex64``, ``complex128``, or raise an exception. Behavior is implementation-specific.
-
    Behavior is also not specified for integers outside of the bounds of a given integer data type. Integers outside of bounds may result in overflow or an error.
+
+
+.. note::
+    Mixing a Python ``complex`` and an array with a real-valued data type is allowed; the result is an array of the data type of the same *precision* as the array operand. In other word, ``1j <op> float32_array`` results in a ``complex64`` array; ``1j <op> float64_array`` results in a ``complex128`` array.
+
+   Behavior is not specificed when mixing ``complex`` scalars with integer arrays; this may give ``complex64``, ``complex128``, or raise an exception. Behavior is implementation-specific.
```
lucascolley, line 152: would we want to add a `versionchanged` directive for this?

kgryte: Typically, I add all version changed directives at the same time after compiling the changelog, not on a PR-by-PR basis. While it is thus far common for us to release a December version, this isn't guaranteed, so I have typically avoided adding such annotations until the spec release has been finalized.

lucascolley (approved):

kgryte:
Made a couple of small tweaks, but otherwise this LGTM. Thanks, @ev-br and @lucascolley.

kgryte (approved):

Issue #397, opened 2022-02-28
Capitalizing RFC 2119 words like SHOULD


asmeurer:
@tomwhite suggested at https://github.com/dask/dask/pull/8750#discussion_r815757335 that we should capitalize RFC 2119 words like "SHOULD" in the spec to make it clearer that we are following the RFC. 

Another option might be to bold the words instead of capitalizing them.

kgryte:
I think the option of using bold text should suffice. While all-caps is a common convention, it does add a fair amount of visual noise. I think encoding RFC terms using bold text is something we can address in the current draft specification and then continue to enforce moving forward. For prior revisions, we can leave as is.

rgommers:
That seems like a reasonable option. I'd prefer to avoid capitalized words yell at me:)

asmeurer:
An idea to automatically bold those words (I haven't tested it). https://chat.openai.com/share/7467ec88-0d8f-4352-9a5b-5b43fe6b8f02

PR #872, opened 2024-12-10
format RFC 2119 keywords

ev-br:
aims to address https://github.com/data-apis/array-api/issues/397, https://github.com/data-apis/array-api/issues/796

typeset RFC 2119 magic keywords (should/must/must not) in lowercase italics; the preference in gh-397 seems to avoid ALLCAPS. 

No specific preference for italics either, can change to boldface---did not do it just yet because boldface is already in use in phrases like `The full **singular value decomposition** of an :math:`m \times n` matrix ...`

Either way, if there's a strong preference for boldface or other, it's easy to change, just slightly tweak the `sed` stanzas under the fold.

Note that I only spot-checked the diff, so it's possible that the search-and-replace changed too much.

<details>

```
    $ sed -E -i  "/not/! s/must/\*must\*/g" ./*.rst
    $ sed -E -i  "s/must not/\*must not\*/g" ./*.rst
    $ sed -E -i  "/not/! s/should/\*should\*/g" ./*.rst
```

</details>

### DIFF
--- a/spec/draft/API_specification/array_object.rst
+++ b/spec/draft/API_specification/array_object.rst
@@ -5,12 +5,12 @@ Array object
 
     Array API specification for array object attributes and methods.
 
-A conforming implementation of the array API standard must provide and support an array object having the following attributes and methods.
+A conforming implementation of the array API standard *must* provide and support an array object having the following attributes and methods.
 
-Furthermore, a conforming implementation of the array API standard must support, at minimum, array objects of rank (i.e., number of dimensions) ``0``, ``1``, ``2``, ``3``, and ``4`` and must explicitly document their maximum supported rank ``N``.
+Furthermore, a conforming implementation of the array API standard *must* support, at minimum, array objects of rank (i.e., number of dimensions) ``0``, ``1``, ``2``, ``3``, and ``4`` and *must* explicitly document their maximum supported rank ``N``.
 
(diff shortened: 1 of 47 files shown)


rgommers:
I noticed that gh-801 tries doing the same thing. Could you compare and recommend which approach to use?

ev-br:
Ah, indeed. gh-801 looks more polished of course. Can't see what the rendered docs look like since the link has expired and CircleCI needs a kick from somebody with the right permissions (does not seem to react to my prods).

Assuming the rendered docs in gh-801 are what we want them to be, the choice boils down to whether we prefer to maintain a dedicated Sphinx extension (gh-801) or run `sed` once (this PR) and remember to bold/italicize the RFC words in new prose/docstrings. 

My personal preference would be for the latter, but I'm not the main writer here.

kgryte:
IMO, the approach in this PR is preferred when compared to https://github.com/data-apis/array-api/pull/801. The main issue which I raised there is that changes should actually be constrained to normative content.

At least to start, I recommend only modifying the use of `should` and `must` in the stubs files. In design topics, e.g., we are not as careful with our language.

ev-br:
OK, reverted all changes to `spec/`, so that only files in `src/array_api_stubs/_draft/*` are affected.

PR #873, opened 2024-12-12
feat: add support for scalars in `result_type`

### LINKED ISSUE #805: RFC: add support for scalar arguments in `result_type`
The array API's [type promotion rules](https://data-apis.org/array-api/latest/API_specification/type_promotion.html#mixing-arrays-with-python-scalars) support mixed scalar/array operations, e.g., `1 + xp.arange(3)`.

For Xarray, we would like to be able to figure out the resulting dtype from this sort of operation before actually doing it (https://github.com/pydata/xarray/pull/8946).

Ideally, we could use `xp.result_type()` for this purpose, but as documented [result_type only supports arrays and dtype objects](https://data-apis.org/array-api/latest/API_specification/generated/array_api.result_type.html#array_api.result_type). Could we potentially extend `result_type` to also handle Python scalars? It is worth noting that this already works today in NumPy, e.g.,
```
>>> np.result_type(1, np.arange(3))
dtype('int64')
```

kgryte:
This PR:

- closes https://github.com/data-apis/array-api/issues/805 by adding scalar support to `result_type`.
- requires that at least one argument be an array or dtype object (ref: https://github.com/data-apis/array-api/issues/805#issuecomment-2117651614).
- moves a note to a dedicated notes section.

### DIFF
--- a/src/array_api_stubs/_draft/data_type_functions.py
+++ b/src/array_api_stubs/_draft/data_type_functions.py
@@ -209,20 +209,25 @@ def isdtype(
     """
 
 
-def result_type(*arrays_and_dtypes: Union[array, dtype]) -> dtype:
+def result_type(
+    *arrays_and_dtypes: Union[array, int, float, complex, bool, dtype]
+) -> dtype:
     """
-    Returns the dtype that results from applying the type promotion rules (see :ref:`type-promotion`) to the arguments.
-
-    .. note::
-       If provided mixed dtypes (e.g., integer and floating-point), the returned dtype will be implementation-specific.
+    Returns the dtype that results from applying type promotion rules (see :ref:`type-promotion`) to the arguments.
 
     Parameters
     ----------
-    arrays_and_dtypes: Union[array, dtype]
-        an arbitrary number of input arrays and/or dtypes.
+    arrays_and_dtypes: Union[array, int, float, complex, bool, dtype]
+        an arbitrary number of input arrays, scalars, and/or dtypes.
 
     Returns
     -------
     out: dtype
-        the dtype resulting from an operation involving the input arrays and dtypes.
+        the dtype resulting from an operation involving the input arrays, scalars, and/or dtypes.
+
+    Notes
+    -----
+
+    -   At least one argument must be an array or a dtype.
+    -   If provided array and/or dtype arguments having mixed data type kinds (e.g., integer and floating-point), the returned dtype is unspecified and is implementation-dependent.
     """


lucascolley (approved):

kgryte:
As this PR has received one approval and follows changes proposed and discussed during workgroup meetings, I'll go ahead and merge. Any further refinements can be made in follow-up PRs. Thanks for reviewing, @lucascolley!

mdhaber:
Looking at the underlying issue, it doesn't seem to have come up, but I thought I'd check before opening a new one - was there an explicit decision to allow *instances* of Python scalars but NOT the types themselves (i.e., `bool`, `int`, `float`, `complex`)? It is easy enough to work around, but the feature seems natural to support, and not standardizing it allows gotchas like https://github.com/numpy/numpy/issues/29821#issuecomment-4063343360, in which the types follow different promotion rules than the instances (without a use-case in mind).

kgryte:
@mdhaber The spec does allow dtypes. See https://data-apis.org/array-api/latest/API_specification/generated/array_api.result_type.html.

Maybe I am misunderstanding your question.

mdhaber:
A strict reading of the standard suggests that this will fail: 
```python3
import array_api_strict as xp
xp.result_type(xp.asarray(1), float)
```
and it does.
```
TypeError: result_type() inputs must be Array API arrays, dtypes, or scalars
```
`float` is not a dtype or a scalar; it is a Python `type`.

kgryte:
@mdhaber I added to the upcoming workgroup agenda.

ev-br:
A bit of due diligence for the support across array libraries:

```
In [5]: cupy.result_type(cupy.ones(3), float)
Out[5]: dtype('float64')

In [6]: torch.result_type(torch.ones(3), float)
TypeError                                 Traceback (most recent call last)
Cell In[6], line 1
----> 1 torch.result_type(torch.ones(3), float)

TypeError: result_type() received an invalid combination of arguments - got (Tensor, type), but expected one of:
 * (Tensor tensor, Tensor other)
      didn't match because some of the arguments have invalid types: (Tensor, type)
 * (Number scalar, Tensor tensor)
      didn't match because some of the arguments have invalid types: (Tensor, type)
 * (Tensor tensor, Number other)
      didn't match because some of the arguments have invalid types: (Tensor, type)
 * (Number scalar1, Number scalar2)
      didn't match because some of the arguments have invalid types: (Tensor, type)

In [7]: jnp.result_type(jnp.ones(3), float)
Out[7]: dtype('float32')

In [8]: np.result_type(np.ones(3), float)
Out[8]: dtype('float64')

```

rgommers:
I commented on the NumPy issue: I think @mdhaber makes a good point there. For the standard however, if NumPy is inconsistent with itself and might be changing behavior (or not), and if PyTorch doesn't support passing in the `int`/`float` types themselves, then probably it's unwise to change anything in the standard in the near future?

asmeurer:
I might be misunderstanding the use-case here, but you can easily go from a Python scalar type to an instance by calling it with no arguments

```py
>>> int()
0
>>> float()
0.0
>>> bool()
False
>>> complex()
0j
```

mdhaber:
Yes, it's easy to work around if you know to expect it. It just seemed accidental to accept arrays, scalars, and array dtypes but *not* scalar types. And it's surprising that the standard implicitly permits the behavior for scalars and scalar types to be different; this gotcha could easily lead to dtype bugs.

Issue #804, opened 2024-05-03
xp.linalg.vecdot should have axis=-1 default


oleksandr-pavlyk:
`xp.linalg.vecdot` is documented as alias to `xp.vecdot`. 

However, `xp.linalg.vecdot` signature uses `axis=None` default, while `xp.vecdot` has `axis=-1` as the default. 

![image](https://github.com/data-apis/array-api/assets/21087696/925ff22f-673d-4f28-b309-36ae7b6c595b)

![image](https://github.com/data-apis/array-api/assets/21087696/89ac3099-25af-4b5a-9c44-5485cf814c7d)

They should be consistent.

kgryte:
Hmmm...this should have been fixed already: https://github.com/data-apis/array-api/issues/355. The regression happened during https://github.com/data-apis/array-api/commit/cbbab62922ab7ddd4f31302c21f22d2db62d6f16. I can work on a fix and backport.

PR #880, opened 2025-01-09
fix: address default value regression for `axis` kwarg in `vecdot`

kgryte:
This PR:

- closes: https://github.com/data-apis/array-api/issues/804
- addresses a regression introduced in https://github.com/data-apis/array-api/commit/cbbab62922ab7ddd4f31302c21f22d2db62d6f16.
- was previously fixed in https://github.com/data-apis/array-api/pull/358
- backports changes to previous specification versions.

### DIFF
--- a/src/array_api_stubs/_2021_12/linalg.py
+++ b/src/array_api_stubs/_2021_12/linalg.py
@@ -474,7 +474,7 @@ def trace(x: array, /, *, offset: int = 0) -> array:
     """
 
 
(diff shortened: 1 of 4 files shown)


kgryte:
As this PR fixes a regression from a previous fix, I'll go ahead and merge. This PR simplify makes the defaults consistent between `vecdot` and `linalg.vecdot` where the default `axis` should be `-1`.

Issue #852, opened 2024-11-05
diff() does not specify type promotion behavior


asmeurer:
`diff()` potentially takes multiple arrays as input (`x`, `prepend`, and `append`). But it doesn't specify if they should type promote together, or if the standard is only defined when `prepend` and `append` have the same data type as `x` 

https://data-apis.org/array-api/draft/API_specification/generated/array_api.diff.html#diff

kgryte:
In the current specification guidance, we state that `prepend` and `append` should have the same data type as `x`. Given that `prepend` and `append` are kwarg-only and we have yet, TMK, specify an API in which array kwargs may affect the output dtype, I do not think we should allow type promotion such that the output array may be of a different dtype than `x`.

However, I think we can clarify, in a note, that if an array library chooses to support `prepend` and `append` which are not of the same dtype as `x`, then those arrays should promote to the dtype as `x`. The output array should always have the same data type as `x`.

PR #881, opened 2025-01-09
docs: clarify type promotion behavior in `diff`

kgryte:
This PR:

- closes: https://github.com/data-apis/array-api/issues/852
- clarifies that, when either `prepend` and `append` are of a different dtype than `x`, behavior is left unspecified, allowing type promotion, casting, or raising an exception.
- maintains the status quo that the only portable behavior is when `append` and `prepend` have the same data type as `x`.

## Notes

NumPy supports type promotion:

```python
In [14]: x = np.ones(10,dtype="float32")

In [15]: x.dtype
Out[15]: dtype('float32')

In [16]: np.diff(x).dtype
Out[16]: dtype('float32')

In [17]: np.diff(x,prepend=np.ones(10))
Out[17]:
array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0.])

In [18]: np.diff(x,prepend=np.ones(10)).dtype
Out[18]: dtype('float64')
```

We have yet, TMK, allowed optional array kwargs to affect the output array dtype. It is not obvious to me that we should establish such a precedent here.

### DIFF
--- a/src/array_api_stubs/_draft/utility_functions.py
+++ b/src/array_api_stubs/_draft/utility_functions.py
@@ -125,5 +125,5 @@ def diff(
     -----
 
     -   The first-order differences are given by ``out[i] = x[i+1] - x[i]`` along a specified axis. Higher-order differences must be calculated recursively (e.g., by calling ``diff(out, axis=axis, n=n-1)``).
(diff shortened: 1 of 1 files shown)


rgommers:
> clarifies that, when either `prepend` and `append` are of a different dtype than `x`, behavior is left unspecified, thus allowing array libraries to opt for either type promotion, casting, or raising an exception.

That seems like the right decision to me, and aligns with what was discussed in the community meeting yesterday. It's also the most flexible decision - in case a real-world use case for something more prescriptive regarding type promotion arises in the future, we can deal with it then.

kgryte:
As this change is a relatively minor documentation change, was discussed and agreed upon during a prior workgroup meeting, and has not received any objections, I'll go ahead and merge.

Issue #830, opened 2024-07-31
Minor clarification on allowed rounding mode


hpkfft:
The accuracy section of this standard states
```
The results of element-wise arithmetic operations ...
for floating-point operands must return the nearest representable value
according to IEEE 754-2019 and a supported rounding mode.
```
I believe this intends to allow (optionally, and not by default) the use of directed rounding, i.e., the IEEE 754-2019 rounding attributes `roundTowardPositive`, `roundTowardNegative`, and `roundTowardZero`.
However, directed rounding does not, in general, return the nearest representable value--only an attribute that rounds to nearest does that.
IEEE 754-2019 defines the term "correctly rounded", which I suggest is preferred here:
```
The results of element-wise arithmetic operations ...
for floating-point operands must return the correctly rounded value
according to IEEE 754-2019 and a supported rounding mode.
```

hpkfft:
By the way, I would support a recommendation that array libraries offer directed rounding as an option.
One way to try to assess the effects of roundoff upon a floating-point computation is to repeat the computation in arithmetic of the same precision but rounded differently, say TowardNegative, TowardPositive, and maybe TowardZero, besides "to nearest", and compare three or four results.  I'm quoting Prof. W. Kahan.
```
Though far from foolproof, rounding every inexact arithmetic operation (but not constants)
in the same direction for each of two or three directions besides the default "to nearest"
is very likely to confirm accidentally exposed hypersensitivity to roundoff.
When feasible, this scheme offers the best Benefit/Cost ratio [of the five he mentions].
```
The paper is [Mindless.pdf](https://people.eecs.berkeley.edu/~wkahan/Mindless.pdf).

kgryte:
Ref: https://github.com/data-apis/array-api/pull/882. This PR should close this issue.

Re: directed rounding. Agreed, this would be ideal; however, I do not believe any array libraries (e.g., NumPy et al) support this at the moment. We can revisit adding such language to the spec once there is ecosystem support and real-world implementations.

Issue #826, opened 2024-07-28
Specify correct rounding for sqrt


hpkfft:
In accuracy.rst, correct rounding is required for add, sub, mul, and divide.   I propose adding sqrt to this list.
Note that sqrt is not listed later among the mathematical functions whose accuracy is not precisely defined.

Note that the IEEE Std 754-2019 for Floating-Point Arithmetic mandates correct rounding for `squareRoot`.

asmeurer:
Is this something that's actually the case for existing libraries/devices? If so, there shouldn't be an issue with adding it. 

By the way, is your proposal for this to apply to both real and complex inputs or just to real inputs?

hpkfft:
Excellent point!  My proposal is for this to apply just to real inputs.  Given industry adoption of IEEE Std 754-2019, correct rounding **is** actually the case for existing libraries/devices.
Sometimes there is an option for less accuracy (and presumably higher performance).   For example, CUDA in single precision requires compiling with `-prec-sqrt=true`, and division requires `-prec-div=true`, for correct rounding.  Since this document requires correctly rounded division, it equally well ought to require correctly rounded square root.  [Double precision CUDA divide and square root are always correctly rounded.   See the "Mathematical Functions" section of the "CUDA C++ Programming Guide".]

hpkfft:
I believe the existing accuracy requirements for element-wise arithmetic operations are only intended to apply to real inputs.
For example, the usual 6 flop implementation of complex multiplication
```
(a + bi)(c + di) = (ac - bd) + (ad + bc)i
```
does not result in correctly rounded real and imaginary components.
I do not believe this array standard means to prohibit this implementation, but this should be clarified.

kgryte:
> I do not believe this array standard means to prohibit this implementation

Yes, I think this is correct. As the specification advocates for IEEE 754 compliance (with some caveats; e.g., subnormals) and IEEE 754 only applies to reals, we don't have an equivalent mandate for complex number operations. Accordingly, we should make this distinction explicit.

Re: `sqrt`. Yes, I think this is a reasonable recommendation. The main point for the arithmetic operations being correctly rounded is to limit error accumulation. As `sqrt` is a fundamental operation--and more fundamental than the various transcendentals--I agree that requiring correctly rounded behavior is reasonable and should be added.

hpkfft:
For those interested in the accuracy of complex operators/functions, my colleagues and I have done a comparison with different
[...]

PR #882, opened 2025-01-09
docs: clarify that `sqrt` must be correctly rounded in accordance with IEEE 754

kgryte:
This PR:

- closes https://github.com/data-apis/array-api/issues/826 and closes https://github.com/data-apis/array-api/issues/830
- clarifies that `sqrt` should follow IEEE 754 and always return correctly rounded result. This was implied (i.e., conforming implementations should be IEEE 754 compliant), but never made explicit.
- clarifies that accuracy requirements apply to real-valued floating-point operands and not complex-valued floating-point operands.
- adds missing functions to list of functions not covered by accuracy requirements.
- specifies that subnormals may or may not be supported.

### DIFF
--- a/spec/draft/design_topics/accuracy.rst
+++ b/spec/draft/design_topics/accuracy.rst
@@ -25,6 +25,8 @@ including the corresponding element-wise array APIs defined in this standard
 
 for real-valued floating-point operands must return the correctly rounded value according to IEEE 754-2019 and a supported rounding mode. By default, the rounding mode should be ``roundTiesToEven`` (i.e., round to nearest with ties rounded toward the nearest value with an even least significant bit).
 
+IEEE 754-2019 requires support for subnormal (a.k.a., denormal) numbers, which are useful for supporting gradual underflow. However, hardware support for subnormal numbers is not universal, and many platforms (e.g., accelerators) and compilers support toggling denormals-are-zero (DAZ) and/or flush-to-zero (FTZ) behavior to increase performance and to guard against timing attacks. Accordingly, conforming implementations may vary in their support for subnormal numbers.
+
 Mathematical Functions
 ----------------------
 
--- a/spec/draft/design_topics/accuracy.rst
+++ b/spec/draft/design_topics/accuracy.rst
@@ -23,7 +23,7 @@ including the corresponding element-wise array APIs defined in this standard
 -   multiply
 -   divide
 
-for real-valued floating-point operands must return the correctly rounded value according to IEEE 754-2019 and a supported rounding mode. By default, the rounding mode should be ``roundTiesToEven`` (i.e., round to nearest with ties rounded toward the nearest value with an even least significant bit).
+for real-valued floating-point operands must return a correctly rounded value according to IEEE 754-2019 and a supported rounding mode. By default, the rounding mode should be ``roundTiesToEven`` (i.e., round to nearest with ties rounded toward the nearest value with an even least significant bit).
 
 IEEE 754-2019 requires support for subnormal (a.k.a., denormal) numbers, which are useful for supporting gradual underflow. However, hardware support for subnormal numbers is not universal, and many platforms (e.g., accelerators) and compilers support toggling denormals-are-zero (DAZ) and/or flush-to-zero (FTZ) behavior to increase performance and to guard against timing attacks. Accordingly, conforming implementations may vary in their support for subnormal numbers.
 
@@ -35,7 +35,7 @@ The results of the following functions
 -   reciprocal
 -   sqrt
 
-for real-valued floating-point operands must return the nearest representable value according to IEEE 754-2019 and a supported rounding mode.
+for real-valued floating-point operands must return a correctly rounded value according to IEEE 754-2019 and a supported rounding mode.
 
 This specification does **not** precisely define the behavior of the following functions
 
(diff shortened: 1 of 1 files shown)


hpkfft:
The default for single-precision divide in CUDA isn't to round with this precision either.
Nevertheless, this spec did the right thing by requiring CUDA libraries to be compiled with the flag that enables correct rounding.
It should be equally useful/achievable for `sqrt`.

spec/draft/design_topics/accuracy.rst
```diff
@@ -23,11 +23,18 @@ including the corresponding element-wise array APIs defined in this standard
 -   multiply
 -   divide
 
-for floating-point operands must return the nearest representable value according to IEEE 754-2019 and a supported rounding mode. By default, the rounding mode should be ``roundTiesToEven`` (i.e., round to nearest with ties rounded toward the nearest value with an even least significant bit).
+for real-valued floating-point operands must return the nearest representable value according to IEEE 754-2019 and a supported rounding mode. By default, the rounding mode should be ``roundTiesToEven`` (i.e., round to nearest with ties rounded toward the nearest value with an even least significant bit).
```
hpkfft, line 26: I suggest changing `nearest representable` to `correctly rounded`.
If the rounding mode is not `roundTiesToEven` the value will be correctly rounded, but in general, it will not be the nearest representable.  For example, if the rounding mode is `roundTowardPositive`, then the correctly rounded value is rounded up regardless of whether the rounded down value is nearer to the infinitely precise mathematical result.

spec/draft/design_topics/accuracy.rst
```diff
@@ -23,11 +23,18 @@ including the corresponding element-wise array APIs defined in this standard
...
 
 Mathematical Functions
 ----------------------
 
+The results of the following functions
+
+-   reciprocal
+-   sqrt
+
+for real-valued floating-point operands must return the nearest representable value according to IEEE 754-2019 and a supported rounding mode.
```
hpkfft, line 36: I suggest changing `nearest representable` to `correctly rounded`.

leofang:
> Given [#826 (comment)](https://github.com/data-apis/array-api/issues/826#issuecomment-2257211013) says that the default for single-precision `sqrt` in CUDA isn't to round with this precision, I'm not sure how useful/achievable this will be. @leofang any comments on that?

The default is to do correct rounding, as @hpkfft noted. However, it involves to not set `-ftz=true` which is a requirement we've discussed to avoid, ex: https://github.com/data-apis/array-api/blob/532db5baeb7d47f28e6b0e3f0ede2e180bc0688f/src/array_api_stubs/_draft/elementwise_functions.py#L1478-L1481

Therefore, I do not think the "correctly rounded" requirement is achievable. Certainly it is [not CuPy's default](https://github.com/cupy/cupy/blob/5b7423d844a800c38f8fcf21235475bfba45f130/cupy/cuda/compiler.py#L550).

Ref: https://docs.nvidia.com/cuda/floating-point/index.html#compiler-flags

leofang (changes requested):
(see above)

hpkfft:
Denormal result support (as opposed to flush to zero (`ftz`)) is orthogonal.  It's not specific to square root, but rather applies to everything: addition, subtraction, ....
Yes, flushing to zero gives zero, which is neither the correctly rounded result nor the nearest representable value.
I think the note makes it clear that this is a global exception to requiring IEEE 754-2019 conformance.  (Of course, suggestions to improve/clarify the wording are welcome if you feel that would be helpful.)
Otherwise, when the result is normal, I think it would be strange for this spec to deviate from the IEEE 754-2019 requirement of correct rounding for add, sub, mul, div, and sqrt.   As @kgryte pointed out, conformance was already implied by this spec, and this PR merely makes it explicit.

hpkfft:
Oh, I think the note you referenced is not in the spec itself?
I agree that the spec should explicitly mention that denormal support is *not* a requirement, i.e., `ftz` and/or `daz` are acceptable.  [The former applies to denormal results and the latter applies to denormal inputs.]

But, I would suggest that your observation ought to be considered a separate bug/PR and not block this PR.
It might be the case that denormal support requires discussion.

leofang:
Hi @hpkfft @kgryte, sorry for my late reply. I reached out to our math team (at NVIDIA) asking for clarification, and I believe what @hpkfft stated above is generally correct. Denormals can be flushed while still keeping the rounding behavior correct, at least for `sqrt` this is documented in the PTX instruction: 
https://docs.nvidia.com/cuda/parallel-thread-execution/#floating-point-instructions-sqrt
(both `sqrt.rnd.f32` and `sqrt.rnd.ftz.f32` are considered IEEE compliant). With this understanding, @kgryte could you please kindly apply @hpkfft's suggestions, then we can approve/merge?

spec/draft/design_topics/accuracy.rst
kgryte: Updated.

spec/draft/design_topics/accuracy.rst
kgryte: Done.

kgryte:
@leofang and @hpkfft I've updated the PR accordingly. If you can give it a once over, that would be appreciated!

hpkfft (approved):
Looks great!    Thank you.

kgryte:
Thanks for the review @hpkfft! As the changes implement the suggestions discussed above, I'll go ahead and merge. Thanks all!

leofang:
Thanks to you both 🙏

Issue #824, opened 2024-07-18
RFC: revisit requirement that inputs to `real`, `imag`, and `conj` be complex-valued


mdhaber:
gh-427 summarizes the rationale for requiring that inputs to `real`, `imag`, and `conj` be complex-valued:

> This PR chooses to be more restrictive based on the premise that consumers should be explicit when performing a complex number operation (e.g., if real-valued inputs are allowed, was calling `real(x)` intentional when providing a real-valued `x` or was this an unintended bug where `x` should have been complex valued?). Given potential ambiguity between intended versus unintended behavior, this PR places the onus on consumers to resolve this ambiguity.

https://github.com/data-apis/array-api/pull/446#discussion_r991577991 states that this is open to being revisited.

If the concern is ambiguity between intended and unintended use, was there a similar discussion for functions that are no-ops for integers, such as `round` and `isinf`? It looks like the initial [commit](https://github.com/data-apis/array-api/pull/12/commits/f3d5cd6352b74941a45af5ab067133cf88dd55c4) that added rounding functions already included the special case for integers, but I couldn't find a discussion about it. 

Perhaps it didn't come up because it is common for code paths that work on floating point dtypes to also be used for integer dtypes. In the same way, I'd suggest that code for complex dtypes can also be valid for real dtypes. If it is not easy to tell whether unintended or intended uses of these functions on real data is more common in the wild, perhaps the precedent set by NumPy and other libraries (e.g. CuPy, JAX, PyTorch, Dask Array, Tensorflow) is not wrong, and these functions can allow real input.

asmeurer:
Has this come up in real-world code? I can definitely see wanting to write functions that work on both real and complex dtypes. But I also am curious how often a complex implementation can be directly reused for a real implementation.

mdhaber:
Yes. I have a complex number that is the logarithm of a negative number. I take the real component to get the magnitude.

asmeurer:
Wouldn't `real(log(x))` return `nan` if `x` is a negative real?

mdhaber:
It would if `x` has a real dtype. Let's expand the definition of "real" to include complex dtype with zero imaginary component.

asmeurer:
There must be a misunderstanding here. `real` is already supposed to work on any array with a complex64 or complex128 dtype, regardless of what the value of the imaginary component is. The above discussion was specifically about whether it should work on float32 and float64 arrays.

asmeurer:
[...]

PR #884, opened 2025-01-18
feat: add support for real-valued arrays in `real` and `conj`

ev-br:
Closes https://github.com/data-apis/array-api/issues/824 by allowing `real` and `conj` to accept real-valued arrays, and leaving `imag` to still require complex-valued arguments.

One possibly thorny issue is whether `xp.real(x)` is a view or a copy of of `x`. This is also discussed in gh-824, not 100% sure what the resolution is though.
In this PR, err on the side of caution and require that both `real` and `conj` always return copies. 

Prior art is not very consistent AFAICS: numpy can return either a view or a copy depending on the dtype, not sure what other libraries do. 

OTOH, 

1. if we always require a copy, we can consider allowing `imag` to create a new array of zeros.
2. or maybe we can leave it unspecified and implementation dependent.

### DIFF
--- a/src/array_api_stubs/_draft/elementwise_functions.py
+++ b/src/array_api_stubs/_draft/elementwise_functions.py
@@ -881,8 +881,9 @@ def conj(x: array, /) -> array:
 
     .. versionadded:: 2022.12
 
-    If ``x`` array has a real floating-point dtype, the returned array is a copy of ``x``.
-
+    A conforming implementation may return a "view" into the input array or a copy. Therefore, whether
+    the output array and the input array share the underlying memory buffer is unspecified and
+    thus implementation-defined.
     """
 
 
@@ -2361,8 +2362,9 @@ def real(x: array, /) -> array:
 
     .. versionadded:: 2022.12
 
-    If ``x`` array has a real floating-point dtype, the returned array is a copy of ``x``.
-
+    A conforming implementation may return a "view" into the input array or a copy. Therefore, whether
+    the output array and the input array share the underlying memory buffer is unspecified and
+    thus implementation-defined.
     """
 
 
--- a/src/array_api_stubs/_draft/elementwise_functions.py
+++ b/src/array_api_stubs/_draft/elementwise_functions.py
@@ -869,7 +869,7 @@ def conj(x: array, /) -> array:
     Parameters
     ----------
     x: array
-        input array. Should have a floating-point data type.
+        input array. Should have a numeric data type.
 
     Returns
     -------
@@ -2350,7 +2350,7 @@ def real(x: array, /) -> array:
     Parameters
     ----------
     x: array
-        input array. Should have a floating-point data type.
+        input array. Should have a numeric data type.
 
     Returns
     -------
--- a/src/array_api_stubs/_draft/elementwise_functions.py
+++ b/src/array_api_stubs/_draft/elementwise_functions.py
@@ -869,7 +869,7 @@ def conj(x: array, /) -> array:
     Parameters
     ----------
     x: array
-        input array. Should have a complex floating-point data type.
+        input array. Should have a floating-point data type.
 
     Returns
     -------
@@ -880,6 +880,9 @@ def conj(x: array, /) -> array:
     -----
 
     .. versionadded:: 2022.12
+
+    If ``x`` array has a real floating-point dtype, the returned array is a copy of ``x``.
+
     """
 
 
@@ -2346,7 +2349,7 @@ def real(x: array, /) -> array:
     Parameters
     ----------
     x: array
-        input array. Should have a complex floating-point data type.
+        input array. Should have a floating-point data type.
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


kgryte:
@ev-br Thanks for opening this PR. My sense is that we should leave it unspecified and implementation-dependent. Forcing a copy is likely a no-go for libraries leveraging strided arrays and focused on performance.

ev-br:
Agreed. Trying it on numpy, it seems that `np.conj` returns a copy, while `np.real` returns a view --- at least for 1D inputs. What it does in higher-dimensions or for arrays with more complicated stride arrangements, who knows. Mandating this seems premature unless all major array libraries agree on something coherent (which is to say, never). 

Nonetheless, allowing `xp.{conj, real}(real_valued_array)` still makes sense IMO: if only for clarity when writing mathematical code where 1) arrays can legitimately be either real floating or complex, and 2) copy-vs-view semantics is not relevant. This is not an edge case. The status quo requires that everybody adds explicit branching on the array dtypes; the proposed wording makes it that only those who want to mutate (or prohibit mutating) the result of `xp.real` need to copy or branch in the user code. I would expect this to be rare, and the vast majority of the user code benefits.

Therefore, the last commit updates the wording to make the view/semantics explicitly unspecified.

rgommers:
> Mandating this seems premature unless all major array libraries agree on something coherent (which is to say, never).

A view is not a thing in the standard (xref https://data-apis.org/array-api/latest/design_topics/copies_views_and_mutation.html), so I don't think we will ever want to mandate anything here. View vs. copy is only observable when mutating aliased memory (which should be avoided) and otherwise a performance optimization that isn't visible in syntax or semantics.

kgryte:
In principle, we *could* support a `copy` kwarg on `real` and `conj` to support forcing a copy. We do this for `astype` and a few other functions.

src/array_api_stubs/_draft/elementwise_functions.py
```diff
@@ -869,7 +869,7 @@ def conj(x: array, /) -> array:
     Parameters
     ----------
     x: array
-        input array. Should have a complex floating-point data type.
+        input array. Should have a floating-point data type.
```
kgryte, line 872:
```suggestion
        input array. Should have a numeric data type.
```

We don't need to limit to only floating-point dtypes if providing a real-valued array is simply a no-op.

src/array_api_stubs/_draft/elementwise_functions.py
```diff
@@ -2346,7 +2350,7 @@ def real(x: array, /) -> array:
...
```
kgryte, line 2353:
```suggestion
        input array. Should have a numeric data type.
```

Same comment. There is a clearer case here for allowing any numeric data type.

ev-br:
My only wish is that there's _some_ wording, however mysterious, to show that the spec thought about the semantics and left it to be implementation-defined. For me as a user even "if you plac the output to the left of the assignment sign, the result is implementation-defined" would have been sufficient (and having no note would be worse).

Re: forcing a copy with a keyword, I'd suggest to to have a minimal version in 2024.12 spec for downstream to use, and refine based on usage.
In principle, if the wording in the Notes proves contentious, punting on it --- thus limiting this PR to just "real-valued arrays are allowed" --- would be great at least for scipy.signal.

src/array_api_stubs/_draft/elementwise_functions.py
ev-br: A quick check.... yes, agreed:

```
In [47]: t = torch.arange(4, dtype=torch.uint8)

In [48]: torch.conj_physical(t).dtype
Out[48]: torch.uint8

In [49]: torch.real(t).dtype
Out[49]: torch.uint8

In [50]: a = np.arange(3, dtype=np.uint8)

In [51]: np.real(a).dtype
Out[51]: dtype('uint8')

In [52]: np.conj(a).dtype
Out[52]: dtype('uint8')

```

kgryte:
@ev-br I tweaked the language a bit, removing explicit mention of "views", and I also updated the guidance to require numeric input types. The previous guidance was "should" in order to allow libraries supporting real-valued dtypes to continue doing so. Now that we're expanding the scope of these functions to all numeric dtypes that wiggle room is no longer necessary.

src/array_api_stubs/_draft/elementwise_functions.py
```diff
@@ -879,6 +879,8 @@ def conj(x: array, /) -> array:
     Notes
     -----
 
+    -   Whether or not the returned array and the input array share the same underlying memory is unspecified and thus implementation-defined.
```
kgryte, line 882:
```suggestion
    -   Whether the returned array and the input array share the same underlying memory is unspecified and thus implementation-defined.
```

src/array_api_stubs/_draft/elementwise_functions.py
```diff
@@ -2356,6 +2358,8 @@ def real(x: array, /) -> array:
...
```
kgryte, line 2361:
```suggestion
    -   Whether the returned array and the input array share the same underlying memory is unspecified and thus implementation-defined.
```

kgryte (approved):

PR #885, opened 2025-01-20
feat: add `dtype` kwarg support to `fftfreq` and `rfftfreq`

kgryte:
This PR

- closes https://github.com/data-apis/array-api/issues/717 by adding `dtype` kwarg support to `fft.fftfreq` and `fft.rfftfreq`.
- aligns `fftfreq` and `rfftfreq` with other creation functions in allowing one to specify an output data type. Currently, these two FFT APIs have no means of inferring an output dtype precision, and thus default to the default real-valued floating-point data type.
- limits what dtypes can be specified to real-valued floating-point data types.

### DIFF
--- a/src/array_api_stubs/_draft/fft.py
+++ b/src/array_api_stubs/_draft/fft.py
@@ -15,7 +15,7 @@ __all__ = [
     "ifftshift",
 ]
 
-from ._types import Tuple, Union, Sequence, array, Optional, Literal, device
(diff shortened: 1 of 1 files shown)


ev-br (approved):
LGTM

kgryte:
As this was discussed in the last workgroup meeting and received no objections, I'll go ahead and merge. Thanks, @ev-br for the review!

Issue #495, opened 2022-10-12
RFC: add `copy`


jakirkham:
Some array libraries implement a `.copy()` method ([like NumPy]( https://numpy.org/doc/stable/reference/generated/numpy.copy.html )). While there are some indirect ways to get at this now (`asarray`, `reshape`, etc.), currently the API lacks a way to do this directly and without other potential side-effects. Should add a method (unlike a function) would ensure the new array has the original array's type simply. Curious if there is appetite for including this in the API.

Related is a question of what interplay there is with [`__copy__` and/or `__deepcopy__`]( https://docs.python.org/3/library/copy.html ) (if any).

rgommers:
Thanks @jakirkham. A couple of thoughts:
- NumPy has both a function and a method. The function has an extra keyword (`subok`)
- Either way, the new keywords don't seem appropriate, so it would simply be `copy(x)` or `x.copy()`
- At that point, it's the same as `copy.copy` and `copy.deepcopy` I"d think (at least for NumPy)
- PyTorch calls it [clone](https://pytorch.org/docs/stable/generated/torch.clone.html#torch.clone), it doesn't have `copy`
- Is there a reason to add it? Autograd or compilers related perhaps - better than the stdlib function?

asmeurer:
Isn't this `asarray(x, copy=True)`? Does that have side-effects?

jakirkham:
To summarize the ask, library functions handling general arrays may want to copy as they want an array they can mutate safely without affecting user provided data.

We concluded that we want a function (not a method) and one needs to do a namespace lookup (`x.__array_namespace__`) to figure it out (as the type is likely not known by the library).

There is a separate question of what we call it. Either a new function (`copy`, `clone`) or use an existing one (`asarray(x, copy=True)`). We would also want to spell out that some libraries (Dask, JAX, maybe others) may not actually copy the underlying data.

`__copy__` and `__deepcopy__` are different enough (Dask would copy graphs, JAX something similar, etc.) that it is worth specifying this path may not copy the array data (if that is the user's concern) and that using the function above (name to be decided) would be preferred for data copying.

asmeurer:
[...]

Issue #866, opened 2024-12-05
Disambiguate what copy=True means for dask


crusaderky:
The current documentation for the `copy` parameter of `asarray` states:
https://data-apis.org/array-api/latest/API_specification/generated/array_api.asarray.html#asarray

> copy (Optional[bool]) – boolean indicating whether or not to copy the input. If True, the function must always copy. If False, the function must never copy for input which supports the buffer protocol and must raise a ValueError in case a copy would be necessary. If None, the function must reuse existing memory buffer if possible and copy otherwise. Default: None.

The meaning of `copy=True` is ambiguous for dask.
There are two possible interpretations:
1. updating the output array will never alter the input array; OR
2. the output array will never share memory with the input array; in other words dereferencing the input array can release its memory even if you still hold a reference to the output array.

In dask, updating a collection actually creates brand new graph nodes under the hood and repoints the collection to those nodes, so the original is never modified. However, the original chunks, generated e.g. by `from_array`, are still held inside the graph.

I strongly prefer the first definition, as IMHO decisions around memory management should be considered low level and delegated to the individual libraries.

rgommers:
> I strongly prefer the first definition, as IMHO decisions around memory management should be considered low level and delegated to the individual libraries.

Yes agreed. In all such cases we've cared about the semantics, not about physical memory layout. A library should always be able to reuse memory or other such optimizations if it guarantees that that does not affect the semantics.

I'm pretty that using definition (1) is not controversial, and I remember clarifying this in some other place in the docs already - just can't remember where.

ev-br:
If anything, it'd be quite helpful to clarify what copy=None means for dask, https://github.com/data-apis/array-api-compat/issues/209

crusaderky:
> If anything, it'd be quite helpful to clarify what copy=None means for dask, [data-apis/array-api-compat#209](https://github.com/data-apis/array-api-compat/issues/209)

We could change "if possible" to "if possible and sensible"...

asmeurer:
> I'm pretty that using definition (1) is not controversial, and I remember clarifying this in some other place in the docs already - just can't remember where.

https://github.com/data-apis/array-api/issues/788#issuecomment-2080416220

rgommers:
Thanks for the link @asmeurer!

[...]

Issue #788, opened 2024-04-15
RFC: Consider updating `copy` semantics in `astype` to `False/None/True`


Micky774:
Currently the specification for `astype`, as added in https://github.com/data-apis/array-api/pull/290, specifies that `copy=False/True` wherein there is no room for a "copy never" option, and the default `copy=True` means that calls to `astype` that do not specify the `copy` kwarg never have a chance to be a no-op -- this results in unnecessary copies as _default_ behavior.

Other functions which use the `copy=False/None/True` semantics include:
- `asarray`
- `__dlpack__`
- `from_dlpack`
- `reshape`

As far as I can tell, `astype` is the exception to the rule.

kgryte:
The default specification behavior follows NumPy and its derivatives. From the [NumPy docs](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.astype.html),

> By default, astype always returns a newly allocated array.

and the signature

```
ndarray.astype(dtype, order='K', casting='unsafe', subok=True, copy=True)
```

The default is `True`, not `None`. So the original specification simply followed suit.

jakevdp:
If NumPy is the motivation for this, then I think there may be a problem: NumPy has explicitly decided to differ from the Array API in this respect; see https://numpy.org/neps/nep-0056-array-api-main-namespace.html#copy-keyword-semantics

kgryte:
My reading from that NEP is not that NumPy is choosing to differ. NumPy is following the spec where `copy` kwarg behavior for `astype` is not the same as for `asarray` and others.

However, the situation is actually reversed. We based the current spec on what NumPy and its kin did at the time, which was to, by default, always copy.

jakevdp:
I see, thanks.

jakevdp:
In NumPy's implementation, do `copy=None` and `copy=False` have the same semantics of "copy only if necessary"?

rgommers:
> In NumPy's implementation, do `copy=None` and `copy=False` have the same semantics of "copy only if necessary"?

Yes. But that's only because there was resistance to changing `copy=False` behavior to match the `asarray` semantics, and it didn't seem nearly as important as `asarray`. So `astype(..., copy=False)` still means "if needed" not "never".

rgommers:
Maybe it'll be changed on a longer timescale, I don't know.

Micky774:
[...]

PR #886, opened 2025-01-23
docs: clarify `copy` kwarg behavior in `asarray` for libraries disallowing mutation

kgryte:
This PR:

- addresses discussion in https://github.com/data-apis/array-api/issues/495 by clarifying that, for libraries disallowing mutation, when provided known array input, `copy=True` can be ignored.
- moves notes to a dedicated notes section, rather than interleaved between API descriptions.

The principle ask from https://github.com/data-apis/array-api/issues/495 was to have an explicit API to guarantee an output array which can be safely mutated. `asarray(x, copy=True)` satisfies that ask; however, for libraries such as JAX, performing an explicit copy is unnecessary, as mutation is disallowed by design. This PR provides an escape hatch for such libraries by allowing them to disregard `copy=True` when provided known array input.


kgryte:
Moving to draft, as the language in this PR is related and likely dependent on the outcome of other discussions:

- https://github.com/data-apis/array-api/issues/866
- https://github.com/data-apis/array-api/issues/788

PR #887, opened 2025-01-23
fix: resolve conflicting exception guidance in `__dlpack__`

### LINKED ISSUE #831: Minor clarification for conflicting information in `__dlpack__`
The specification page for [`__dlpack__`](https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__dlpack__.html) requires the following two, conflicting things:

for the `dlpack_device` argument

> The v2023.12 standard only mandates that a compliant library should offer a way for __dlpack__ to return a capsule referencing an array whose underlying memory is accessible to the Python interpreter (represented by the kDLCPU enumerator in DLPack). **If a copy must be made to enable this support but copy is set to False, the function must raise ValueError.**

for the `copy` argument

> If False, the function must never copy, and **raise a BufferError in case a copy is deemed necessary** (e.g. if a cross-device data movement is requested, and it is not possible without a copy).

These appear to be mutually exclusive and in direct conflict. It would be nice to clarify the correct intention

kgryte:
This PR:

- resolves #831 by fixing the conflicting guidance regarding raised exceptions in `__dlpack__`.
- backports the fix to the `2023.12` revision of the standard.

### DIFF
--- a/src/array_api_stubs/_2023_12/array_object.py
+++ b/src/array_api_stubs/_2023_12/array_object.py
@@ -364,7 +364,7 @@ class _array:
             ``__dlpack__`` to return a capsule referencing an array whose underlying memory is
             accessible to the Python interpreter (represented by the ``kDLCPU`` enumerator in DLPack).
             If a copy must be made to enable this support but ``copy`` is set to ``False``, the
-            function must raise ``ValueError``.
+            function must raise ``BufferError``.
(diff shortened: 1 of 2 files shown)


kgryte:
As this is a minor textual change and addresses an oversight, I'll go ahead and merge.

Issue #823, opened 2024-07-17
RFC: clarify `broadcast_to` semantics


adityagoel4512:
I'm finding the ``broadcast_to`` [specification](https://data-apis.org/array-api/latest/API_specification/generated/array_api.broadcast_to.html#broadcast-to) a little underspecified. In the docs we see the following for the shape parameter:

> shape (Tuple[int, ...]) – array shape. Must be compatible with x (see [Broadcasting](https://data-apis.org/array-api/latest/API_specification/broadcasting.html#broadcasting)). If the array is incompatible with the specified shape, the function should raise an exception.

The broadcasting link goes on to specify **bidirectional broadcasting**. That would imply to me that ``np.broadcast_to(np.asarray([[-1, -1], [-1, -1]]), (2, 1, 2))`` should work since shapes (2, 2) and (2, 1, 2) are bidirectionally compatible. Somewhat reasonably in my opinion, NumPy did not interpret this in that way and raises an exception.

Since ``np.broadcast_to(np.asarray([[-1, -1], [-1, -1]]), (1, 2, 2))`` does work, it seems that broadcasting compatibility is unidirectional. i.e. `x.shape` must be broadcastable to `shape`. Is it worth spelling out explicitly the difference in how this works, [like ONNX does](https://github.com/onnx/onnx/blob/main/docs/Broadcasting.md#unidirectional-broadcasting)? I couldn't find any explanation in the standard itself.

It does say the following, although I read the "a specified shape" part as "any shape" rather than simply the `shape` parameter.

> Returns:
out (array) – an array having a specified shape. Must have the same data type as x.

If this ambiguity is shared I am happy to contribute a clarification.

asmeurer:
I agree it should be updated. The unidirectional broadcasting is important for in-place operators (see https://data-apis.org/array-api/latest/API_specification/broadcasting.html#in-place-semantics).

cbourjau:
@asmeurer While I'm not aware of a read-only concept in the array API, I just wanted to point out that NumPy's `broadcast_to` does place that restriction on the returned array: https://numpy.org/doc/stable/reference/generated/numpy.broadcast_to.html#numpy-broadcast-to

asmeurer:
Read-only isn't a concept that's in the array API. Not all libraries might implement it. The array API leaves all mutation with views undefined so this isn't an issue.

seberg:
[...]

PR #888, opened 2025-01-23
fix: clarify broadcast behavior in `broadcast_to`

kgryte:
This PR:

- closes https://github.com/data-apis/array-api/issues/823 by clarifying that (a) broadcast compatibility only applies to `x` being capable of being broadcast to a specified shape and (b) the returned array must have **the** specified shape.

## Questions

1. Is this worth backporting to prior revisions?

### DIFF
--- a/src/array_api_stubs/_draft/manipulation_functions.py
+++ b/src/array_api_stubs/_draft/manipulation_functions.py
@@ -42,14 +42,14 @@ def broadcast_to(x: array, /, shape: Tuple[int, ...]) -> array:
     Parameters
     ----------
     x: array
-        array to broadcast.
+        array to broadcast. Must be capable of being broadcast to the specified ``shape`` (see :ref:`broadcasting`). If the array is incompatible with the specified shape, the function should raise an exception.
     shape: Tuple[int, ...]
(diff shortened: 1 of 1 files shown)


kgryte:
As is a minor clarification and has been discussed on the referenced issue, I'll go ahead and merge. If we need to backport, we can do so in a follow-up PR.

src/array_api_stubs/_draft/manipulation_functions.py
```diff
@@ -42,14 +42,14 @@ def broadcast_to(x: array, /, shape: Tuple[int, ...]) -> array:
     Parameters
     ----------
     x: array
-        array to broadcast.
+        array to broadcast. Must be capable of being broadcast to the specified ``shape`` (see :ref:`broadcasting`). If the array is incompatible with the specified shape, the function should raise an exception.
```
seberg, line 45: Sorry never looked it, but a small nit: it should be "must raise an error".

kgryte: Opened https://github.com/data-apis/array-api/pull/897 to resolve.

PR #889, opened 2025-01-23
feat!: apply type promotion rules according to the device context

kgryte:
This PR:

- closes https://github.com/data-apis/array-api/issues/672 by clarifying that device info must be taken into account when applying type promotion rules in `result_type` and `can_cast`. As not all devices can support all data types, both functions need to explicitly account for whether an array or dtype can be cast and/or promoted to a desired dtype on a given device.
- when provided only dtypes, both functions are required to only consider the full type promotion graph.
- when at least one argument is an array, both functions must account for device capabilities.
- for `result_type`, if two or more array arguments belong to separate devices, behavior is left unspecified and thus implementation-dependent.

### DIFF
--- a/src/array_api_stubs/_draft/data_type_functions.py
+++ b/src/array_api_stubs/_draft/data_type_functions.py
@@ -65,7 +65,7 @@ def astype(
 
 def can_cast(from_: Union[dtype, array], to: dtype, /) -> bool:
     """
-    Determines if one data type can be cast to another data type according type promotion rules (see :ref:`type-promotion`).
+    Determines if one data type can be cast to another data type according to type promotion rules (see :ref:`type-promotion`).
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


src/array_api_stubs/_draft/data_type_functions.py
```diff
@@ -65,7 +65,7 @@ def astype(
 
 def can_cast(from_: Union[dtype, array], to: dtype, /) -> bool:
     """
-    Determines if one data type can be cast to another data type according :ref:`type-promotion` rules.
+    Determines if one data type can be cast to another data type according type promotion rules (see :ref:`type-promotion`).
```
kgryte, line 68:
```suggestion
    Determines if one data type can be cast to another data type according to type promotion rules (see :ref:`type-promotion`).
```

kgryte:
As this was previously discussed during workgroup meetings, I'll go ahead and merge. If any further changes and clarifications are necessary, we can address in follow-up PRs.

PR #896, opened 2025-02-06
docs: clarify `clip` behavior when arguments have different data types

kgryte:
This PR

- follows up on https://github.com/data-apis/array-api/pull/814#issuecomment-2639228207 and https://github.com/data-apis/array-api/issues/807 and supersedes https://github.com/data-apis/array-api/pull/811 to indicate that behavior is only defined when `min` and `max` have the same data type as `x`. After reviewing SciPy and scikit-learn usage of `xp.clip` and `np.clip`, the overwhelming usage was using scalar `min` and `max`. For all instances that I found with array `min` and `max`, type promotion behavior was not expected. Given the various edge cases when `min` and `max` are not the same type as `x`, it seems prudent to limit portable behavior to the scenario which avoids edge cases altogether.
- clarifies the definition of `clamp` as suggested in https://github.com/data-apis/array-api/pull/814#issuecomment-2428088370.
- specifies behavior when `min` and/or `max` are scalar values.
- clarifies three-way broadcasting as done in https://github.com/data-apis/array-api/pull/811.
- relaxes the output array data type requirement in order to accommodate libraries such as NumPy which support type promotion behavior and cannot adopt a stricter implementation due to backward compatibility concerns.
- backports a typo fix to the v2023 revision.

### DIFF
--- a/src/array_api_stubs/_draft/elementwise_functions.py
+++ b/src/array_api_stubs/_draft/elementwise_functions.py
@@ -831,7 +831,7 @@ def clip(
     Returns
     -------
     out: array
-      an array containing element-wise results. The returned array must have the same data type as ``x``.
+      an array containing element-wise results. The returned array should have the same data type as ``x``.
 
     Notes
     -----
--- a/src/array_api_stubs/_draft/elementwise_functions.py
+++ b/src/array_api_stubs/_draft/elementwise_functions.py
@@ -841,7 +841,7 @@ def clip(
     - If a broadcasted element in ``min`` is greater than a corresponding broadcasted element in ``max``, behavior is unspecified and thus implementation-dependent.
     - For scalar ``min`` and/or ``max``, the scalar values should follow type promotion rules for operations involving arrays and scalar operands (see :ref:`type-promotion`). Hence, if ``x`` and either ``min`` or ``max`` have different data type kinds (e.g., integer versus floating-point), behavior is unspecified and thus implementation-dependent.
     - If ``x`` has an integral data type and a broadcasted element in ``min`` or ``max`` is outside the bounds of the data type of ``x``, behavior is unspecified and thus implementation-dependent.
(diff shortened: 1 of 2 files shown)


src/array_api_stubs/_draft/elementwise_functions.py
```diff
@@ -836,10 +836,12 @@ def clip(
     Notes
     -----
 
+    - This function is conceptually equivalent to ``maximum(minimum(x, max), min)`` when ``x``, ``min``, and ``max`` have the same data type.
+    - For scalar ``min`` and/or ``max``, the scalar values must be converted to zero-dimensional arrays having the same data type as ``x`` prior to broadcasting.
```
rgommers, line 840: This sounds a bit like an implementation requirement, rather than only clarifying the intended semantics. It's also not what implementations do for mixed-kind dtypes:
```python
>>> np.clip(np.arange(3, dtype=np.int32), -1, 1.5).dtype
dtype('float64')
```

I suggest saying instead that for scalars regular type promotion rules have to be followed (and hence mixed-kind is undefined).

rgommers (commented):
Thanks @kgryte. Looks good overall, one minor comment about phrasing how Python scalars are handled.

src/array_api_stubs/_draft/elementwise_functions.py
kgryte: Yeah, but wouldn't that be functionally the same? According to type promotion rules involving scalars and arrays, the first step is to convert the scalar to the same dtype as the array. When I included the line about conversion, that was what I had in mind.

kgryte: Okay. I've updated the text to hopefully address your concerns regarding outside-the-standard behavior.

kgryte: ...and I needed to relax the output array data type requirement.

PR #897, opened 2025-02-06
fix: resolve typo in `broadcast_to` to require an exception

kgryte:
This PR

- addresses the comment https://github.com/data-apis/array-api/pull/888/files#r1944366064 by changing `should` to `must`. This was an oversight in the original specification, as there isn't a likely scenario in which a library would want to do anything other than raise an exception.


kgryte:
cc @seberg

kgryte:
As this PR addresses an oversight and is a minor textual change, I'll go ahead and merge...

Issue #861, opened 2024-11-25
[BUG]: `searchsorted` has incorrect boundary conditions


mdhaber:
The "index conditions" of `searchsorted`, which are documented in the description of the `side` argument, are:

![image](https://github.com/user-attachments/assets/fe616cdd-292a-475e-8da3-7999aacc2311)

In "Let `x` be an array of rank `N`":
- I think `x` is supposed to be `x2`, and
- it looks like `N` is _not_ supposed to be the `N` defined later as the "number of elements in `x1`".

Also, it looks like these conditions don't cover the edge cases where either
- `side='left'` and `v` exceeds all elements of `x1` or
- `side='right'` and `v` is less than all elements of `x1`:

```python3
import array_api_strict as xp
x1 = xp.asarray([1, 2, 3])

# `side='left'` and `v` exceeds all elements of `x1` or
x2 = xp.asarray(4)
i = xp.searchsorted(x1, x2, side='left')
x1[i]  # IndexError: index 3 is out of bounds for axis 0 with size 3

# `side='right'` and `v` is less than all elements of `x1`:
x2 = xp.asarray(0)
i = xp.searchsorted(x1, x2, side='right')
x1[i-1] < x1[i]
# Array(False, dtype=array_api_strict.bool)
```

I think complete index conditions, which describe the `out` mathematically, would be helpful! Then, perhaps the details of the `side` argument could be restricted to what happens when `v` "lands exactly on an edge", which I interpret to mean "is equal to an element of `x1`": If `side == 'left'`, then `v = x1[i]`; if `side == 'right'`, then `v = x1[i-1]`.

kgryte:
PR: https://github.com/data-apis/array-api/pull/898

PR #898, opened 2025-02-06
fix: address incorrect boundary conditions in `searchsorted`

kgryte:
This PR

- closes https://github.com/data-apis/array-api/issues/861 by fixing boundary conditions in `searchsorted`. The original specification incorrectly described behavior when an index fails to satisfy the index condition. This bug has been addressed and backported to the v2023 revision of the standard.
- fixes typos in the original specification and current draft.

cc @mdhaber

### DIFF
--- a/src/array_api_stubs/_2023_12/searching_functions.py
+++ b/src/array_api_stubs/_2023_12/searching_functions.py
@@ -109,15 +109,10 @@ def searchsorted(
 
         Let ``x2`` be an array of rank ``N`` where ``v`` is an individual element given by ``v = x2[n,m,...,j]``.
 
-        If ``side == 'left'``, then
-
-        - each returned index ``i`` must satisfy the index condition ``x1[i-1] < v <= x1[i]``.
-        - if no index satisfies the index condition, then the returned index for that element must be ``0``.
-
-        Otherwise, if ``side == 'right'``, then
-
-        - each returned index ``i`` must satisfy the index condition ``x1[i-1] <= v < x1[i]``.
-        - if no index satisfies the index condition, then the returned index for that element must be ``M``, where ``M`` is the number of elements in ``x1``.
+        - If ``v`` is less than all elements in ``x1``, then the returned index for that element must be ``0``.
+        - If ``v`` is greater than all elements in ``x1``, then the returned index for that element must be ``M``, where ``M`` is the number of elements in ``x1``.
+        - If ``side == 'left'``, then each returned index ``i`` must satisfy the index condition ``x1[i-1] < v <= x1[i]``.
+        - If ``side == 'right'``, then each returned index ``i`` must satisfy the index condition ``x1[i-1] <= v < x1[i]``.
 
         Default: ``'left'``.
     sorter: Optional[array]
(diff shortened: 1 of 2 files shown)


src/array_api_stubs/_2023_12/searching_functions.py
```diff
@@ -107,17 +107,11 @@ def searchsorted(
     side: Literal['left', 'right']
         argument controlling which index is returned if a value lands exactly on an edge.
 
-        Let ``x`` be an array of rank ``N`` where ``v`` is an individual element given by ``v = x2[n,m,...,j]``.
+        Let ``x2`` be an array of rank ``N`` where ``v`` is an individual element given by ``v = x2[n,m,...,j]``.
```
mdhaber, line 110: I don't think `N` is defined or referred to before or again, so it might be fine to just omit that.
Similarly for `n`, `m`, and `j`. To be precise, it will help to define one of these and refer to it later, so:
```suggestion
        Let ``v`` be an element of ``x2`` given by ``v = x2[j]``, where ``j`` refers to a valid scalar or tuple index.
```
(If you feel it's necessary to state that scalar index would be for a 1-D array and tuple for multidimensional, that's fine, too.)

src/array_api_stubs/_2023_12/searching_functions.py
```diff
@@ -107,17 +107,11 @@ def searchsorted(
...
-        If ``side == 'left'``, then
-
-        - each returned index ``i`` must satisfy the index condition ``x1[i-1] < v <= x1[i]``.
-        - if no index satisfies the index condition, then the returned index for that element must be ``0``.
-
-        Otherwise, if ``side == 'right'``, then
-
-        - each returned index ``i`` must satisfy the index condition ``x1[i-1] <= v < x1[i]``.
-        - if no index satisfies the index condition, then the returned index for that element must be ``N``, where ``N`` is the number of elements in ``x1``.
+        - If ``v`` is less than all elements in ``x1``, then the returned index for that element must be ``0``.
```
mdhaber, line 112:
```suggestion
        - If ``v`` is less than all elements in ``x1``, then ``out[j]`` must be ``0``.
```

src/array_api_stubs/_2023_12/searching_functions.py
```diff
@@ -107,17 +107,11 @@ def searchsorted(
...
+        - If ``v`` is greater than all elements in ``x1``, then the returned index for that element must be ``M``, where ``M`` is the number of elements in ``x1``.
```
mdhaber, line 113:
```suggestion
        - If ``v`` is greater than all elements in ``x1``, then ``out[j]`` must be ``M``, where ``M`` is the number of elements in ``x1``.
```
Could also consider ``x1.size`` for concision.
```suggestion
        - If ``v`` is greater than all elements in ``x1``, then ``out[j]`` must be ``x1.size``.
```

src/array_api_stubs/_2023_12/searching_functions.py
```diff
@@ -107,17 +107,11 @@ def searchsorted(
...
+        - Otherwise, if ``side == 'left'``, then each returned index ``i`` must satisfy the index condition ``x1[i-1] < v <= x1[i]``. If ``side == 'right'``, then each returned index ``i`` must satisfy the index condition ``x1[i-1] <= v < x1[i]``.
```
mdhaber, line 114:
```suggestion
        - Otherwise, if ``side == 'left'``, then each returned index ``i = out[j]`` must satisfy the index condition ``x1[i-1] < v <= x1[i]``. If ``side == 'right'``, then each returned index ``i`` must satisfy the index condition ``x1[i-1] <= v < x1[i]``.
```

mdhaber (commented):
This is much better. For maximal clarity within the structure you have proposed, I have a few suggestions.

kgryte:
Thanks, @mdhaber, for your review. As all suggestions were accepted, I'll go ahead and merge.

Issue #669, opened 2023-07-31
RFC: indexing with multi-dimensional integer arrays


shoyer:
I'd like to revisit the discussion from https://github.com/data-apis/array-api/issues/177 about adding support for indexing with arrays of indices, because this is by far the largest missing gap in functionality required for Xarray.

My specific proposal would be to standardize indexing with a tuple of length equal to the number of array dimensions, where each element in the indexing tuple is either (1) an integer or (2) an integer array. This avoids all the [messy edge cases](https://numpy.org/neps/nep-0021-advanced-indexing.html#mixed-indexing) in NumPy related to mixing slices & arrays.

The last time we talked about it, integer indexing via `__getitem__` somehow got held up on discussions of mutability with `__setitem__`. Mutability would be worth resolving at some point, for sure, but IMO is a rather different topics.

leofang:
> where each indexer is either (1) an integer or (2) an integer array.

What's an "indexer"? Each element in the tuple? Or the entire tuple?

seberg:
Do you suggest a dedicated function for this?  Or is the `oindex` vs normal advanced indexing behavior not a problem?

shoyer:
> > where each indexer is either (1) an integer or (2) an integer array.
> 
> What's an "indexer"? Each element in the tuple? Or the entire tuple?

Clarified -- each element in the tuple should be an integer or integer array.

shoyer:
> Do you suggest a dedicated function for this? Or is the `oindex` vs normal advanced indexing behavior not a problem?

I think we should standardize on a subset of vectorized indexing which is common to both `vindex` and standard NumPy advanced indexing.

`oindex` is certainly occasionally useful, but vectorized indexing can substitute -- and isn't too much harder to implement once you understand how it works.

asmeurer:
Was there ever an analysis done of what integer array indexing the different array libraries support? I don't think it would show up in the existing API comparison data because that data only looks at function definitions.

kgryte:
> Was there ever an analysis done of what integer array indexing the different array libraries support?

No, not in explicit detail.

kgryte:
Based on the history of NEP 21 (https://numpy.org/neps/nep-0021-advanced-indexing.html) and the following related discussions

- Discussion: <https://mail.python.org/pipermail/numpy-discussion/2015-April/072550.html> (the first half of the thread is most relevant)
- Discussion: <https://github.com/numpy/numpy/pull/6256>
[...]

Issue #864, opened 2024-11-28
RFC: assignment via integer array indexing


crusaderky:
[EDIT] this issue is about the lack of `xp.put` or equivalent `__setitem_` semantics for integar array indices. Read comments below.

[original post]
The current situation when an index is an array of ints or bools is very messy:

- numpy blindly passes every unexpected object it finds to np.asarray, which makes it accept np.ndarray, lists, tuples, but also memoryviews and anything with an `__array__` interface
- Sparse accepts lists, tuples, or numpy arrays; other sparse arrays don't work
- PyTorch won't accept numpy arrays of unsigned integers
- PyTorch won't accept PyTorch arrays of integers with dtype other than the native int
- JAX won't accept lists or tuples
- dask won't accept tuples

I think the array API standard should define a reasonable common surface.

xref https://github.com/data-apis/array-api-compat/pull/205#discussion_r1861288136

NOTE: the much more complicated case of multiple fancy indices, e.g. `a[[0, 1], [1, 2]]` is out of scope for this issue.

rgommers:
Just a quick note that `take` is the standardized/uniform way of doing indexing with an integer array (xref gh-416 and the issue that PR links to).

You're completely right that it's very messy across libraries - it's just very hard, since we can't mandate that libraries stop doing something and break backwards compatible (or technically we can, but it'll probably just be ignored rather than implemented).

crusaderky:
> Just a quick note that `take` is the standardized/uniform way of doing indexing with an integer array (xref [gh-416](https://github.com/data-apis/array-api/pull/416) and the issue that PR links to).

I couldn't find `put` or anything similar though? How do you update along a fancy index?

ev-br:
cross-ref https://github.com/data-apis/array-api/pull/900/ which aims to actually standardize it.

ev-br:
Looks like gh-900 covered this request, and we can close it as completed, I'd think.

ev-br:
Would be great to update the issue title/test then, if we repurpose it for `__setitem__` specifically.

amacati:
Ah right, then maybe this is a topic for array-api-extra instead. Because this code snippet should be extremely common in most frameworks:

```python
import array_api_strict as xp
import jax.numpy as jp

x = jp.array([[1, 2, 3], [4, 5, 6]])
idx = jp.array([0])
# This could be emulated in array-api-extra with xpx.at(x)[idx, ...] = xp.asarray([[7, 8, 9]])
[...]

PR #900, opened 2025-02-17
feat: add support for integer array indexing

kgryte:
This PR:

- closes https://github.com/data-apis/array-api/issues/669 by adding support for integer array indexing to the specification.
- constrains support to indexing tuples which combine integers and integer arrays.
- Punts supporting tuples containing integer arrays along with ellipses and slices to a future revision of the standard.
- specifies the behavior of "vectorized" indexing, as implemented in NumPy et al.
- Does **not** limit integer arrays to only one-dimension, even though Dask does not support multi-dimensional integer arrays.

## Notes

- Based on test results in https://github.com/data-apis/array-api-tests/pull/341, NumPy, CuPy, PyTorch, JAX, and Dask all implement at least a limited form of vectorized indexing. Dask does not currently support providing a multi-dimensional integer array in order to return an array having more than one-dimension.

## References

- Additional context: https://numpy.org/neps/nep-0021-advanced-indexing.html


kgryte:
cc @shoyer

spec/draft/API_specification/indexing.rst
```diff
@@ -173,6 +179,43 @@ Multi-dimensional arrays must extend the concept of single-axis indexing to mult
...
+
+- If ``T`` consists of only integers and zero-dimensional integer arrays, the result must be equivalent to indexing multiple axes using integer indices (see :ref:`indexing-multi-axis`).
+
+- When ``T`` contains a one-dimensional integer array ``J`` having shape ``S2 = (m,)``, where ``m`` is greater than or equal to the number of elements in any other one-dimensional integer array in ``T``, each element of ``T`` must be broadcast to the same shape as ``J``.
+
+- An ``IndexError`` exception must be raised if any element of ``T`` is not broadcast-compatible with ``J`` (see :ref:`broadcasting`).
+
+- After broadcasting, ``T`` must be equivalent to a tuple ``U = (u1, u2, ..., uN)`` containing only one-dimensional integer arrays having shape ``S2``.
+
+- Let ``v_i`` be the tuple formed by the integer indices ``u1[i], u2[i], ..., uN[i]``. When applying the indexing tuple ``U`` to ``A``, the resulting array must be a one-dimensional array having shape ``(m,)`` and containing each element ``A[v_i]`` for ``i`` on the half-open interval ``[0, m)``.
```
ev-br, line 214: This reads as if I have an N-dimensional array A, I index it with a tuple of 1D integer arrays  and the result is a 1D array?

ev-br:
One other thing I don't immediately see in this addition is whether this relates go `__getitem__` only or also  `__setitem__`.

kgryte:
> Do any of the above libraries fail this test?

@jni Dask seems to fail with a multi-dimensional integer array, but will work with one-dimensional integer arrays. @ev-br Can confirm.

spec/draft/API_specification/indexing.rst
```diff
@@ -173,6 +179,43 @@ Multi-dimensional arrays must extend the concept of single-axis indexing to mult
 
     *Rationale: this is consistent with bounds-checking for single-axis indexing. An implementation may choose to set the axis (dimension) size of the result array to* ``0`` *, raise an exception, return junk values, or some other behavior depending on device requirements and performance considerations.*
 
+Integer Array Indexing
+----------------------
+
+An array must support indexing a one-dimensional array by a one-dimensional integer array according to the following rules. Let ``A`` be a one-dimensional array with shape ``S1 = (n,)``, and let ``J`` be a one-dimensional integer array with ``S2 = (k,)``.
```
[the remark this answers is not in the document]
kgryte: > The wording here only allows arrays, and does not allow lists. Would be best to stress this explicitly.

I've updated the text to allow sequences equivalent to integer arrays.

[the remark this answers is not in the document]
kgryte: > So we only allow J.ndim == 1 for A.ndim == 1, while for A.ndim > 1 we allow J.ndim > 1?

I've updated the text to be less restrictive here, such that the guidance does not preclude multi-dimensional `J`. However, behavior when `J` is mult-dimensional is not currently specified.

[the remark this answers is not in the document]
kgryte: I've added a note regarding unspecified behavior when `J.ndim > 1`.

spec/draft/API_specification/indexing.rst
```diff
@@ -173,6 +179,43 @@ Multi-dimensional arrays must extend the concept of single-axis indexing to mult
...
+  - A negative index ``j`` is related to a zero-based nonnegative index ``i`` via ``i = n+j``.
+
+  .. note::
+    This specification does not require bounds checking. The behavior for out-of-bounds integer indices is left unspecified.
+
+- Providing duplicate integer index elements in ``J`` must result in the duplication of the corresponding elements of ``A`` in the resulting array.
+
+- The result of providing a valid integer index for each element in ``J`` must be an array having shape ``S3 = (k,)``.
+
+An array must support indexing an array having more than one dimension by an indexing tuple which includes only integers and one-dimensional integer arrays according to the following rules. Let ``A`` be an ``N``-dimensional array with shape ``S1 = (s1, s2, ..., sN)`` and ``T`` be a tuple ``(t1, t2, ..., tN)`` having length ``N`` and containing only valid integer and one-dimensional integer array indices.
```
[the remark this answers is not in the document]
kgryte: I've reworked the copy to be more generalized and added a note concerning integer arrays having more than one dimension.

spec/draft/API_specification/indexing.rst
```diff
@@ -173,6 +179,43 @@ Multi-dimensional arrays must extend the concept of single-axis indexing to mult
...
+
+- The result of providing a valid integer index for each element in ``J`` must be an array having shape ``(k,)``.
+
+An array must support indexing an array having more than one dimension by an indexing tuple which includes only integers and one-dimensional integer arrays according to the following rules. Let ``A`` be an ``N``-dimensional array with shape ``S1 = (s1, s2, ..., sN)`` and ``T`` be a tuple ``(t1, t2, ..., tN)`` having length ``N`` and containing only valid integer and one-dimensional integer array indices.
+
+- Providing an integer tuple element ``tk`` having value ``k`` must be equivalent to providing a zero-dimensional integer array ``K`` containing ``k``.
+
+- If ``T`` consists of only integers and zero-dimensional integer arrays, the result must be equivalent to indexing multiple axes using integer indices (see :ref:`indexing-multi-axis`).
+
+- When ``T`` contains a one-dimensional integer array ``J`` having shape ``S2 = (m,)``, where ``m`` is greater than or equal to the number of elements in any other one-dimensional integer array in ``T``, each element of ``T`` must be broadcast to the same shape as ``J``.
```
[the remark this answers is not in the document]
kgryte: I've updated the copy to be more generalized.

kgryte:
Further discussion of this PR and issue https://github.com/data-apis/array-api/issues/669 is slated for the upcoming workgroup meeting on Thursday, 20 February.

ev-br:
> Dask seems to fail with a multi-dimensional integer array, but will work with one-dimensional integer arrays. @ev-br Can confirm.

This is https://github.com/data-apis/array-api/issues/669#issuecomment-2651546755

spec/draft/API_specification/indexing.rst
```diff
@@ -173,6 +179,54 @@ Multi-dimensional arrays must extend the concept of single-axis indexing to mult
 
     *Rationale: this is consistent with bounds-checking for single-axis indexing. An implementation may choose to set the axis (dimension) size of the result array to* ``0`` *, raise an exception, return junk values, or some other behavior depending on device requirements and performance considerations.*
 
+Integer Array Indexing
+----------------------
+
+An array must support indexing a one-dimensional array by an integer array (or an equivalent sequence) according to the following rules. Let ``A`` be a one-dimensional array with shape ``(n,)``, and let ``J`` be an integer array (or an equivalent sequence).
```
shoyer, line 185: Nit: I found it a little confusing that this section starts by describing special behavior for 1D ararys, which is entirely a special case of the behavior for ND arrays. It would be helpful to clarify this up front.

(The only way in which 1D arrays need to be special cased is that indexing like `a[b]` should be defined as equivalent to  `a[(b,)]`, but that is almost just a detail of Python syntax.)

spec/draft/API_specification/indexing.rst
```diff
@@ -173,6 +179,54 @@ Multi-dimensional arrays must extend the concept of single-axis indexing to mult
...
+
+- If ``J`` is a zero-dimensional array containing an integer index ``k``, the result must be equivalent to providing a single-axis integer index ``k`` (i.e., if ``J`` is a zero-dimensional array, ``A[J]`` must equal ``A[k]``; see :ref:`indexing-single-axis`).
+
+- If ``J`` is a one-dimensional array (or an equivalent sequence) with shape ``(k,)``, the result must be a one-dimensional array ``B``having shape ``(k,)``. Each element of ``B`` must follow the relation ``B[i] = A[J[i]]``.
+
+.. note::
+   If ``J`` is an array having more than one dimension, behavior is left unspecified and thus implementation-defined. This may be revisited in a future revision of this standard.
+
+An array must support indexing an array having more than one dimension by an indexing tuple which includes only integers and integer arrays (or equivalent sequences) according to the following rules. Let ``A`` be an ``N``-dimensional array with shape ``S1 = (s1, s2, ..., sN)`` and ``T`` be a tuple ``(I, J, ..., K)`` having length ``N`` and containing only valid integers and integer arrays (or equivalent sequences).
```
shoyer, line 210: > ``T`` be a tuple ``(I, J, ..., K)``

I find this very confusing, specially given the reuse of the symbols `J` and `K` below. Can you write this as ``(t0, t1, ..., tN)` instead?

spec/draft/API_specification/indexing.rst
```diff
@@ -173,6 +179,54 @@ Multi-dimensional arrays must extend the concept of single-axis indexing to mult
...
+
+- After broadcasting, ``T`` must be equivalent to a tuple ``U = (u1, u2, ..., uN)`` containing only integer arrays having shape ``S2``.
+
+- Let ``v_i`` be the tuple formed by the integer indices ``u1[i], u2[i], ..., uN[i]``. When applying the indexing tuple ``U`` to ``A``, the result must be an array having shape ``S2`` and contain each element ``A[v_i]``.
+
+.. note::
+   This specification does not currently address indexing tuples which combine slices and integer arrays. Behavior for such indexing tuples is left unspecified and thus implementation-defined. This may be revisited in a future revision of this standard.
+
+.. note::
+   If an indexing tuple contains an integer array having more than one dimension, behavior is left unspecified and thus implementation-defined. This may be revisited in a future revision of this standard.
```
shoyer, line 228: I think leaving out indexing by multi-dimensional arrays is a mistake. It is not hard to describe indexing by multi-dimensional arrays (they are broadcast first, and then the shared shape is used to define the result, very similar to the existing logic), and this is extremely useful and general indexing behavior. For example, without indexing by multi-dimensional arrays, Xarray will be unable to use the array API standard.

spec/draft/API_specification/indexing.rst
shoyer, line 210: To avoid special case logic, indexing of 0D arrays should also be valid, with the only valid indexer being an empty tuple `()`.

shoyer:
> * And further limits integer arrays to one-dimension, as Dask does not support integer arrays having more than one dimension.

I would really like to see this restriction relaxed, which significantly limits the utility of this feature. Dask could absolutely support this via existing functionality (e.g., broadcast_arrays and ravel before indexing, followed by reshape).

Not every indexing operation can be supported efficiently in the fully distributed case, but it is far better to support indexing with multi-dimensional arrays in the standard with compatibility code in dask, than to force users to do indexing with flattened 1D arrays, which can have severe negative performance implications and makes it even harder to write efficient distributed code.

rgommers:
> Dask could absolutely support this via existing functionality (e.g., broadcast_arrays and ravel before indexing, followed by reshape).

Is the only reason Dask doesn't have it that no one has done the work? You still have a Dask commit bit, right? Any sense of whether a PR there will meet resistance?

seberg (commented):
Feel free to ignore the "I feel it is a bit verbose", obviously.  Mostly saw the same things Stephan did, I think a few points should be highlighted more, because it is easy to miss the interesting bits if you know indexing well enough.

Notes/Questions:
* Do we need to spell out whether tuple _subclasses_ should be supported or are undefined?
* Should we note the fact that if the result is a copy or not depends on library specific rules?  (if a library supports view semantics.)
* I think it makes sense to prescribe an `IndexError` on non-integer arrays.  (Again, de-factor prescribed be early text probably.)

shoyer:
> > Dask could absolutely support this via existing functionality (e.g., broadcast_arrays and ravel before indexing, followed by reshape).
> 
> Is the only reason Dask doesn't have it that no one has done the work? You still have a Dask commit bit, right? Any sense of whether a PR there will meet resistance?

Efficient fully general indexing doesn't fit easily into Dask's data model, because Dask needs to understand the structure of the distributed computation without evaluating data. So if you index a dask array by another chunked dask array, there is no way to avoid the need to do all-to-all transfer between chunks, basically duplicating the data on every distributed worker.

That said, the same thing is true for `dask.array.take()`, which Dask does implement.

I do probably still have a commit bit for Dask, but it's been many years since I contributed anything to the project.

spec/draft/API_specification/indexing.rst
```diff
@@ -173,6 +179,54 @@ Multi-dimensional arrays must extend the concept of single-axis indexing to mult
 
     *Rationale: this is consistent with bounds-checking for single-axis indexing. An implementation may choose to set the axis (dimension) size of the result array to* ``0`` *, raise an exception, return junk values, or some other behavior depending on device requirements and performance considerations.*
 
+Integer Array Indexing
+----------------------
+
+An array must support indexing a one-dimensional array by an integer array (or an equivalent sequence) according to the following rules. Let ``A`` be a one-dimensional array with shape ``(n,)``, and let ``J`` be an integer array (or an equivalent sequence).
+
+.. note::
+   An **equivalent sequence** is defined as a sequence such as a list or tuple which, when provided as an argument to ``asarray``, results in an integer array.
```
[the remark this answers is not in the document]
ev-br: This part is a bit problematic, actually. First and foremost, tuples are ambiguous: `a[(1, 2)]` is exactly equivalent to `a[1, 2]`: it indexes a 2D array with two integers, not a 1D array with a sequence. So this only leaves a list as a possible fancy indexer, like numpy.

And then not every library allows lists:

```
In [7]: a = jnp.arange(5)

In [8]: a[[0, 1]]
...
TypeError: Using a non-tuple sequence for multidimensional indexing is not allowed; use `arr[array(seq)]` instead of `arr[seq]`. See https://github.com/jax-ml/jax/issues/4564 for more information.
```

(cross-ref https://github.com/data-apis/array-api/issues/864 which brought in the last example.)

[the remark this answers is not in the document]
shoyer: Agreed, list sequences are ambiguous. I would not support them, for the reasons discussed in the JAX issue referenced from that error message: https://github.com/jax-ml/jax/issues/4564

spec/draft/API_specification/indexing.rst
kgryte: Done!

spec/draft/API_specification/indexing.rst
```diff
@@ -173,6 +179,54 @@ Multi-dimensional arrays must extend the concept of single-axis indexing to mult
...
+
+- If ``T`` consists of only integers and zero-dimensional integer arrays, the result must be equivalent to indexing multiple axes using integer indices. For example, if ``A`` is a two-dimensional array, ``T`` is the tuple ``(i, J)``, ``i`` is a valid integer index, and ``J`` is a zero-dimensional array containing a valid integer index ``j``, the result of ``A[T]`` must be equivalent to ``A[(i,j)]`` (see :ref:`indexing-multi-axis`).
+
+- If ``T`` consists of only integers and integer arrays (or equivalent sequences), with at least one of those integer arrays being a one-dimensional integer array (or an equivalent sequence) ``J`` having shape ``S2 = (m,)``, where ``m`` is greater than or equal to the number of elements in any other integer array (or an equivalent sequence) in ``T``, each element of ``T`` must be broadcast to the same shape as ``J``.
+
+- An ``IndexError`` exception must be raised if any element of ``T`` is not broadcast-compatible with ``J`` (see :ref:`broadcasting`).
+
+- After broadcasting, ``T`` must be equivalent to a tuple ``U = (u1, u2, ..., uN)`` containing only integer arrays having shape ``S2``.
```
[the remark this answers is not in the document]
kgryte: I've removed the `IndexError` prescription in favor of just saying "an exception must be raised".

spec/draft/API_specification/indexing.rst
```diff
@@ -173,6 +179,54 @@ Multi-dimensional arrays must extend the concept of single-axis indexing to mult
...
+   An **equivalent sequence** is defined as a sequence such as a list or tuple which, when provided as an argument to ``asarray``, results in an integer array.
+
+- Each integer index element in ``J`` must satisfy the rules stated above for indexing a single-axis (see :ref:`indexing-single-axis`). Namely,
+
+  - Nonnegative indices must start at ``0`` (i.e., zero-based indexing).
+  - **Valid** nonnegative indices must reside on the half-open interval ``[0, n)``.
+  - Negative indices must count backward from the last index, starting from ``-1`` (i.e., negative-one-based indexing, where ``-1`` refers to the last index).
+  - **Valid** negative indices must reside on the closed interval ``[-n, -1]``.
+  - A negative index ``j`` is related to a zero-based nonnegative index ``i`` via ``i = n+j``.
+
```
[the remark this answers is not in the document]
kgryte: Removed.

spec/draft/API_specification/indexing.rst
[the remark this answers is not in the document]
kgryte: I've updated the copy to generalize beyond one-dimensional arrays.

spec/draft/API_specification/indexing.rst
kgryte: I've removed and reworked as part of generalization to ND arrays.

spec/draft/API_specification/indexing.rst
kgryte: Agreed. I think this falls out from "``T`` be a tuple ``(t0, t1, ..., tN)`` having length ``N``", where `N` is the dimensionality of `A`.

spec/draft/API_specification/indexing.rst
```diff
@@ -173,6 +179,43 @@ Multi-dimensional arrays must extend the concept of single-axis indexing to mult
...
+  .. note::
+    This specification does not require bounds checking. The behavior for out-of-bounds integer indices is left unspecified.
+
+- Providing duplicate integer index elements in ``J`` must result in the duplication of the corresponding elements of ``A`` in the resulting array.
+
+- The result of providing a valid integer index for each element in ``J`` must be an array having shape ``(k,)``.
+
+An array must support indexing an array having more than one dimension by an indexing tuple which includes only integers and one-dimensional integer arrays according to the following rules. Let ``A`` be an ``N``-dimensional array with shape ``S1 = (s1, s2, ..., sN)`` and ``T`` be a tuple ``(t1, t2, ..., tN)`` having length ``N`` and containing only valid integer and one-dimensional integer array indices.
```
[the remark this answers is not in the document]
kgryte: I've rephrased this. Thanks for flagging.

kgryte:
Thanks, @ev-br, @shoyer, and @seberg for the reviews. I've updated the proposed text based on your feedback with generalization to multi-dimensional integer arrays.

spec/draft/API_specification/indexing.rst
```diff
@@ -173,6 +179,34 @@ Multi-dimensional arrays must extend the concept of single-axis indexing to mult
 
     *Rationale: this is consistent with bounds-checking for single-axis indexing. An implementation may choose to set the axis (dimension) size of the result array to* ``0`` *, raise an exception, return junk values, or some other behavior depending on device requirements and performance considerations.*
 
+Integer Array Indexing
+----------------------
+
+An array must support indexing by an indexing tuple which contains only integers and integer arrays according to the following rules. Let ``A`` be an ``N``-dimensional array with shape ``S1``. Let ``T`` be a tuple ``(t1, t2, ..., tN)`` having length ``N``. Let ``tk`` be an individual element of ``T``.
```
shoyer, line 185: I wonder if it might be helpful to help with an informal description  followed by the details would make this section easier to read, e.g., "The array standard supported vectorized indexing by multi-dimensional integer arrays with broadcasting, as long as there is an indexing array for each dimension of the indexed array"

shoyer (approved):

spec/draft/API_specification/indexing.rst
kgryte: I've added a note as a preface to this section.

spec/draft/API_specification/indexing.rst
```diff
@@ -173,6 +179,37 @@ Multi-dimensional arrays must extend the concept of single-axis indexing to mult
...
+
+- Providing a zero-dimensional integer array ``tk`` containing an integer index must be equivalent to providing an integer index having the value ``tk[()]``. Conversely, each integer index ``tk`` must be equivalent to a zero-dimensional integer array containing the same value and be treated as such in all contexts, including shape inference and broadcasting. Accordingly, if ``T`` consists of only integers and zero-dimensional integer arrays, the result must be equivalent to indexing multiple axes using integer indices. For example, if ``A`` is a two-dimensional array, ``T`` is the tuple ``(i, J)``, ``i`` is a valid integer index, and ``J`` is a zero-dimensional array containing a valid integer index ``j``, the result of ``A[T]`` must be equivalent to ``A[(i,j)]`` (see :ref:`indexing-multi-axis`).
```
asmeurer, line 190: NumPy does distinguish `a[0]` and `a[asarray(0)]` in that the latter is considered a "fancy" index and creates a copy whereas `a[0]` creates a view. This isn't relevant for the standard, but maybe the "all contexts" should be worded slightly differently.

kgryte: I'll remove "in all contexts".

kgryte: Done.

kgryte:
We discussed this PR during the workgroup meeting on Feb 20, 2025, and arrived at the following consensus:

1. Dask should be able to support multi-dimensional array indexing. While random access of multiple elements may not be particularly efficient in a distributed context, there is no technical barrier altogether preventing its implementation.
2. While ndonnx does not currently support bracket integer array indexing, it does support integer array indexing via its `gatherND` API: <https://onnx.ai/onnx/operators/onnx__GatherND.html>. As such, it should be possible to implement.
3. While Dask does not currently support indexing with multiple non-zero-dimensional integer arrays, it was determined that, without being able to provide multiple integer arrays, integer array indexing would be too limited to be useful.
4. Similarly, being able to index with integer arrays having more than one dimension was considered desirable. In the compat layer, we should be able to add a helper function which can provide a workaround for Dask. Namely, flatten and then unravel.
5. While lists and other sequences can help with ergonomics, consensus was to only specify behavior for integer arrays, not array-like values.
6. `__setitem__` semantics are punted to v2025 and are to be left unspecified for the 2024 revision.
7. Integer arrays should have the default array index data type in order for array indexing operations to be portable across array libraries.

Based on the above, I've updated the text accordingly. The main updates were items 6-7. Otherwise, the text was given thumbs of approval during the workgroup meeting.

spec/draft/API_specification/indexing.rst
```diff
@@ -173,6 +179,45 @@ Multi-dimensional arrays must extend the concept of single-axis indexing to mult
...
 
+Integer Array Indexing
+----------------------
+
+.. note::
+  Integer array indexing, as described in this specification, is a reduced subset of "vectorized indexing" semantics, as implemented in libraries such as NumPy. In vectorized indexing, integers and integer arrays are broadcasted to integer arrays having a common shape before being "zipped" together to form a list of index coordinates. This form of indexing diverges from the multi-axis indexing semantics described above (see :ref:`indexing-multi-axis`) where each element of an indexing tuple comprised of integers and slices independently indexes a particular axis. This latter form of indexing is commonly referred to as "orthogonal indexing" and is the default form of indexing outside of Python in languages such as Julia and MATLAB.
+
+An array must support indexing by an indexing tuple which contains only integers and integer arrays according to the following rules. Let ``A`` be an ``N``-dimensional array with shape ``S1``. Let ``T`` be a tuple ``(t1, t2, ..., tN)`` having length ``N``. Let ``tk`` be an individual element of ``T``.
```
ev-br, line 188:
```suggestion
An array must support vectorized indexing by an indexing tuple which contains only integers and integer arrays according to the following rules. Let ``A`` be an ``N``-dimensional array with shape ``S1``. Let ``T`` be a tuple ``(t1, t2, ..., tN)`` having length ``N``. Let ``tk`` be an individual element of ``T``.
```

spec/draft/API_specification/indexing.rst
```diff
@@ -173,6 +179,45 @@ Multi-dimensional arrays must extend the concept of single-axis indexing to mult
...
+
+- Providing a zero-dimensional integer array ``tk`` containing an integer index must be equivalent to providing an integer index having the value ``tk[()]``. Conversely, each integer index ``tk`` must be equivalent to a zero-dimensional integer array containing the same value and be treated as such, including shape inference and broadcasting. Accordingly, if ``T`` consists of only integers and zero-dimensional integer arrays, the result must be equivalent to indexing multiple axes using integer indices. For example, if ``A`` is a two-dimensional array, ``T`` is the tuple ``(i, J)``, ``i`` is a valid integer index, and ``J`` is a zero-dimensional array containing a valid integer index ``j``, the result of ``A[T]`` must be equivalent to ``A[(i,j)]`` (see :ref:`indexing-multi-axis`).
```
ev-br, line 201:
```suggestion
- Providing a zero-dimensional integer array ``tk`` containing an integer index must be equivalent to providing an integer index having the value ``int(tk)``. Conversely, each integer index ``tk`` must be equivalent to a zero-dimensional integer array containing the same value and be treated as such, including shape inference and broadcasting. Accordingly, if ``T`` consists of only integers and zero-dimensional integer arrays, the result must be equivalent to indexing multiple axes using integer indices. For example, if ``A`` is a two-dimensional array, ``T`` is the tuple ``(i, J)``, ``i`` is a valid integer index, and ``J`` is a zero-dimensional array containing a valid integer index ``j``, the result of ``A[T]`` must be equivalent to ``A[(i,j)]`` (see :ref:`indexing-multi-axis`).
```

ev-br (approved):
LGTM modulo optional nits.

kgryte:
As we discussed this PR during the workgroup meeting and no further comments have been made, I'll go ahead and merge. Any further adjustments can be made as errata to the 2024 specification revision. Thanks all!

PR #905, opened 2025-02-26
fix: resolve missing `complex` type in applicable function signatures

kgryte:
This PR

- addresses an oversight when adding complex number support to the specification. Namely, the signatures for dunder methods were not updated to reflect `complex` scalar argument support. This oversight has already been addressed in the current draft (see https://github.com/data-apis/array-api/pull/862). This PR backports the changes to previous revisions of the specification.
- removes a stray Sphinx annotation for the `__or__` specification in the v2022.12 revision of the standard. It currently says that complex dtype support was added for this API, but this is not correct.
- adds a missing Sphinx annotation to `__eq__` as to when complex dtype support was added.


kgryte:
As this PR addresses a typing omission for already specified behavior per the v2022.12 revision of the Array API specification, I'll go ahead and merge.

PR #915, opened 2025-03-21
docs: fix `indexing` kwarg typing in `meshgrid`

crusaderky:


kgryte:
I believe this change may need to be backported to prior revisions of the standard.

src/array_api_stubs/_2021_12/creation_functions.py
```diff
@@ -311,15 +312,15 @@ def linspace(
...
+def meshgrid(*arrays: array, indexing: Literal["xy", "ij"] = "xy") -> List[array]:
     """
     Returns coordinate matrices from coordinate vectors.
 
     Parameters
     ----------
     arrays: array
         an arbitrary number of one-dimensional arrays representing grid coordinates. Each array should have the same numeric data type.
-    indexing: str
+    indexing: Literal["xy", "ij"]
```
lucascolley, line 323: stylistically, we could also opt for just `"xy" or "ij"`. I know that the sentiment of "keep typing syntax out of the type descriptions for readability" has sometimes been raised. On the other hand, https://numpydoc.readthedocs.io/en/latest/format.html#parameters says:

> For the parameter types, be as precise as possible.

so I am fine with either way.

lucascolley (approved):
thanks Guido!

Issue #918, opened 2025-03-31
Multi-device support meta-thread


crusaderky:
This is a tracker of the current state of support for more than one device at once in the Array API, its helper libraries, and the libraries that implement it.

Supporting multiple devices at the same time is typically substantially more fragile than pinning one of the available devices at interpreter level and then using that one exclusively, which typically works as intended.

### Array API
- Dictates that the device of the output arrays must always follow from that of the input(s), unless explicitly overridden by a `device=` kwarg, where allowed.
- Is frequently misinterpreted when it comes to priority of input arrays vs. global/context device: https://github.com/data-apis/array-api/pull/919
- There is controversy on what `__array_namespace_info__().default_device()` should return: https://github.com/data-apis/array-api/issues/835

### array-api-strict
- Supports three hardcoded devices, "cpu", "device1", "device2". This is fit for purpose for testing downstream bugs re. device propagation.

### array-api-tests
- Devices are untested: https://github.com/data-apis/array-api-tests/issues/302

### array-api-compat
- Adds device param to numpy 1, cupy, torch, and dask (read below).
- Implements helper functions `device()` and `to_device()` to work around non-compliance of wrapped libraries

### array-api-extra
- Full support and testing for non-default devices, using array-api-strict only. Actual support from real backends entirely depends on the below.

### NumPy
- It supports a single dummy device, "cpu".
- `array-api-compat` backports it to NumPy 1.x.

### CuPy
- Non-compliant support for multiple devices.
- array-api-compat adds a dummy `device=` parameter to functions.
- A compatibility layer is being added at the moment of writing by https://github.com/data-apis/array-api-compat/pull/293. [EDIT] it can't work, as array-api-compat can't patch methods.
- As it doesn't have a "cpu" device, it's impossible to test multi-device ops without access to a dual-GPU host.

### PyTorch
- Fully supported (with array-api-compat shims)
- However there's a bug that hampers testing on GPU CI: https://github.com/pytorch/pytorch/issues/150199

### JAX
- Bugs in `__array_namespace_info__`: https://github.com/jax-ml/jax/issues/27606
- Inside `jax.jit`, input-to-output device propagation works, but it's impossible to call creation functions (`empty`, `zeros`, `full`, etc.) on a non-default device: https://github.com/jax-ml/jax/issues/26000

### Dask
- Dask doesn't have a concept of device
[...]

PR #919, opened 2025-04-01
Clarify input arrays in device placement

crusaderky:
Clarify that input arrays should be considered for device placement.
This new line is a repetition of bullet point 2, three lines above. However, I am observing very high levels of ambiguity among the community in the interpretation of this paragraph, where I've had multiple people insist that the device set in the context manager or global default device should trump the device of input arrays.

### DIFF
--- a/spec/draft/design_topics/device_support.rst
+++ b/spec/draft/design_topics/device_support.rst
@@ -82,9 +82,11 @@ rather than hard requirements:
 
   1. If ``device=`` keyword is specified, that always takes precedence
 
-  2. If ``device=None``, then use the setting from a context manager, if set.
+  2. If ``device=None``, then use the device of the input array(s), if any.
 
-  3. If no context manager was used, then use the global default device/strategy
+  3. If there are no input arrays, then use the setting from a context manager, if set.
+
+  4. If no context manager was used, then use the global default device/strategy
 
 .. _device-out-of-scope:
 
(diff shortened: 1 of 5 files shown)


rgommers:
This change seems reasonable to me, it's more of a clarification than anything else. This numbered list was written with creation functions in mind I believe. The `device=` keyword descriptions already say something like "if `device` is `None`, use the device from the input array" in the relevant cases.

Before merging this, let's ensure the discussions in https://github.com/data-apis/array-api-compat/pull/293 (and perhaps other places) have been settled to ensure it's all consistent.

crusaderky:
XREF #918 
@kgryte ready for review and merge

spec/draft/design_topics/device_support.rst
```diff
@@ -77,14 +77,14 @@ rather than hard requirements:
 - Respect explicit device assignment (i.e. if the input to the ``device=`` keyword is not ``None``, guarantee that the array is created on the given device, and raise an exception otherwise).
 - Preserve device assignment as much as possible (e.g. output arrays from a function are expected to be on the same device as input arrays to the function).
 - Raise an exception if an operation involves arrays on different devices (i.e. avoid implicit data transfer between devices).
+- When a function accepts a mix of arrays and Python scalars, the scalars should inherit the device of the arrays, much like it happens with :ref:`type-promotion`.
```
lucyleeow, line 80:
```suggestion
- When a function accepts a mix of arrays and Python scalars, the scalars should inherit the device of the arrays, much like what happens with :ref:`type-promotion`.
```
nit

PR #920, opened 2025-04-03
`__setitem__` type promotion rules for array values

### LINKED ISSUE #916: Clarification on `__setitem__` and in-place ops type promotion rules
`__setitem__` states:
https://data-apis.org/array-api/2024.12/API_specification/generated/array_api.array.__setitem__.html#array_api.array.__setitem__

> Setting array values must not affect the data type of self.
> When value is a Python scalar (i.e., int, float, complex, bool), behavior must follow specification guidance on mixing arrays with Python scalars (see [Type Promotion Rules](https://data-apis.org/array-api/2024.12/API_specification/type_promotion.html#type-promotion)).

The above unambiguously disallows setting a python float into an array of integers, etc.
However, these use cases are left to interpretation:

- setting a float array value into an integer array, e.g.
```python
>>> a = xp.asarray([1], dtype=xp.int64)
>>> a[0] = xp.asarray(1.0)
```
this, by logic of simmetry with pure python floats above, should be disallowed.

- setting a signed array value into an unsigned array, e.g.
```python
>>> a = xp.asarray([1], dtype=xp.unt64)
>>> a[0] = xp.asarray(-2, dtype=xp.int8)
```
 
- setting an array value with a larger dtype into a smaller dtype of the same kind, e.g. 
```python
>>> a = xp.asarray([1], dtype=xp.int8)
>>> a[0] = xp.asarray(2, dtype=xp.int64)
```

All these use cases are quietly allowed by numpy.
_Sanity_ would dictate for them to be all disallowed, exactly like in binops.

If the last use case is allowed, it also opens the issue of in-place binops (`__iadd__`) etc.
Whereas out-of-place binops unambiguously must cast the smaller dtype to the larger one and then perform the operation, what is the correct process for in-place ones, when lhs has smaller dtype than rhs?
```python
lhs[idx] = op(lhs[idx], xp.astype(rhs[idx], lhs.dtype))
```
or
```python
t = xp.result_type(lhs, rhs)
lhs[idx] = xp.astype(op(xp.astype(lhs[idx], t), xp.astype(rhs[idx], t), lhs.dtype)
``` 
? the output will subtly be different.

crusaderky:
Closes #916

### DIFF
--- a/src/array_api_stubs/_draft/array_object.py
+++ b/src/array_api_stubs/_draft/array_object.py
@@ -1134,8 +1134,10 @@ class _array:
                Indexing semantics when ``key`` is an integer array or a tuple of integers and integer arrays is currently unspecified and thus implementation-defined. This will be revisited in a future revision of this standard.
 
         -   Setting array values must not affect the data type of ``self``.
-        -   When ``value`` is a Python scalar (i.e., ``int``, ``float``, ``complex``, ``bool``), behavior must follow specification guidance on mixing arrays with Python scalars (see :ref:`type-promotion`).
-        -   When ``value`` is an ``array`` of a different data type than ``self``, how values are cast to the data type of ``self`` is implementation defined.
+        -   ``value`` must be  promoted to ``self.dtype`` according to :ref:`type-promotion`; if this is not possible, behavior is undefined.
+
+        .. versionchanged:: 2025.12
+            Specified :ref:`type-promotion` when ``value`` is an array.
         """
 
     def __sub__(self: array, other: Union[int, float, complex, array], /) -> array:


src/array_api_stubs/_draft/array_object.py
```diff
@@ -1134,8 +1134,10 @@ def __setitem__(
                Indexing semantics when ``key`` is an integer array or a tuple of integers and integer arrays is currently unspecified and thus implementation-defined. This will be revisited in a future revision of this standard.
 
         -   Setting array values must not affect the data type of ``self``.
-        -   When ``value`` is a Python scalar (i.e., ``int``, ``float``, ``complex``, ``bool``), behavior must follow specification guidance on mixing arrays with Python scalars (see :ref:`type-promotion`).
-        -   When ``value`` is an ``array`` of a different data type than ``self``, how values are cast to the data type of ``self`` is implementation defined.
+        -   ``value`` must be  promoted to the data type of ``self`` according to :ref:`type-promotion`. If this is not possible, behavior is unspecified and thus implementation-defined.
```
kgryte, line 1137:
```suggestion
        -   ``value`` must be  promoted to the data type of ``self`` according to :ref:`type-promotion`. If this is not supported according to :ref:`type-promotion`, behavior is unspecified and thus implementation-defined.
```

kgryte (approved):
LGTM. Thanks, @crusaderky!

PR #924, opened 2025-04-09
fix: update type signature of `eigh`

### LINKED ISSUE #922: Wrong return type for `linalg.eigh` in function signature
`eigh` returns a `Tuple[array]` according to the function signature -

https://github.com/data-apis/array-api/blob/0941b216f6231195f1f6a332f957e45554f5da79/src/array_api_stubs/_2024_12/linalg.py#L166

but it actually returns a `NamedTuple` (hence, `Tuple[array, array]`) -

https://github.com/data-apis/array-api/blob/0941b216f6231195f1f6a332f957e45554f5da79/src/array_api_stubs/_2024_12/linalg.py#L198-L204

which causes problems with type checkers.

Is this intentional or is it a bug? I'd be happy to submit a quick patch to fix the function signature.

Saransh-cpp:
Fixes #922 

The type signature of `eigh` has been updated from `Tuple[array]` to `Tuple[array, array]`.

I am not sure how things are backported in this repository, but for now I have updated the signature in all the versions. Please let me know if this should be changed!

### DIFF
--- a/src/array_api_stubs/_2021_12/linalg.py
+++ b/src/array_api_stubs/_2021_12/linalg.py
@@ -86,7 +86,7 @@ def diagonal(x: array, /, *, offset: int = 0) -> array:
     """
 
 
-def eigh(x: array, /) -> Tuple[array]:
+def eigh(x: array, /) -> Tuple[array, array]:
     """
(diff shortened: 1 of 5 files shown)


lucascolley (approved):
thanks @Saransh-cpp, LGTM!

kgryte (approved):
LGTM.

PR #926, opened 2025-04-14
docs: `clip()` with scalar min/max clarifications

### LINKED ISSUE #925: `clip`: ambiguous wording for scalar min/max
[clip](https://data-apis.org/array-api/latest/API_specification/generated/array_api.clip.html) states:

> - min [...] Should have the same data type as x. Default: None.
> - max [...] Should have the same data type as x. Default: None.
> - For scalar min and/or max, the scalar values should follow type promotion rules for operations involving arrays and scalar operands (see [Type Promotion Rules](https://data-apis.org/array-api/latest/API_specification/type_promotion.html#type-promotion)). Hence, if x and either min or max have different data type kinds (e.g., integer versus floating-point), behavior is unspecified and thus implementation-dependent.

The third point could use rewording in my opinion.
I believe that the intended meaning was that `clip(Array[int64], 1.0)` has unspecified behaviour, as per type promotion rules.
However, the wording as it is could be interpreted that `clip(Array[float64], 1)` also has unspecified behaviour, which doesn't make sense to me.

# Proposal
Change
> Hence, if `x` and either `min` or `max` have different data type kinds (e.g., integer versus floating-point), behavior is unspecified and thus implementation-dependent.

To
> Hence, if the scalar type of either `min` or `max` is incompatible with the data type of `x` (`x` has integer data type and either `min` or `max` are scalar floats), behavior is unspecified and thus implementation-dependent.

# Backporting

As I believe this is only a clarification, it should be backported to previous versions. Note that, unlike most binops, this function has always accepted scalars for min/max and wording was the same in 2023.12.

# XREFs
-  https://github.com/data-apis/array-api-tests/pull/360

crusaderky:
Closes #925

### DIFF
--- a/src/array_api_stubs/_2023_12/elementwise_functions.py
+++ b/src/array_api_stubs/_2023_12/elementwise_functions.py
@@ -807,6 +807,7 @@ def clip(
     - If a broadcasted element in ``min`` is greater than a corresponding broadcasted element in ``max``, behavior is unspecified and thus implementation-dependent.
     - If ``x`` has an integral data type and a broadcasted element in ``min`` or ``max`` is outside the bounds of the data type of ``x``, behavior is unspecified and thus implementation-dependent.
     - If ``x`` and either ``min`` or ``max`` have different data type kinds (e.g., integer versus floating-point), behavior is unspecified and thus implementation-dependent.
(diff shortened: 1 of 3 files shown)


kgryte (approved):
LGTM

Issue #854, opened 2024-11-21
RFC: add `isin` for elementwise set inclusion test


lucascolley:
## Prior art
- NumPy - https://numpy.org/doc/stable/reference/generated/numpy.isin.html
- CuPy - https://docs.cupy.dev/en/latest/reference/generated/cupy.isin.html
- Dask - https://docs.dask.org/en/stable/generated/dask.array.isin.html
- JAX - https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.isin.html
- PyTorch - https://pytorch.org/docs/stable/generated/torch.isin.html
- ndonnx - https://ndonnx.readthedocs.io/en/latest/api/ndonnx.additional.html#ndonnx.additional.isin
- MLX - `isin` is not present yet

## Motivation

This function is used in scikit-learn. They've implemented it in terms of the standard, and that implementation could find a home in array-api-extra: https://github.com/data-apis/array-api-extra/issues/34. @asmeurer suggested there that we should also consider adding this to the standard.

asmeurer:
Another potential reason for adding it is that it uses a nontrivial implementation which depends on some heuristics based on the input size.

rgommers:
Thanks @lucascolley. I've added `ndonnx` (which has it) and MLX (which doesn't) to the issue description.

This seems like a very reasonable proposal to me. Implementing `isin` in terms of other primitives in the standard is a little complex indeed.

The return type should always be a boolean array. The NumPy docs say it can be a `bool` for a single input element, but that's actually a bug in the docs. I checked NumPy, JAX, and PyTorch and all return a 0-D array.

rgommers:
The thing to discuss here is what keywords are desired I think. NumPy and Dask use:
```python
def isin(element, test_elements, assume_unique=False, invert=False, *, kind=None)
```

The private scikit-learn implementation [here](https://github.com/scikit-learn/scikit-learn/blob/439ea045ad44e6a09115dc23e9bf23db00ff41de/sklearn/utils/_array_api.py#L931-L982) is:
```python
def isin(element, test_elements, xp, assume_unique=False, invert=False)
```

JAX:
```python
def isin(element, test_elements, assume_unique=False, invert=False, *, method='auto')
```

PyTorch:
```python
def isin(elements, test_elements, *, assume_unique=False, invert=False)
```

ndonnx:
```python
def isin(x: Array, items: Sequence[Scalar]) -> Array
```

[...]

PR #959, opened 2025-06-12
feat: add `isin` to the specification

kgryte:
This PR:

- resolves https://github.com/data-apis/array-api/issues/854 by adding `isin` to the specification.
- of the keyword arguments determined according to array comparison [data](https://github.com/data-apis/array-api-comparison/blob/8cd85d8bcafdd71b4f18df3e7bd9fccd1507d1ee/signatures/sets/isin.md), this PR chooses to support only the `invert` kwarg. The `assume_unique` kwarg was not included for the following reasons:

    1. not all array libraries support this kwarg (e.g., ndonnx and CuPy). CuPy lists the kwarg in its documentation but states that this kwarg is ignored.
    2. when doing a quick [search](https://github.com/search?q=repo%3Ascikit-learn%2Fscikit-learn+np.isin&type=code&p=1) through sklearn, I was only able to find one usage of `assume_unique` when using `isin` and that was when searching lists of already known unique values.
    3. `assume_unique` is something of a performance optimization/implementation detail which we have generally attempted to avoid when standardizing APIs.

- does not place restrictions on the shape of `x2`. While some libraries may choose to flatten a multi-dimensional `x2`, that is something of an implementation detail and not strictly necessary. For example, an implementation could defer to an "includes" kernel which performs nested loop iteration without needing to perform explicit reshapes/copies.

- adds support for scalar arguments for either `x1` or `x2`. This follows recent general practice in standardized APIs, with the restriction that at least one of `x1` or `x2` must be an array.

- specifies that value equality **should** be used, but not **must** be used. This follows other set APIs (e.g., `unique*`). As a consequence of value equality, `NaN` values can never test as `True` and there is no distinction between signed zeros.

- allows both `x1` and `x2` to be of any data type. However, if `x1` and `x2` have no promotable data type, behavior is left unspecified and thus implementation-defined.

## Questions

- Would we be okay with requiring that value equality **must** be used? Is there a scenario where we want to allow libraries some wiggle room, such as with `NaN` and signed zero comparison?
- Are we okay with leaving out `assume_unique`?
- Are we okay with not mandating reshape behavior if `x2` is multi-dimensional?

### DIFF
--- a/src/array_api_stubs/_draft/set_functions.py
+++ b/src/array_api_stubs/_draft/set_functions.py
@@ -1,7 +1,51 @@
-__all__ = ["unique_all", "unique_counts", "unique_inverse", "unique_values"]
+__all__ = ["isin", "unique_all", "unique_counts", "unique_inverse", "unique_values"]
 
 
 from ._types import Tuple, array
+
+
+def isin(
+    x1: Union[array, int, float, complex, bool],
+    x2: Union[array, int, float, complex, bool],
+    /,
+    *,
+    invert: bool = False,
+) -> array:
+    """
+    Tests whether each element in ``x1`` is in ``x2``.
+
+    Parameters
+    ----------
+    x1: Union[array, int, float, complex, bool]
+        first input array. **May** have any data type.
+    x2: Union[array, int, float, complex, bool]
+        second input array. **May** have any data type.
+    invert: bool
+        boolean indicating whether to invert the test criterion. If ``True``, the function **must** test whether each element in ``x1`` is *not* in ``x2``. If ``False``, the function **must** test whether each element in ``x1`` is in ``x2``. Default: ``False``.
+
+    Returns
+    -------
+    out: array
+        an array containing element-wise test results. The returned array **must** have the same shape as ``x1`` and **must** have a boolean data type.
+
+    Notes
+    -----
+
+    -   At least one of ``x1`` or ``x2`` **must** be an array.
+
+    -   If an element in ``x1`` is in ``x1``, the corresponding element in the output array **must** be ``True``; otherwise, the corresponding element in the output array **must** be ``False``.
+
+    -   Testing whether an element in ``x1`` corresponds to an element in ``x2`` **should** be determined based on value equality (see :func:`~array_api.equal`). For input arrays having floating-point data types, value-based equality implies the following behavior.
+
+        -   As ``nan`` values compare as ``False``, if an element in ``x1`` is ``nan`` and ``invert`` is ``False``, the corresponding element in the returned array **should** be ``False``. Otherwise, if an element in ``x1`` is ``nan`` and ``invert`` is ``True``, the corresponding element in the returned array **should** be ``True``.
+        -   As complex floating-point values having at least one ``nan`` component compare as ``False``, if an element in ``x1`` is a complex floating-point value having one or more ``nan`` components and ``invert`` is ``False``, the corresponding element in the returned array **should** be ``False``. Otherwise, if an element in ``x1`` is a complex floating-point value having one or more ``nan`` components and ``invert`` is ``True``, the corresponding element in the returned array **should** be ``True``.
+        -   As ``-0`` and ``+0`` compare as ``True``, if an element in ``x1`` is ``±0`` and ``x2`` contains at least one element which is ``±0``
+
+            -   if ``invert`` is ``False``, the corresponding element in the returned array **should** be ``True``.
+            -   if ``invert`` is ``True``, the corresponding element in the returned array **should** be ``False``.
+
+    -   Comparison of arrays without a corresponding promotable data type (see :ref:`type-promotion`) is unspecified and thus implementation-defined.
+    """
 
 
 def unique_all(x: array, /) -> Tuple[array, array, array, array]:
--- a/spec/draft/API_specification/set_functions.rst
+++ b/spec/draft/API_specification/set_functions.rst
@@ -18,6 +18,7 @@ Objects in API
    :toctree: generated
    :template: method.rst
 
+   isin
    unique_all
    unique_counts
    unique_inverse


src/array_api_stubs/_draft/set_functions.py
```diff
@@ -1,7 +1,51 @@
...
+        -   As complex floating-point values having at least one ``nan`` component compare as ``False``, if an element in ``x1`` is a complex floating-point value having one or more ``nan`` components and ``invert`` is ``False``, the corresponding element in the returned array **should** be ``False``. Otherwise, if an element in ``x1`` is a complex floating-point value having one or more ``nan`` components and ``invert`` is ``True``, the corresponding element in the returned array **should** be ``True``.
+        -   As ``-0`` and ``+0`` compare as ``True``, if an element in ``x1`` is ``±0`` and ``x2`` contains at least one element which is ``±0``
+
+            -   if ``invert`` is ``False``, the corresponding element in the returned array **should** be ``True``.
+            -   if ``invert`` is ``True``, the corresponding element in the returned array **should** be ``False``.
```
rgommers, line 45: Same comment here about not duplicating with `invert=True`

rgommers (commented):
Thanks @kgryte. Looks pretty good to me. I agree with the design choices in the PR description.

> Would we be okay with requiring that value equality must be used? Is there a scenario where we want to allow libraries some wiggle room, such as with NaN and signed zero comparison?

I am not sure wiggle room is needed here. This function has more to do with `equal` than with `unique` I think. I just checked NumPy, PyTorch, JAX and CuPy - all seem to be using value equality for `nan`.

> Are we okay with leaving out `assume_unique`?

Yes. 

> Are we okay with not mandating reshape behavior if `x2` is multi-dimensional?

I think that that part of the `np.isin` docstring is confusing. Reshaping is meaningless, the only point of that is trying to express that the comparisons are element-wise. It'd be better to have a simple double for-loop with pseudo-code. There is no broadcasting either, any shapes should work and the output has the same shape as `x1`.

src/array_api_stubs/_draft/set_functions.py
```diff
@@ -1,7 +1,51 @@
...
+    *,
+    invert: bool = False,
+) -> array:
+    """
+    Tests whether each element in ``x1`` is in ``x2``.
+
+    Parameters
+    ----------
+    x1: Union[array, int, float, complex, bool]
+        first input array. **May** have any data type.
```
ev-br, line 20: Just to double-check, are we happy with e.g. torch not allowing complex values here:

```
In [16]: torch.isin(1j, torch.arange(3, dtype=torch.float64))
RuntimeError                              Traceback (most recent call last)
Cell In[16], line 1
----> 1 torch.isin(1j, torch.arange(3, dtype=torch.float64))

RuntimeError: Unsupported input type encountered for isin(): ComplexDouble
```

src/array_api_stubs/_draft/set_functions.py
```diff
@@ -1,7 +1,51 @@
...
+    x2: Union[array, int, float, complex, bool]
+        second input array. **May** have any data type.
+    invert: bool
+        boolean indicating whether to invert the test criterion. If ``True``, the function **must** test whether each element in ``x1`` is *not* in ``x2``. If ``False``, the function **must** test whether each element in ``x1`` is in ``x2``. Default: ``False``.
```
ev-br, line 24: So if `isin(x1, x2, invert=True)` is exactly equivalent to `logical_not(isin(x1, x2))`, we could drop the argument completely.

src/array_api_stubs/_draft/set_functions.py
```diff
@@ -1,7 +1,51 @@
...
+
+    -   Testing whether an element in ``x1`` corresponds to an element in ``x2`` **should** be determined based on value equality (see :func:`~array_api.equal`). For input arrays having floating-point data types, value-based equality implies the following behavior.
+
+        -   As ``nan`` values compare as ``False``, if an element in ``x1`` is ``nan`` and ``invert`` is ``False``, the corresponding element in the returned array **should** be ``False``. Otherwise, if an element in ``x1`` is ``nan`` and ``invert`` is ``True``, the corresponding element in the returned array **should** be ``True``.
+        -   As complex floating-point values having at least one ``nan`` component compare as ``False``, if an element in ``x1`` is a complex floating-point value having one or more ``nan`` components and ``invert`` is ``False``, the corresponding element in the returned array **should** be ``False``. Otherwise, if an element in ``x1`` is a complex floating-point value having one or more ``nan`` components and ``invert`` is ``True``, the corresponding element in the returned array **should** be ``True``.
```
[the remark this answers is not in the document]
kgryte: I assume you are arguing that semantically the behavior should be equivalent to `xp.logical_not(xp.isin(x1, x2))`. Implementation-wise, I don't think this is the case. For example, NumPy [branches](https://github.com/numpy/numpy/blob/v2.3.0/numpy/lib/_arraysetops_impl.py#L1020) base on the value of `invert`.

I can, however, rephrase to state the semantic equivalence.

src/array_api_stubs/_draft/set_functions.py
kgryte: This should now be resolved.

src/array_api_stubs/_draft/set_functions.py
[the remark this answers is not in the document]
kgryte: Updated.

src/array_api_stubs/_draft/set_functions.py
kgryte: I recognize that the performance optimization argument I just stated is a bit tenuous here, given that the OP explicitly advocates against including `assume_unique`, but the difference between `invert` and `assume_unique` is that the former conveys semantic meaning, whereas the latter is primarily about making implementer lives easier.

src/array_api_stubs/_draft/set_functions.py
ev-br: It's doable of course.  That said, we are then adding to the growing "thickness" of the allegedly thin compat shim that is array-api-compat, and we'd be adding one more thing that realistically pytorch is not going to implement in a foreseeable future. Meaning we should be very clear that these small steps all move the compat layer from a temporary solution to permanent. Which looks like we should just stop pretending that the compat layer is temporary.

src/array_api_stubs/_draft/set_functions.py
```diff
@@ -1,7 +1,50 @@
...
+
+def isin(
+    x1: Union[array, int, float, complex, bool],
+    x2: Union[array, int, float, complex, bool],
+    /,
+    *,
+    invert: bool = False,
+) -> array:
+    """
+    Tests whether each element in ``x1`` is in ``x2``.
```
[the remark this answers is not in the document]
kgryte: You're right that the current description is ambiguous whether the function performs an element-wise operation or a reduction. Will update.

src/array_api_stubs/_draft/set_functions.py
```diff
@@ -1,7 +1,50 @@
...
+__all__ = ["isin", "unique_all", "unique_counts", "unique_inverse", "unique_values"]
 
 
-from ._types import Tuple, array
+from ._types import Tuple, Union, array
+
+
+def isin(
+    x1: Union[array, int, float, complex, bool],
+    x2: Union[array, int, float, complex, bool],
```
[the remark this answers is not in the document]
cbourjau: We indeed fall back to broadcasting in our implementation of searchsorted (see [here](https://github.com/Quantco/ndonnx/blob/dccfc0f5546e7cf77106db6f72b0520ef71a534f/ndonnx/_typed_array/onnx.py#L851-L857)).

src/array_api_stubs/_draft/set_functions.py
```diff
@@ -1,7 +1,50 @@
...
+    invert: bool = False,
+) -> array:
+    """
+    Tests whether each element in ``x1`` is in ``x2``.
+
+    Parameters
+    ----------
+    x1: Union[array, int, float, complex, bool]
+        first input array. **May** have any data type.
+    x2: Union[array, int, float, complex, bool]
```
cbourjau: Should there maybe be a constraint on the rank of `x2`?

kgryte: We intentionally did not impose a constraint and it is not clear whether there is a conceptual reason to do so, as this API is a vectorized API for finding whether a needle (a value) is in a haystack (an array) regardless of the dimensionality of the haystack.

ev-br (approved):
LGTM modulo an optional nit.

A preliminary test does not surface any problems, https://github.com/data-apis/array-api-tests/pull/407. The test does not perform value tests though, so it won't catch if a library does something other than equality testing. As long as only integer dtypes are allowed, there's not that many options though.

Issue #812, opened 2024-06-05
RFC: add bincount


ogrisel:
It seems quite widely adopted:

- https://numpy.org/doc/stable/reference/generated/numpy.bincount.html
- https://pytorch.org/docs/stable/generated/torch.bincount.html
- https://docs.cupy.dev/en/stable/reference/generated/cupy.bincount.html
- https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.bincount.html
- https://docs.dask.org/en/stable/generated/dask.array.bincount.html

Although I have not checked how compatible they are.

JAX in particular requires an extra `length` argument to statically define the length of the output array to make JIT possible and dask has a return value with [a dynamically sized shape](https://data-apis.org/array-api/latest/design_topics/data_dependent_output_shapes.html) represented by `(nan,)`.

rgommers:
Thanks for the suggestion @ogrisel!

`bincount` is indeed a heavily used function. I did a quick tally: scikit-learn currently has 117 usages of `np.bincount`, and SciPy 38. It's also used as a primitive for `histogram` & co, and it's not easy to implement efficiently with other existing functions in the standard. So from that perspective, I think it meets the bar for inclusion fairly easily.

Signatures do all seem compatible. There is at least one difference in behavior I spotted: the accepted input is non-negative integer values, but the case of negative values is not treated the same - it may raise (NumPy), or clip (JAX).

The NumPy issue tracker contains a lot of [open issues about `bincount`](https://github.com/numpy/numpy/issues?q=is%3Aissue+is%3Aopen+bincount), so the constraints on values, weights etc. seem to need careful specification.

The PyTorch docs also contain a warning about non-deterministic gradients - probably related to the data-dependent behavior:

> JAX in particular requires an extra `length` argument to statically define the length of the output array to make JIT possible and dask uses `(nan,)` shape to represent a [data-dependent output shapes](https://data-apis.org/array-api/latest/design_topics/data_dependent_output_shapes.html).

Yes, that's non-ideal. CuPy also adds a warning about `bincount` possibly causing synchronization.

The number of APIs that use data-dependent shapes is still quite low. As of now, I think this is the full list:
- boolean indexing
- the four `unique_*` functions
- `repeats`
- `nonzero`

NeilGirdhar:
[...]

PR #960, opened 2025-06-12
feat: add `bincount` to the specification

kgryte:
This PR

- resolves https://github.com/data-apis/array-api/issues/812 by adding `bincount` to the specification for counting the number of occurrences of each element in an input integer array.
- based on [comparison data](https://github.com/data-apis/array-api-comparison/blob/312c75bb5589a1c5fe8c6de6f8b8a196dc5f1e71/signatures/statistics/bincount.md), only supports a `minlength` keyword argument.
- allows `weights` to be both a positional and keyword argument.
- allows `weights` to have any numeric data type, including complex.
- specifies that, when `weights` is not provided, the output data type must be an integer data type. The data type rules follow other statistical functions (e.g., `sum`), where a minimum precision is required.
- specifies that, when `weights` is provided, the output data type must have the same data type as `weights`.
- specifies that an input array `x` **should** (not **must**) be a one-dimensional array. TensorFlow allows `x` to be multi-dimensional, and this PR chooses to provide wiggle room to allow supporting multi-dimensional arrays. One reason this isn't commonly supported in other libraries is that `bincount` has a data-dependent output shape; however, TF supports kwargs which allow specifying a static output shape, thus allowing `bincount` to generalize to multiple dimensions.
- includes an admonition that `bincount` has a data-dependent output shape and thus certain libraries are allowed to omit this function if too difficult to implement. This follows similar practice for other APIs having data-dependent output shapes (e.g., `unique*`).
- specifies that, if `x` contains negative values, behavior is unspecified and thus implementation-defined. NumPy raises an exception, while JAX clips.
- specifies that `weights` have the same shape as `x`; although, IMO, this restriction is not necessary and could be relaxed to broadcast-compatibility.
- specifies that the returned array **should** (not **must**) have shape `(N,)`, where `N = max(xp.max(x)+1, minlength)`. The use of **should** is intentional, in order to allow libraries such as JAX and TF to support other keyword arguments which may constrain the output shape.
- specifies that the default value of `minlength` must be `0`. According to docs, both CuPy and TF use a default of `None`.

## Questions

- Are we okay with allowing `weights` to be complex? This does not appear to be supported in NumPy (ref: https://github.com/numpy/numpy/issues/23313 and https://github.com/numpy/numpy/issues/16903); however, there isn't a technical reason why weights cannot be complex, as summation is well-defined for complex numbers.
- Similar to `sum` and other statistical functions, should we support an output `dtype` kwarg (e.g., in order to support overriding the default integer output type behavior)?
- Are we okay with `weights` being both positional and a keyword argument?

### DIFF
--- a/spec/draft/API_specification/statistical_functions.rst
+++ b/spec/draft/API_specification/statistical_functions.rst
@@ -18,6 +18,7 @@ Objects in API
    :toctree: generated
    :template: method.rst
 
+   bincount
    cumulative_prod
    cumulative_sum
    max
--- a/src/array_api_stubs/_draft/statistical_functions.py
+++ b/src/array_api_stubs/_draft/statistical_functions.py
@@ -1,4 +1,5 @@
 __all__ = [
+    "bincount",
     "cumulative_sum",
     "cumulative_prod",
     "max",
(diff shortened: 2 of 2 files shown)


NeilGirdhar:
> The shape of the output array for this function depends on the data values in ``x``; hence, array libraries which build computation graphs (e.g., JAX, Dask, etc.) can find this function difficult to implement without knowing the values in ``x``. Accordingly, such libraries **may** choose to omit this function. See :ref:`data-dependent-output-shapes` section for more details.

Why not allow Jax to implement this function by adding an optional `length` argument?  Most algorithms are amenable to that.

kgryte:
> I'd vote for keyword-only, since it's a very descriptive name and there's no real reason to use positional-only as far as I can tell.

@rgommers I am fine making the change to kwarg-only for guaranteed portability. sklearn includes both positional and kwarg usage, with the latter being more predominant. Similarly, from a search on sourcegraph, kwarg usage is more common, although positional usage of `np.bincount` is not uncommon.

jakevdp:
> @jakevdp Would be good to update the docstring then for `bincount`

Thanks for pointing that out – I'll get it updated.

betatim:
Slight preference for weights as keyword only. It is easy enough to update in scikit-learn.

As a user it is annoying/tedious if different libraries require different treatment. The whole point of array API is to have something uniform instead of maintaining a big bunch of `if` statements. From that point of view it would be nice to have something that works for jax as well. This would mean making `length` a argument mentioned in the standard. Is it possible to make a generic recommendation for what to pass as value. At least my first reaction to "you have to provide `length` was "how would I know what it should be, can't you work it out for me far better than I can?" But maybe `max(a)` and `len(a)` cover the vast majority of cases for naive users/get people started and then they can ponder if there is a better value? Because if it is that easy and it would remove the need to special case libraries like jax ... that might be a tradeoff worth making? Or am I missing something?

Issue #935, opened 2025-04-29
Add non-symmetric eigenvalue solvers, `eig` / `eivals`, to the linalg extension


ev-br:
Currently the linalg extension only contains symmetric/hermitian eigenvalue routines, `eigh` and `eigvalsh`. The general/non-symmetric versions are missing. 
IIUC one reason for the omission is that the low-level routines were not universally available on CUDA or were of allegedly questionable quality [1].
Now that the needed LAPACK-like routine was added to cuSolver 12.6 update 2 [2], the low-level blocker is no more.

Availability:
----------------

- *numpy*

`np.linalg.eig`, `np.linalg.eigvals` are available: https://numpy.org/doc/2.1/reference/generated/numpy.linalg.eig.html
 
 - *torch*

  `torch.linalg.eig`, `torch.linalg.eigvals` are available:  https://pytorch.org/docs/stable/generated/torch.linalg.eig.html

 - *jax*

 `jax.numpy.linalg.eig`, `jax.numpy.linalg.eigvals` are available: https://docs.jax.dev/en/latest/_autosummary/jax.numpy.linalg.eig.html

Under the hood, it uses the same MAGMA kernels as pytorch, so [1] applies. However, https://github.com/jax-ml/jax/issues/27265 tracks a potential update to use the new routines from [2].

* `cupy`

Not available in CuPy 13.4 yet; There is a pull request to add it: https://github.com/cupy/cupy/pull/8980, hopefully to CuPy 14.x.

Proposed API
--------------------

The API should closely follow `eigh` and `eigvalsh`:

- `eigvals(x, /)` should return an array of eigenvalues in an unspecified order;

- `eig(x, /)` should return a namedtuple `(eigenvalues, right eigenvectors)`. The order of eigenvalues and eigenvectors is unspecified, and the only requirement is that `eigenvalues[i]` corresponds to `eigenvectors[:, i]`.  

Whether the eigenvectors have a unit norm is unspecified and thus implementation-defined.

An option to return left eigenvectors may be added in the future if the array libraries add this option (currently, neither numpy nor torch provide this option).

[1] https://github.com/cupy/cupy/issues/6359#issuecomment-1019917948
[2] https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html

rgommers:
[...]

PR #978, opened 2025-10-13
feat: add `eig` and `eigvals` to the `linalg` extension

ev-br:
Following up on an RFC at https://github.com/data-apis/array-api/issues/935, add non-symmetric eigenvalue solvers, `eig` and `eigvals`, to the `linalg` extension.

The routines are available in `numpy`, `torch`, `jax.numpy` and `cupy` (as of to-be-released cupy 1.14, cf https://github.com/cupy/cupy/pull/8980). The main point of contention in gh-935 was the behavior for all eigenvalues on the real line. NumPy returns a real array and all other libraries return a complex array (with zero imaginary parts if they happen to be zero). 
Based on the discussion in https://github.com/numpy/numpy/issues/29000, it might it be difficult to change in numpy itself. Therefore, the suggestion is to make the precise dtype implementation-defined.
I believe the benefit of having `eig` in the standard outweighs this wart.

One other minor point of inconsistency is <s>https://github.com/jax-ml/jax/issues/32558: `jax` returns a standard-compliant namedtuple for `eigh` but a bare tuple for `eig`. I believe this is a minor issue though. </s> already fixed.

### DIFF
--- a/src/array_api_stubs/_draft/linalg.py
+++ b/src/array_api_stubs/_draft/linalg.py
@@ -359,7 +359,6 @@ def eigvals(x: array, /) -> array:
     """
 
 
-
 def inv(x: array, /) -> array:
     r"""
     Returns the multiplicative inverse of a square matrix (or a stack of square matrices) ``x``.


seberg:
> the preference is to require eigenvalues to be always of a complex dtype.

Grumpy old man hat, but: If the NumPy issue isn't resoelved (and it possibly cannot, at least not for 2 years even if I agree), then I don't like this.

I am not opposed at all to just codifying this, but we should clarify the scope somewhere in this repo more clearly.
As is someone will open an issue on the NumPy repo that NumPy "must follow the Array API" which is nonsense, because if NumPy cannot follow it then we can't put this in to begin with.

To me that is easy to resolve: Accept that exactly matching API in the main namespace is aspirational at best and actually document that clearly somewhere so I can point at it from the NumPy repo.  Otherwise, I don't know what to do, because it seems to me I would have to veto adding this until the NumPy issue is resolved.

rgommers:
I'm fine with accepting this PR as long as NumPy has made a decision to made the change (whether through flipping a switch in a future release, or through a deprecation process). That the standard and NumPy are _temporarily_ misaligned is okay, and `array-api-compat` can be used by library authors to deal with that. A _permanent_ mismatch isn't okay though, that would be a blocker for accepting this PR.

src/array_api_stubs/_draft/linalg.py
```diff
@@ -257,6 +251,117 @@ def eigvalsh(x: array, /) -> array:
...
+
+          p(\lambda) = \operatorname{det}(x - \lambda I_n)
+
+    - does not in general have :math:`n` linearly independent eigenvectors if it has
+      eigenvalues with multiplicity larger than one.
+
+    .. note::
+       The eigenvalues of a non-symmetric real matrix in general are complex: for
+       :math:x \in \mathbb{R}^{n \times n}`, the eigenvalues, :math:`\lambda \in \mathbb{C}`,
+       and may or may not reside on the real axis of the complex plane.
```
kgryte:
```suggestion
       may or may not reside on the real axis of the complex plane.
```

kgryte (approved):
LGTM

PR #984, opened 2025-12-12
[DLPack] Update to include DLPack C Exchange API

tqchen:
This PR updates to include dlpack C exchange api for fast exchange at c-extension level without going through python.

### DIFF
--- a/spec/draft/design_topics/data_interchange.rst
+++ b/spec/draft/design_topics/data_interchange.rst
@@ -85,6 +85,18 @@ page gives a high-level specification for data exchange in Python using DLPack.
    below. They are not required to return an array object from ``from_dlpack``
    which conforms to this standard.
 
+
+DLPack C Exchange API
+---------------------
+
+DLPack 1.3 introduces a C-level exchange API that can be used to speed up
+data exchange between arrays at the C-extension level. This API is available via
+the ``type(array_instance).__dlpack_c_exchange_api__`` attribute on the array type object.
+For more details, see the `Python specification of DLPack <https://dmlc.github.io/dlpack/latest/python_spec.html>`__
+We recommend consumer libraries to start first using the python level ``__dlpack__`` first which will covers
+most cases, then start to use the C-level ``__dlpack_c_exchange_api__`` for performance critical cases.
+
+
 Non-supported use cases
 -----------------------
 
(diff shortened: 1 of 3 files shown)


tqchen:
Respective dlpack PR on python spec https://github.com/dmlc/dlpack/pull/180

tqchen:
cc @rgommers @leofang

spec/draft/design_topics/data_interchange.rst
```diff
@@ -85,6 +85,18 @@ page gives a high-level specification for data exchange in Python using DLPack.
    below. They are not required to return an array object from ``from_dlpack``
    which conforms to this standard.
 
+
+DLPack C Exchange API
+---------------------
+
+DLPack 1.3 introduces a C-level exchange API that can be used to speed up
+data exchange between arrays at the C-extension level. This API is available via
+the ``type(array_instance).__dlpack_c_exchange_api__`` attribute on the array type object.
```
rgommers: This `type` usage is unusual, so it might benefit from an extra sentence here, explaining this is a static global object.

rgommers (commented):
Thanks @tqchen! I refreshed my memory here, I think this PR largely looks good and we're probably ready to put this in. My main question right now is: given it's been in use in PyTorch for a couple of months now, do you have any real-world experience to share so far? No issues, did it help a lot performance-wise, etc.

src/array_api_stubs/_draft/array_object.py
```diff
@@ -17,6 +17,19 @@
...
+    ----------
+    __dlpack_c_exchange_api__: PyCapsule
+        An optional static array type attribute stored in ``type(array_instance).__dlpack_c_exchange_api__``
+        that can be used to retrieve the DLPack C-API exchange API struct in DLPack 1.3 or later to speed up
+        exchange of array data at the C extension level without going through Python-level exchange.
+        See :ref:`data-interchange` section for more details.
+    """
+
+    # use None for placeholder
+    __dlpack_c_exchange_api__: PyCapsule = None
```
rgommers: Rendering this locally shows the one-line description goes missing:

Image

I think the correct fix is to make it an `@property` just like `dtype`, `device` and other properties. These are only stubs that are primarily used for generating documentation, so I don't see a problem with this being a `type`-specific rather than an instance-specific property.

rgommers (approved):
Thanks @tqchen. LGTM now, and CI is happy. I've put a reminder in my calendar to merge this by the end of next week unless there are new unresolved comments.

rgommers:
All right, no more comments so in it went. Thanks again @tqchen!

PR #987, opened 2026-01-16
fix: ensure correct special case for complex-valued input in `sqrt`

ev-br:
Fix what looks like a small typo in a docstring. 
Caught it in https://github.com/data-apis/array-api-tests/pull/409


kgryte (changes requested):
@ev-br Given that there are two mistakes, would you mind backporting to previous revisions of the standard? Can be done in this PR.

PR #999, opened 2026-03-13
docs: add migration guides and tutorial

mtsokol:
This PR adds a migration guide (versions for array consumers and producers) and one migration tutorial showing how a simple power-iteration based algorithm from GraphBLAS can be moved to an Array API compatible version.


spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,205 @@
...
+(array-api-strict)=
+
+### Array API Strict
+
+GitHub: [array-api-strict](https://github.com/data-apis/array-api-strict)
+
+`array-api-strict` is a library that provides a strict and minimal
+implementation of the Array API Standard. It is designed to be used as
+a reference implementation for testing and development purposes. By comparing
+your API calls with `array-api-strict` counterparts, you can ensure that your
```
ev-br: This suggestion reads as aimed at array API producers. Consider expanding to mention that an array API consumer can run their test suite with array-api-strict as a producer.

As a larger edit, consider briefly stressing in these short descriptions that array-api-compat is mainly for consumer's code, array-api-strict is a testing implement, also mostly for a consumer; array-api-tests is for a producer. While you are discussing this in more detail below, I think it'd be helpful to stress it in the very beginning, too.

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,205 @@
...
+
+The setup details are enclosed in the GitHub repository, so here we
+cover only the minimal workflow:
+
+1. Install your package, for example in editable mode.
+2. Clone `array-api-tests`, and set `ARRAY_API_TESTS_MODULE` environment
+   variable to your package import name.
+3. Inside the `array-api-tests` directory run `pytest` command. There are
+   multiple useful options delivered by the test suite, a few worth mentioning:
+   - `--max-examples=2` - maximal number of test cases to generate by the
```
ev-br: Please show a larger value, at least `--max-examples=100` or `1000`. That `hypothesis` is not deterministic is very much not obvious for a casual reader. Running two examples (one for "unvectorized" tests) would pretty much only check that a function imports.

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,205 @@
...
+The setup details are enclosed in the GitHub repository, so here we
+cover only the minimal workflow:
+
+1. Install your package, for example in editable mode.
+2. Clone `array-api-tests`, and set `ARRAY_API_TESTS_MODULE` environment
+   variable to your package import name.
+3. Inside the `array-api-tests` directory run `pytest` command. There are
+   multiple useful options delivered by the test suite, a few worth mentioning:
+   - `--max-examples=2` - maximal number of test cases to generate by the
+     hypothesis. This allows you to balance between execution time of the test
```
ev-br, line 130: Minor: add a link to https://hypothesis.readthedocs.io/en/latest/ + a couple of sentensies to explicitly state that hypothesis generates this many examples, where each example is a valid combination of inputs; and that the recommendation is to use as many examples as reasonable for the time budget.

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,205 @@
...
+     hypothesis. This allows you to balance between execution time of the test
+     suite and thoroughness of the testing.
+   - With `--xfails-file` option you can describe which tests are expected to
+     fail - it's impossible to get the whole API perfectly implemented on a
+     first try, so tracking what still fails gives you more control over the
+     state of your API.
+   - `-o xfail_strict=<bool>` is often used with the previous one. If a test
+     expected to fail actually passes (`XPASS`) then you can decide whether
+     to ignore that fact or raise it as an error.
+   - `--skips-file` for skipping files. At times some failing tests might stall
```
ev-br:
```suggestion
   - `--skips-file` for skipping tests. At times some failing tests might stall
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,205 @@
...
+API. For a reference here's a [NumPy Array API Tests CI setup](https://github.com/numpy/numpy/blob/581d10f43b539a189a2d37856e5130464de9e5f6/.github/workflows/linux.yml#L296).
+
+
+#### Array API Strict
+
+A simpler, and more manual, way of testing the Array API coverage is to
+run your API calls along with {ref}`array-api-strict` Python implementation.
+
+This way you can ensure the outputs coming from your API match the minimal
+reference implementation, but bare in mind you need to write the tests cases
```
ev-br:
```suggestion
reference implementation, but bear in mind you need to write the tests cases
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,205 @@
...
+library**. For instance, if you use NumPy for arrays, then your code could
+contain:
+
+```python
+import numpy as np
+
+# ...
+b = np.full(shape, val, dtype=dtype) @ a
+c = np.mean(a, axis=0)
+return np.dot(c, b)
```
ev-br, line 177: Nit, optional: are you deliberately showing the transformation from `np.dot` to a spec-compliant `xp.tensordot` below? Otherwise at least one reader (me) stops here and wonders about a subtle difference between a matmul two lines above and  `np.dot` here.

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,205 @@
...
+
+# ...
+b = np.full(shape, val, dtype=dtype) @ a
+c = np.mean(a, axis=0)
+return np.dot(c, b)
+```
+
+The first step should be as simple as assigning `np` namespace to a dedicated
+namespace variable - the convention in the ecosystem is to name it `xp`. Then
+Making sure that each method and function call is something that Array API
```
ev-br: typo "Then Making sure...."

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,205 @@
...
+return xp.tensordot(c, b, axes=1)
+```
+
+Then replacing one backend with another one should rely on providing a different
+namespace, such as: `xp = torch`, e.g. via environment variable. This can be useful
+if you're writing a script or in your custom software. The other alternatives are:
+
+- If you are building a library where the backend is determined by input arrays
+  passed by the end-user, then a recommended way is to ask your input arrays for a
+  namespace to use: `xp = arr.__array_namespace__()`
```
ev-br: For consumers, the support is less than ideal, consider mentioning `array-api-compat` and `array_namespace` function already?

```
In [9]: for xp in [np, torch, jnp, cupy, da, array_api_strict]:
   ...:     try:
   ...:         xp.ones(3).__array_namespace__()
   ...:         print(f"{xp.__name__}  check")
   ...:     except:
   ...:         print(f"{xp.__name__}  nope")
   ...: 
numpy  check
torch  nope
jax.numpy  check
cupy  nope
dask.array  nope
array_api_strict  check
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,205 @@
...
+- Each function you implement can have a namespace `xp` as a parameter in the
+  signature. Then enforcing inputs to be of type by the provided backend can be
+  achieved with `arg1 = xp.asarray(arg1)` for each input array.
```
ev-br: If this is a first tutorial for a beginner, I think it could be helpful to be a bit more descriptive and a bit more opinionated. Roughly, 

- for a function which accepts array arguments, use
```
def func(array1, scalar1, scalar2):
    xp = array1.__array_namespace__()     # or array_namespace(array1)
    return xp.arange(scalar1, scalar2) @ array1
```

- for a function that accepts scalars and returns arrays, use
```
def func(s1, s2, xp):
   return xp.arange(s1, s2)
```

If you prefer to not clutter the short guide with details, consider adding them to a separate page and linking from the short guide.

spec/2025.12/tutorial_basic.md
```diff
@@ -0,0 +1,112 @@
...
+
+def is_converged(xprev, x, N, tol):
+    err = xp.sum(xp.abs(x - xprev))
+    return err < xp.asarray(N * tol)
+```
+
+At this point the actual execution depends only on `xp` namespace,
+and replacing that one variable allow us to switch from e.g. NumPy arrays
+to a JAX execution on a GPU. This allows us to be more flexible, and, for
+example use lazy evaluation and JIT compile a loop body with JAX's JIT compilation.
```
ev-br: But jax jitting this iterative function with data-dependent control flow doesn't work! Simplifying a bit the `hits` function to accept an array-like, and jitting --- which is what I would do first thing after reading this tutorial--- gives


```
In [19]: def hits(A, max_iter=100, tol=1.0e-8, normalized=True):
    ...:     A = xp.asarray(A)
    ...:     N = A.shape[0]
    ...:     h = xp.full(N, 1.0 / N)
    ...:     # Power iteration: make up to max_iter iterations
    ...:     for _i in range(max_iter):
    ...:         hprev = h
    ...:         a = hprev @ A
    ...:         h = A @ a
    ...:         h = h / xp.max(h)
    ...:         if is_converged(hprev, h, N, tol):
    ...:             break
    ...:     else:
    ...:         raise Exception("Didn't converge")
    ...:     if normalized:
    ...:         h = h / xp.sum(xp.abs(h))
    ...:         a = a / xp.sum(xp.abs(a))
    ...:     return h, a
    ...: 
    ...: def is_converged(xprev, x, N, tol):
    ...:     err = xp.sum(xp.abs(x - xprev))
    ...:     return err < xp.asarray(N * tol)
    ...: 

In [20]: import jax.numpy as xp

In [21]: jax.jit(hits)(jnp.eye(3))
TracerBoolConversionError                 Traceback (most recent call last)
Cell In[21], line 1
----> 1 jax.jit(hits)(jnp.eye(3))

    [... skipping hidden 13 frame]

Cell In[19], line 11, in hits(A, max_iter, tol, normalized)
      9     h = A @ a
     10     h = h / xp.max(h)
---> 11     if is_converged(hprev, h, N, tol):
     12         break
     13 else:

    [... skipping hidden 1 frame]

File ~/.conda/envs/scipy-dev/lib/python3.12/site-packages/jax/_src/core.py:1806, in concretization_function_error.<locals>.error(self, arg)
   1805 def error(self, arg):
-> 1806   raise TracerBoolConversionError(arg)

TracerBoolConversionError: Attempted boolean conversion of traced array with shape bool[].
The error occurred while tracing the function hits at <ipython-input-19-d5a4e616b039>:1 for jit. This concrete value was not available in Python because it depends on the value of the argument A.
See https://docs.jax.dev/en/latest/errors.html#jax.errors.TracerBoolConversionError
```

spec/draft/tutorial_basic.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+We replace custom binary operation with the Array API counterpart:
+
+```python
+...(x - xprev)
+```
+
+And last but not least, let's ensure that the result of the convergence
+condition is a scalar coming from our API:
```
ev-br, line 75: I honestly don't understand this statement. `err` is an array, you are converting the r.h.s. to a 0D too, so `is_converged` function returns a 0D array, and `if is_converged(...)` relies on an automagic `bool(0D_array)` conversion. Which would also synchronize a device when run on a GPU.

ev-br (commented):
This is very nice.

I've a left several comments, mostly very minor. There are two themes:
- for the migration guide, I think it would be helpful to be some more specific and some more opinionated. 
- In the tutorial, I love the power iteration example. The only concern is that you talk about JIT, but the example does not jit, not easily at least.  So maybe either expand a bit on how to actually jit it, or state that not everything is benefits from jitting and add one more example which does?

spec/2025.12/migration_guide.md
mtsokol: Done - I added a `User group` clause to each library and mentioned how producers/consumers use it.

spec/2025.12/migration_guide.md
mtsokol: Done

spec/draft/migration_guide.md
mtsokol: Done

spec/2025.12/migration_guide.md
mtsokol: Done

spec/2025.12/migration_guide.md
mtsokol: Done

spec/2025.12/migration_guide.md
mtsokol: Done

spec/2025.12/migration_guide.md
mtsokol: Done

spec/2025.12/tutorial_basic.md
mtsokol: Sorry for not being clear in the text. It says `JIT compile a loop body with JAX's JIT compilation`, so to JIT compile what is in the `for _i in range(...)`. That is the hot part that we want to compile into a single kernel.
It's the same premise as in [the JAX documentation](https://docs.jax.dev/en/latest/jit-compilation.html#why-can-t-we-just-jit-everything):

image

To make it fully digestible by a beginner user I embedded a copy-paste code at the bottom with JAX jit compiled version. Is it sufficiently clear?

spec/draft/tutorial_basic.md
mtsokol: Here rhs and lhs are 0D arrays in the current form, and the result is a 0D so that it can be used as `if` condition.
We're outside of JIT compiled part (as explained in the previous comment) so we don't need to worry about synchronization.

spec/2025.12/migration_guide.md
mtsokol: Done - I mentioned `array_api_compat.array_namespace()` as well there.

spec/draft/migration_guide.md
mtsokol: Yes, I added a note that `dot` is not present in the Array API and that this time we use `tensordot` instead.

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
+(migration-guide)=
+
+# Migration Guide
+
+This page is meant to help migrate your codebase to an Array API compliant
+implementation. The guide is divided into two parts and, depending on your
+exact use-case, you should look thoroughly into at least one of them.
+
+The first part  is dedicated for {ref}`array-producers`. If your library
```
kgryte:
```suggestion
The first part is dedicated for {ref}`array-producers`. If your library
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+mimics e.g. NumPy's or Dask's functionality, then you can find there an
```
kgryte:
```suggestion
mimics, for example, NumPy's or Dask's functionality, then you can find in the first part
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+additional instructions and guidance on how to ensure downstream users can
+easily pick your solution as an array provider for their system/algorithm.
+
+The second part delves into details for Array API compatibility for
+{ref}`array-consumers`. This pertains to any software that performs
+multidimensional array manipulation in Python, such as: scikit-learn, SciPy,
```
kgryte:
```suggestion
multidimensional array manipulation in Python, such as may be found in scikit-learn, SciPy,
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+or statsmodels. If your software relies on a certain array producing library,
+such as NumPy or JAX, then here you can learn how to make it library agnostic
+and interchange them with way less friction.
```
kgryte:
```suggestion
and interchange array namespaces with significantly less friction.
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+The first part  is dedicated for {ref}`array-producers`. If your library
+mimics e.g. NumPy's or Dask's functionality, then you can find there an
+additional instructions and guidance on how to ensure downstream users can
+easily pick your solution as an array provider for their system/algorithm.
+
+The second part delves into details for Array API compatibility for
+{ref}`array-consumers`. This pertains to any software that performs
+multidimensional array manipulation in Python, such as: scikit-learn, SciPy,
+or statsmodels. If your software relies on a certain array producing library,
+such as NumPy or JAX, then here you can learn how to make it library agnostic
```
kgryte:
```suggestion
such as NumPy or JAX, then you can use the second part to learn how to make it library agnostic
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+(array-api-compat)=
+
+### Array API Compat
+
+GitHub: [array-api-compat](https://github.com/data-apis/array-api-compat)
+User group: Array Consumers
+
+Although NumPy, Dask, CuPy, and PyTorch support the Array API Standard, there
+are still some corner cases where their behavior diverges from the standard.
+`array-api-compat` provides a compatibility layer to cover these cases as well.
```
kgryte:
```suggestion
`array-api-compat` provides a compatibility layer to cover these cases.
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+This is also accompanied by a few utility functions for easier introspection
+into array objects. As an array consumer you can still rely on the original API
```
kgryte:
```suggestion
into array objects. As an array consumer, you can still rely on the original API
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+
+(array-api-strict)=
+
+### Array API Strict
+
+GitHub: [array-api-strict](https://github.com/data-apis/array-api-strict)
+User group: Array Consumers, Array Producers (for testing)
+
+`array-api-strict` is a library that provides a strict and minimal
+implementation of the Array API Standard. For array producers it is designed
```
kgryte:
```suggestion
implementation of the Array API Standard. For array producers, it is designed
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+to be used as a reference implementation for testing and development purposes.
+You can compare your API calls with `array-api-strict` counterparts,
```
kgryte:
```suggestion
You can compare your API calls with `array-api-strict` counterparts
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+and ensure that your library is fully compliant with the standard and can
+serve as a reliable reference for other developers in the ecosystem.
+For consumers you can use `array-api-strict` during the development as an
```
kgryte:
```suggestion
For consumers, you can use `array-api-strict` during the development as an
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+
+The sections below mention when and how to use them.
+
+
+(array-producers)=
+
+## Array Producers
+
+For array producers, the central task during the development/migration process
+is adhering user-facing API to the Array API Standard.
```
kgryte:
```suggestion
is ensuring that the user-facing API adheres to the Array API Standard.
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+
+The complete API of the standard is documented on the
+[API specification](https://data-apis.org/array-api/latest/API_specification/index.html)
+page.
```
kgryte:
```suggestion
The complete API of the standard is documented in the
[API specification](https://data-apis.org/array-api/latest/API_specification/index.html).
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+
+There, each function, constant, and object is described with details
+on parameters, return values, and special cases.
+
+### Testing against Array API
+
+There are two main ways to test your API for compliance: Either using
```
kgryte:
```suggestion
There are two main ways to test your API for compliance: either using
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+`array-api-tests` suite or testing your API manually against `array-api-strict`
```
kgryte:
```suggestion
`array-api-tests` suite or testing your API manually against the `array-api-strict`
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+reference implementation.
+
+#### Array API Test suite (Recommended)
+
+{ref}`array-api-tests` is a test suite which verifies that your API
+for adhering to the standard. For each function or method it confirms
```
kgryte:
```suggestion
adheres to the standard. For each function or method, it confirms
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+it's importable, verifies the signature, and generates multiple test
+cases with [hypothesis](https://hypothesis.readthedocs.io/en/latest/)
+package and runs asserts for the outputs.
```
kgryte:
```suggestion
it's importable, verifies the signature, generates multiple test
cases with [hypothesis](https://hypothesis.readthedocs.io/en/latest/)
package, and runs assertions on the outputs.
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+
+The setup details are enclosed in the GitHub repository, so here we
+cover only the minimal workflow:
+
+1. Install your package, for example in editable mode.
+2. Clone `array-api-tests`, and set `ARRAY_API_TESTS_MODULE` environment
+   variable to your package import name.
+3. Inside the `array-api-tests` directory run `pytest` command. There are
+   multiple useful options delivered by the test suite, a few worth mentioning:
```
kgryte:
```suggestion
   multiple useful options delivered by the test suite. A few worth mentioning:
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+   - `--max-examples=1000` - maximal number of test cases to generate by the
+     hypothesis. This allows you to balance between execution time of the test
+     suite and thoroughness of the testing. It's advised to use as many examples
+     as the time buget can fit. Each test case is a random combination of
+     possible inputs: the more cases, the higher chance of finding an unsupported
+     edge case.
+   - With `--xfails-file` option you can describe which tests are expected to
```
kgryte:
```suggestion
   - With the `--xfails-file` option, you can describe which tests are expected to
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+     fail - it's impossible to get the whole API perfectly implemented on a
```
kgryte:
```suggestion
     fail. It's impossible to get the whole API perfectly implemented on a
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+     first try, so tracking what still fails gives you more control over the
+     state of your API.
+   - `-o xfail_strict=<bool>` is often used with the previous one. If a test
+     expected to fail actually passes (`XPASS`) then you can decide whether
```
kgryte:
```suggestion
     expected to fail actually passes (`XPASS`), then you can decide whether
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+     to ignore that fact or raise it as an error.
+   - `--skips-file` for skipping tests. At times some failing tests might stall
```
kgryte:
```suggestion
   - `--skips-file` for skipping tests. At times, some failing tests might stall
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+     the execution time of the test suite - in that case the most convenient
```
kgryte:
```suggestion
     the execution time of the test suite. In that case, the most convenient
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+     option is to skip these for the time being.
+
+We strongly advise you to embed this setup in your CI as well. This will allow
+you to monitor the coverage live, and make sure new changes don't break existing
+API. For a reference here's a [NumPy Array API Tests CI setup](https://github.com/numpy/numpy/blob/581d10f43b539a189a2d37856e5130464de9e5f6/.github/workflows/linux.yml#L296).
```
kgryte:
```suggestion
APIs. For a reference, here's a [NumPy Array API Tests CI setup](https://github.com/numpy/numpy/blob/581d10f43b539a189a2d37856e5130464de9e5f6/.github/workflows/linux.yml#L296).
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+
+
+#### Array API Strict
+
+A simpler, and more manual, way of testing the Array API coverage is to
+run your API calls along with {ref}`array-api-strict` Python implementation.
```
kgryte:
```suggestion
run your API calls along with the {ref}`array-api-strict` Python implementation.
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+
+This way you can ensure the outputs coming from your API match the minimal
+reference implementation, but bear in mind you need to write the tests cases
```
kgryte:
```suggestion
reference implementation. Bear in mind, however, that you need to write the tests cases
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+yourself, so you need to also take into account the edge cases as well.
+
+
+(array-consumers)=
+
+## Array Consumers
+
+For array consumers the main premise is to keep in mind that your **array
```
kgryte:
```suggestion
For array consumers, the main premise is to keep in mind that your **array
```

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+import numpy as np
+
+# ...
+b = np.full(shape, val, dtype=dtype) @ a
+c = np.mean(a, axis=0)
+return np.dot(c, b)
+```
+
+The first step should be as simple as assigning `np` namespace to a dedicated
+namespace variable - the convention in the ecosystem is to name it `xp`. Then
```
kgryte:
```suggestion
namespace variable. The convention in the ecosystem is to name it `xp`. Then
```

spec/draft/migration_guide.md
kgryte: Why tensordot and not vecdot?

spec/2025.12/migration_guide.md
```diff
@@ -0,0 +1,228 @@
...
+- [array_namespace()](https://data-apis.org/array-api-compat/helper-functions.html#array_api_compat.array_namespace)
+  for fetching the namespace based on input arrays.
+- [is_array_api_obj()](https://data-apis.org/array-api-compat/helper-functions.html#array_api_compat.is_array_api_obj)
+  for the introspection whether a given object is Array API compatible.
+- [device()](https://data-apis.org/array-api-compat/helper-functions.html#array_api_compat.device)
+  to get a device the array resides on.
+
+For now the migration from a specific library (e.g. NumPy) to a standard compatible
+setup requires a manual intervention for each failing API call but in the future
+we plan to provide some automation tools for it.
```
kgryte:
```suggestion
For now. the migration from a specific library (e.g., NumPy) to a standard compatible
setup requires a manual intervention for each failing API call, but, in the future,
we plan to provide some automation tools for it.
```

mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/2025.12/migration_guide.md
mtsokol: done

spec/draft/migration_guide.md
mtsokol: In general I wanted to show a case where a given function isn't present in the standard and we need to switch to something else. The code sample is small enough that you can't tell which one is the right choice: assuming that `c` and `b` are rank 2 or higher then `tensordot(..., axes=1)` is closer to `dot(...)` (only the selection of `b` axis is different between them) than to `vecdot`, which performs only vector product.

If we want to be strict in this code example I can make it unambiguous.

kgryte: May be worth more explicitly spelling what you describe above out.

mtsokol: Sure - I'll make an assumption about `c` and `b` in the text explaining the code snippet.

spec/2025.12/tutorial_basic.md
ev-br: Yes, thank you. The updated example is very clear and instructive.

ev-br (approved):

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+User group: Array Consumers, Array Producers (for testing)
+
+`array-api-strict` is a library that provides a strict and minimal
+implementation of the Array API Standard. For array producers, it is designed
+to be used as a reference implementation for testing and development purposes.
+You can compare your API calls with `array-api-strict` counterparts and
+ensure that your library is fully compliant with the standard and can
+serve as a reliable reference for other developers in the ecosystem.
+For consumers, you can use `array-api-strict` during the development as an
+array provider to ensure your code uses API compliant with the standard.
```
kgryte:
```suggestion
array provider to ensure your code uses APIs compliant with the standard.
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+
+GitHub: [array-api-tests](https://github.com/data-apis/array-api-tests)
+
+User group: Array Producers
+
+`array-api-tests` is a collection of tests that can be used to verify the
+compliance of your library with the Array API Standard. It includes tests
+for array producers, covering a wide range of functionalities and use cases.
+By running these tests, you can ensure that your library adheres to the
+standard and can be used with compatible array consumers libraries.
```
kgryte:
```suggestion
standard and can be used with compatible array consumer libraries.
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+There are two main ways to test your API for compliance: either using
+`array-api-tests` suite or testing your API manually against the
+`array-api-strict` reference implementation.
+
+#### Array API Test suite (Recommended)
+
+{ref}`array-api-tests` is a test suite which verifies that your API
+adheres to the standard. For each function or method, it confirms
+it's importable, verifies the signature, generates multiple test
+cases with [hypothesis](https://hypothesis.readthedocs.io/en/latest/)
```
kgryte:
```suggestion
cases with the [hypothesis](https://hypothesis.readthedocs.io/en/latest/)
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+package, and runs assertions on the outputs.
+
+The setup details are enclosed in the GitHub repository, so here we
+cover only the minimal workflow:
+
+1. Install your package, for example in editable mode.
```
kgryte:
```suggestion
1. Install your package (e.g., in editable mode).
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+2. Clone `array-api-tests`, and set `ARRAY_API_TESTS_MODULE` environment
```
kgryte:
```suggestion
2. Clone `array-api-tests`, and set the `ARRAY_API_TESTS_MODULE` environment
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+   variable to your package import name.
+3. Inside the `array-api-tests` directory run `pytest` command. There are
```
kgryte:
```suggestion
3. Inside the `array-api-tests` directory run the command for running pytest: `pytest`. There are
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+   multiple useful options delivered by the test suite. A few worth mentioning:
+   - `--max-examples=1000` - maximal number of test cases to generate by the
```
kgryte:
```suggestion
   - `--max-examples=1000` - maximal number of test cases to generate when using
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+     hypothesis. This allows you to balance between execution time of the test
+     suite and thoroughness of the testing. It's advised to use as many examples
+     as the time buget can fit. Each test case is a random combination of
+     possible inputs: the more cases, the higher chance of finding an
+     unsupported edge case.
+   - With the `--xfails-file` option, you can describe which tests are expected
+     to fail. It's impossible to get the whole API perfectly implemented on a
+     first try, so tracking what still fails gives you more control over the
+     state of your API.
+   - `-o xfail_strict=<bool>` is often used with the previous one. If a test
```
kgryte:
```suggestion
   - `-o xfail_strict=<bool>` is often used with the previous option. If a test
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+     expected to fail actually passes (`XPASS`), then you can decide whether
+     to ignore that fact or raise it as an error.
+   - `--skips-file` for skipping tests. At times, some failing tests might stall
+     the execution time of the test suite. In that case, the most convenient
+     option is to skip these for the time being.
+
+We strongly advise you to embed this setup in your CI as well. This will allow
+you to monitor the coverage live, and make sure new changes don't break existing
```
kgryte:
```suggestion
you to continuously monitor Array API coverage, and make sure new changes don't break existing
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+APIs. For a reference, here's a [NumPy Array API Tests CI setup](https://github.com/numpy/numpy/blob/581d10f43b539a189a2d37856e5130464de9e5f6/.github/workflows/linux.yml#L296).
```
kgryte:
```suggestion
APIs. As a reference, see [NumPy's Array API Tests CI setup](https://github.com/numpy/numpy/blob/581d10f43b539a189a2d37856e5130464de9e5f6/.github/workflows/linux.yml#L296).
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+
+
+#### Array API Strict
+
+A simpler, and more manual, way of testing the Array API coverage is to
```
kgryte:
```suggestion
A simpler, and more manual, way of testing Array API coverage is to
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+run your API calls along with the {ref}`array-api-strict` Python implementation.
+
+This way you can ensure the outputs coming from your API match the minimal
```
kgryte:
```suggestion
This way, you can ensure that the outputs coming from your API match the minimal
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+reference implementation. Bear in mind, however, that you need to write
+the tests cases yourself, so you need to also take into account the edge
+cases as well.
```
kgryte:
```suggestion
the tests cases yourself, so you need to also take into account any applicable edge
cases.
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+```python
+import numpy as np
+
+# ...
+b = np.full(shape, val, dtype=dtype) @ a
+c = np.mean(a, axis=0)
+return np.dot(c, b)
+```
+
+The first step should be as simple as assigning `np` namespace to a dedicated
```
kgryte:
```suggestion
The first step should be as simple as assigning the `np` namespace to a dedicated
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+namespace variable. The convention in the ecosystem is to name it `xp`. Then
```
kgryte:
```suggestion
namespace variable. The convention used in the ecosystem is to name it `xp`. Then,
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+making sure that each method and function call is something that Array API
```
kgryte:
```suggestion
it is vital to ensure that each method and function call is something that the Array API
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+supports is vital. `dot` is present in the NumPy's API but the standard
```
kgryte:
```suggestion
supports. For example, `dot` is present in the NumPy's API, but the standard
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+doesn't support it. For the sake of simplicity let's assume both `c` and `b`
```
kgryte:
```suggestion
doesn't support it. For the sake of simplicity, let's assume both `c` and `b`
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+are `ndim=2`, therefore we select `tensordot` instead - both NumPy and the
```
kgryte:
```suggestion
are `ndim=2`; therefore, we select `tensordot` instead, as both NumPy and the
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+
+xp = np
+
+# ...
+b = xp.full(shape, val, dtype=dtype) @ a
+c = xp.mean(a, axis=0)
+return xp.tensordot(c, b, axes=1)
+```
+
+Then replacing one backend with another one should rely on providing a different
```
kgryte:
```suggestion
At this point, replacing one backend with another one should only require providing a different
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+namespace, such as: `xp = torch`, e.g. via environment variable. This can be useful
```
kgryte:
```suggestion
namespace, such as `xp = torch` (e.g., via an environment variable). This can be useful
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+  your input arrays for a namespace to use: `xp = arr.__array_namespace__()`.
+  If the given library doesn't have it, then [`array_api_compat.array_namespace()`](https://data-apis.org/array-api-compat/helper-functions.html#array_api_compat.array_namespace)
+  should be used instead:
+  ```python
+  def func(array1, scalar1, scalar2):
+    xp = array1.__array_namespace__()  # or array_namespace(array1)
+    return xp.arange(scalar1, scalar2) @ array1
+  ```
+- For a function that accepts scalars and returns arrays, use namespace `xp` as
+  a parameter in the signature. Then enforcing objects to be of type by the
```
kgryte:
```suggestion
  a parameter in the signature. Enforcing objects to have the same type as the
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+  provided backend can be achieved with `arg1 = xp.asarray(arg1)` for each input:
```
kgryte:
```suggestion
  provided backend can then be achieved with `arg1 = xp.asarray(arg1)` for each input:
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+  ```python
+  def func(s1, s2, xp):
+    return xp.arange(s1, s2)
+  ```
+
+If you're relying on NumPy, CuPy, PyTorch, Dask, or JAX then
+{ref}`array-api-compat` can come in handy for the transition. The compat layer
+allows you to still rely on your selection of array producing library, while
```
kgryte:
```suggestion
allows you to still rely on your preferred array producing library, while
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+making sure you're already using standard compatible API. Additionally, it
+offers a set of useful utility functions, such as:
+
+- [array_namespace()](https://data-apis.org/array-api-compat/helper-functions.html#array_api_compat.array_namespace)
+  for fetching the namespace based on input arrays.
+- [is_array_api_obj()](https://data-apis.org/array-api-compat/helper-functions.html#array_api_compat.is_array_api_obj)
+  for the introspection whether a given object is Array API compatible.
```
kgryte:
```suggestion
  for inspecting whether a given object is Array API compatible.
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+- [device()](https://data-apis.org/array-api-compat/helper-functions.html#array_api_compat.device)
+  to get a device the array resides on.
```
kgryte:
```suggestion
  for retrieving the device on which an array resides.
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+
+For now, the migration from a specific library (e.g., NumPy) to a standard
+compatible setup requires a manual intervention for each failing API call,
+but, in the future, we plan to provide some automation tools for it.
```
kgryte:
```suggestion
but, in the future, we're hoping to provide tools for automating the migration process.
```

spec/draft/tutorial_basic.md
```diff
@@ -0,0 +1,158 @@
+(tutorial-basic)=
+
+# Array API Tutorial
+
+In this tutorial we're going to show the migration from the array consumer
```
kgryte:
```suggestion
In this tutorial, we're going to demonstrate how to migrate to the Array API from the array consumer's
```

spec/draft/tutorial_basic.md
```diff
@@ -0,0 +1,158 @@
...
+point of view for a simple graph algorithm.
+
+The example presented here comes from [`graphblas-algorithms`](https://github.com/python-graphblas/graphblas-algorithms).
```
kgryte:
```suggestion
The example presented here comes from the [`graphblas-algorithms`](https://github.com/python-graphblas/graphblas-algorithms).
```

spec/draft/tutorial_basic.md
```diff
@@ -0,0 +1,158 @@
...
+library. There we can find [the HITS algorithm](https://github.com/python-graphblas/graphblas-algorithms/blob/35dbc90e808c6bf51b63d51d8a63f59238c02975/graphblas_algorithms/algorithms/link_analysis/hits_alg.py#L9),
```
kgryte:
```suggestion
library. In particular, we'll be migrating [the HITS algorithm](https://github.com/python-graphblas/graphblas-algorithms/blob/35dbc90e808c6bf51b63d51d8a63f59238c02975/graphblas_algorithms/algorithms/link_analysis/hits_alg.py#L9), which is
```

spec/draft/tutorial_basic.md
```diff
@@ -0,0 +1,158 @@
...
+used for the link analysis for estimating prominence in sparse networks.
```
kgryte:
```suggestion
used for the link analysis for estimating prominence in sparse networks, to be Array API compliant.
```

spec/draft/tutorial_basic.md
```diff
@@ -0,0 +1,158 @@
...
+
+The inlined and slightly simplified (without "authority" feature)
+implementation looks like this:
```
kgryte:
```suggestion
implementation looks similar to the following:
```

spec/draft/tutorial_basic.md
```diff
@@ -0,0 +1,158 @@
...
+    return h, a
+
+def is_converged(xprev, x, tol):
+    xprev << binary.minus(xprev | x)
+    xprev << unary.abs(xprev)
+    return xprev.reduce().get(0) < xprev.size * tol
+```
+
+We can see that the API is specific to the GraphBLAS array object.
+There is `Vector` constructor, overloaded `<<` for assigning new values,
```
kgryte:
```suggestion
There is a `Vector` constructor, overloaded `<<` for assigning new values,
```

spec/draft/tutorial_basic.md
```diff
@@ -0,0 +1,158 @@
...
+and `reduce`/`get` for reductions. We need to replace them, and, by convention,
+we will use `xp` namespace for calling respective functions.
+
+First we want to make sure we construct arrays in an agnostic way:
```
kgryte:
```suggestion
First, we want to make sure we construct arrays in an agnostic way:
```

spec/draft/tutorial_basic.md
```diff
@@ -0,0 +1,158 @@
...
+
+```python
+h = xp.full(N, 1.0 / N)
+A = xp.asarray(G.A)
+```
+
+Then, instead of `reduce` calls we use appropriate reducing
```
kgryte:
```suggestion
Then, instead of `reduce` calls, we will use appropriate reduction
```

spec/draft/tutorial_basic.md
```diff
@@ -0,0 +1,158 @@
...
+```python
+h = h / xp.max(h)
+# ...
+h = h / xp.sum(xp.abs(h))
+a = a / xp.sum(xp.abs(a))
+# ...
+err = xp.sum(xp.abs(...))
+```
+
+We replace custom binary operation with the Array API counterpart:
```
kgryte:
```suggestion
We replace the custom binary operation with the Array API counterpart:
```

spec/draft/tutorial_basic.md
```diff
@@ -0,0 +1,158 @@
...
+
+```python
+...(x - xprev)
+```
+
+And last but not least, let's ensure that the result of the convergence
```
kgryte:
```suggestion
And finally, let's ensure that the result of the convergence
```

spec/draft/tutorial_basic.md
```diff
@@ -0,0 +1,158 @@
...
+        h = h / xp.sum(xp.abs(h))
+        a = a / xp.sum(xp.abs(a))
+    return h, a
+
+def is_converged(xprev, x, N, tol):
+    err = xp.sum(xp.abs(x - xprev))
+    return err < xp.asarray(N * tol)
+```
+
+At this point the actual execution depends only on `xp` namespace,
```
kgryte:
```suggestion
At this point, the actual execution depends only on `xp` namespace,
```

spec/draft/tutorial_basic.md
```diff
@@ -0,0 +1,158 @@
...
+and replacing that one variable allow us to switch from e.g. NumPy arrays
```
kgryte:
```suggestion
and replacing that one variable will allow us to switch from, e.g., NumPy arrays on the CPU
```

spec/draft/tutorial_basic.md
```diff
@@ -0,0 +1,158 @@
...
+to a JAX execution on a GPU. This lets us be more flexible, and, for example,
```
kgryte:
```suggestion
to JAX arrays for running on a GPU. This lets us be more flexible, and, for example,
```

kgryte (approved):
LGTM. Thanks, @mtsokol, and @ev-br for the review!

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
+(migration-guide)=
+
+# Migration Guide
+
+This page is meant to help migrate your codebase to an Array API compliant
```
lucascolley, line 5: here and throughout (maybe dropping 'standard' makes sense in some cases, but see https://github.com/data-apis/array-api/discussions/778 )

```suggestion
This page is meant to help migrate your codebase to an array API standard compliant
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+implementation. The guide is divided into two parts and, depending on your
+exact use-case, you should look thoroughly into at least one of them.
+
+The first part is dedicated for {ref}`array-producers`. If your library
+mimics, for example, NumPy's or Dask's functionality, then you can find in
```
lucascolley, line 10: minor nit, Dask is probably a strange choice here since itself is quite firmly mimicking NumPy. Maybe PyTorch would be a better example of where the standard took influence to differ from (historical) NumPy

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
+(migration-guide)=
+
+# Migration Guide
+
+This page is meant to help migrate your codebase to an Array API compliant
+implementation. The guide is divided into two parts and, depending on your
+exact use-case, you should look thoroughly into at least one of them.
```
lucascolley, line 7: nit, seems unnecessary to me
```suggestion
implementation. The guide is divided into two parts.
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+downstream users can easily pick your solution as an array provider for
+their system/algorithm.
+
+The second part delves into details for Array API compatibility for
+{ref}`array-consumers`. This pertains to any software that performs
+multidimensional array manipulation in Python, such as may be found in
+scikit-learn, SciPy, or statsmodels. If your software relies on a certain
+array producing library, such as NumPy or JAX, then you can use the second
+part to learn how to make it library agnostic and interchange array
+namespaces with significantly less friction.
```
lucascolley, line 21: is "interchange array namespaces" really what the second part is helping with? To me it seems focused on making functions agnostic

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
+(migration-guide)=
+
+# Migration Guide
+
+This page is meant to help migrate your codebase to an Array API compliant
+implementation. The guide is divided into two parts and, depending on your
```
lucascolley, line 6: I think this should be changed since there are actually three parts, the first being `Ecosystem`

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+
+### Array API Compat
+
+GitHub: [array-api-compat](https://github.com/data-apis/array-api-compat)
+
+User group: Array Consumers
+
+Although NumPy, Dask, CuPy, and PyTorch support the Array API Standard, there
+are still some corner cases where their behavior diverges from the standard.
+`array-api-compat` provides a compatibility layer to cover these cases.
```
lucascolley, line 39: I think this is false and conflicts with https://data-apis.org/array-api/draft/purpose_and_scope.html#conformance. Sure, it's unreasonable to require _100%_ conformance to say that a library 'supports the standard', but especially in the cases of Dask and PyTorch there are places where array-api-compat still has to do significant work to add missing support

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+You can compare your API calls with `array-api-strict` counterparts and
+ensure that your library is fully compliant with the standard and can
+serve as a reliable reference for other developers in the ecosystem.
+For consumers, you can use `array-api-strict` during the development as an
+array provider to ensure your code uses APIs compliant with the standard.
+
+
+(array-api-tests)=
+
+### Array API Test
```
lucascolley, line 65:
```suggestion
### Array API Tests
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+
+## Ecosystem
+
+Apart from the documented standard, the Array API ecosystem also provides
+a set of tools and packages to help you with the migration process:
+
+
+(array-api-compat)=
+
+### Array API Compat
```
lucascolley, line 31: I think this is the first I have seen where we refer to these names with capitalisation and spaces? https://data-apis.org/array-api-extra/ for example deliberately keeps the styling lower-case and with hyphens. Especially since these libraries are developer-facing, I think it makes sense to keep the style of names consistent with how they are distributed as packages.

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+GitHub: [array-api-strict](https://github.com/data-apis/array-api-strict)
+
+User group: Array Consumers, Array Producers (for testing)
+
+`array-api-strict` is a library that provides a strict and minimal
+implementation of the Array API Standard. For array producers, it is designed
+to be used as a reference implementation for testing and development purposes.
+You can compare your API calls with `array-api-strict` counterparts and
+ensure that your library is fully compliant with the standard and can
+serve as a reliable reference for other developers in the ecosystem.
```
lucascolley, line 58: I would probably omit this part. The primary source of truth is the standard, with the second being array-api-tests. I don't think array-api-strict provides a particularly useful way to "ensure that your library is fully compliant", aside from comparing every permutation of argument combinations on all functions.

Do any of the libraries adopting the standard actually have infrastructure that compares results against array-api-strict? I was under the assumption that they all just use array-api-tests. (The exception to that is libraries which are inherently both consumers and producers, like https://mdhaber.github.io/marray/intro.html, but I don't think such libraries are the target audience of this doc.)

EDIT: ahh, I see now below the section on testing against array-api-strict for producers. Yeah, especially since that is marked as not the recommended way, I think it is best to omit it in this section here. Makes sense to point out to producers in their section that it is something that might be worth doing, but I don't think it should be included in the description of array-api-strict's purpose.

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+For consumers, you can use `array-api-strict` during the development as an
+array provider to ensure your code uses APIs compliant with the standard.
```
lucascolley, line 60: I would probably be explicit about parametrising tests with it as an array namespace, see https://lucascolley.github.io/talks/pydata-paris-25-array-api/#/5/10. 'using during development' is maybe a bit opaque for someone who is new to the standard.

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+API while having access to the standard compatible one.
+
+
+(array-api-strict)=
+
+### Array API Strict
+
+GitHub: [array-api-strict](https://github.com/data-apis/array-api-strict)
+
+User group: Array Consumers, Array Producers (for testing)
```
lucascolley, line 51: see other comment on the inclusion of array producers here

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+
+GitHub: [array-api-tests](https://github.com/data-apis/array-api-tests)
+
+User group: Array Producers
+
+`array-api-tests` is a collection of tests that can be used to verify the
+compliance of your library with the Array API Standard. It includes tests
+for array producers, covering a wide range of functionalities and use cases.
+By running these tests, you can ensure that your library adheres to the
+standard and can be used with compatible array consumer libraries.
```
lucascolley, line 75: yeah this is good! But it sounds strange that we're mentioning multiple different ways to _ensure_ conformance, hence why I think we should remove the mention for array-api-strict. Ensuring once should be enough! (of course, you can never be 100% sure, but I assume we're writing assuming an ideal state of usage and maintenance of the core packages)

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+(array-api-extra)=
+
+### Array API Extra
+
+GitHub: [array-api-extra](https://github.com/data-apis/array-api-extra)
+
+User group: Array Consumers
+
+`array-api-extra` is a collection of additional utilities and tools that are
+missing from the Array API Standard but can be useful for compliant array
```
lucascolley, line 87: 'missing' perhaps sounds a bit negative, I'm not sure what I would say instead but this seems to imply that, in an ideal state, everything in array-api-extra would be in the standard itself. I don't think that is true.

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+`array-api-tests` is a collection of tests that can be used to verify the
+compliance of your library with the Array API Standard. It includes tests
+for array producers, covering a wide range of functionalities and use cases.
+By running these tests, you can ensure that your library adheres to the
+standard and can be used with compatible array consumer libraries.
+
+
+(array-api-extra)=
+
+### Array API Extra
```
lucascolley, line 80: This does not mention tools for lazy backends (https://lucascolley.github.io/talks/pydata-paris-25-array-api/#/5/13) and `xpx.testing` (https://lucascolley.github.io/talks/pydata-paris-25-array-api/#/5/15)

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+GitHub: [array-api-extra](https://github.com/data-apis/array-api-extra)
+
+User group: Array Consumers
+
+`array-api-extra` is a collection of additional utilities and tools that are
+missing from the Array API Standard but can be useful for compliant array
+consumers. It includes additional array manipulation and statistical functions.
+It is already used by SciPy and scikit-learn.
+
+The sections below mention when and how to use them.
```
lucascolley, line 91: I would move this up above the individual library sections otherwise it looks like this is under the array-api-extra section

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+     expected to fail actually passes (`XPASS`), then you can decide whether
+     to ignore that fact or raise it as an error.
+   - `--skips-file` for skipping tests. At times, some failing tests might stall
+     the execution time of the test suite. In that case, the most convenient
+     option is to skip these for the time being.
+
+We strongly advise you to embed this setup in your CI as well. This will allow
+you to continuously monitor Array API coverage, and make sure new changes don't break existing
+APIs. As a reference, see [NumPy's Array API Tests CI setup](https://github.com/numpy/numpy/blob/581d10f43b539a189a2d37856e5130464de9e5f6/.github/workflows/linux.yml#L296).
+
```
lucascolley, line 149: If I was a library developer reading this, I would be wishing there was also a link to how to set this up within a Pixi workspace, like https://github.com/mdhaber/mparray/blob/0ef47e008fef92c605f73907436d4c6617419161/pixi.toml#L119-L179

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+
+(array-consumers)=
+
+## Array Consumers
+
+For array consumers, the main premise is to keep in mind that your **array
+manipulation operations should not lock in for a particular array producing
+library**. For instance, if you use NumPy for arrays, then your code could
+contain:
```
lucascolley, line 169:
```suggestion
For array consumers, the main premise is that your **array
manipulation operations should not be specific to one particular array producing
library**. For instance, if your code is specific to NumPy, it might contain:
```

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+
+At this point, replacing one backend with another one should only require providing a different
+namespace, such as `xp = torch` (e.g., via an environment variable). This can be useful
+if you're writing a script or in your custom software. The other alternatives are:
+
+- If you are building a library where the backend is determined by input arrays,
+  and your function accepts array arguments, then a recommended way is to ask
+  your input arrays for a namespace to use: `xp = arr.__array_namespace__()`.
+  If the given library doesn't have it, then [`array_api_compat.array_namespace()`](https://data-apis.org/array-api-compat/helper-functions.html#array_api_compat.array_namespace)
+  should be used instead:
```
lucascolley, line 207: I think we should skip the idealism and just recommend `array_namespace()` for now until there are real use-cases with `__array_namespace__()`

spec/draft/migration_guide.md
```diff
@@ -0,0 +1,236 @@
...
+  def func(array1, scalar1, scalar2):
+    xp = array1.__array_namespace__()  # or array_namespace(array1)
+    return xp.arange(scalar1, scalar2) @ array1
+  ```
+- For a function that accepts scalars and returns arrays, use namespace `xp` as
+  a parameter in the signature. Enforcing objects to have the same type as the
+  provided backend can then be achieved with `arg1 = xp.asarray(arg1)` for each input:
+  ```python
+  def func(s1, s2, xp):
+    return xp.arange(s1, s2)
```
lucascolley, line 218: this example isn't particularly compelling, since it is something that is already achievable more easily on the user-side. I think it is worth a sentence stating when this may be worth it (e.g. there may be significant computation that you want to happen native to the array library before returning, https://docs.scipy.org/doc/scipy/reference/generated/scipy.fft.fftfreq.html) versus when it probably isn't (e.g. just trivially wrapping a value with `xp.asarray` before return, https://docs.scipy.org/doc/scipy/reference/constants.html)

lucascolley (commented):
Sorry, I missed this PR. Apologies for the post-merge review, unfortunate that this was merged just before the meeting.

I've only had time to look through the migration guide so far, see below for suggestions.

mtsokol:
@lucascolley Thanks for reviewing! I'll open another PR for it.
And good to know there has been a dicussion around https://github.com/data-apis/array-api/discussions/778 - I often wondered what was the right form and mostly used "Array API Standard" in a daily work.

kgryte:
FTR, nowhere are we consistent with the naming convention.

lucascolley:
> FTR, nowhere are we consistent with the naming convention.

indeed, hence why I'd like to get more consistent with the place where we thought about it 😅

kgryte:
Personally, I think the lack of capitalization (array API standard) was a mistake, as it is more confusing for a proper noun to be lowercase.

spec/draft/migration_guide.md
mtsokol: But then you always enforce one extra dependency. Does it mean that if you want to work with the array API then `array-api-extra` is always a required dependency?

lucascolley: Hmm, I suppose it depends on the target audience. If one just wants to work with NumPy and JAX, then yeah, you can probably avoid the array-api-compat dependency (depending on if you are hitting NumPy edge cases or not) and use `__array_namespace__`. But I think in all other cases you would need `array_namespace` right now?

I'm happy to keep mention of `__array_namespace__`, but I'd hesitate to call it 'a recommended way' right now.

spec/draft/migration_guide.md
mtsokol: done

spec/draft/migration_guide.md
mtsokol: done

spec/draft/migration_guide.md
mtsokol: done

spec/draft/migration_guide.md
mtsokol: Ability to interchange namespaces is the result of making functions agnostic. We don't make functions agnostic for its own sake, but rather to be able to switch from e.g. NumPy to PyTorch. 
So to make it clear:
```
...you can use the second part to learn how to make it
library agnostic and, as a result, interchange
array namespaces with significantly less friction.
```

spec/draft/migration_guide.md
mtsokol: Good point. done

spec/draft/migration_guide.md
mtsokol: Right - I mentioned that `array-api-compat` provides support for an additional subset of these corner cases.

spec/draft/migration_guide.md
mtsokol: done

spec/draft/migration_guide.md
mtsokol: done

spec/draft/migration_guide.md
mtsokol: all agree - done!

spec/draft/migration_guide.md
mtsokol: done

spec/draft/migration_guide.md
mtsokol: removed

spec/draft/migration_guide.md
mtsokol: Sure, I removed the testing aspect in the description of `array-api-strict`.

spec/draft/migration_guide.md
mtsokol: changed to `not present in`.

spec/draft/migration_guide.md
mtsokol: done

spec/draft/migration_guide.md
mtsokol: I just removed this sentence

spec/draft/migration_guide.md
mtsokol: done

spec/draft/migration_guide.md
mtsokol: done

mtsokol:
Here's a follow-up PR: https://github.com/data-apis/array-api/pull/1000 with a round number 💯

spec/draft/migration_guide.md
mtsokol: Ok, changed it so `array_namespace` is mentioned first as a recommended option, and later `__array_namespace__` if one wants to avoid extra dependency.

spec/draft/migration_guide.md
lucascolley: ah maybe we are thinking of 'interchange' differently, I was thinking of e.g. in-memory interchange like DLPack.

mtsokol: ah, what I had in mind was the concept of plain replacing namespaces e.g.:
```
xp = np if ... else torch
```

lucascolley: I'm not sure what words I would use to communicate that clearly...

PR #1003, opened 2026-04-13
docs: update stakeholders section

mtsokol:
Updated Purpose section, introduced stakeholders table, and removed obsolete projects.


mtsokol:
Should we also have this change in 2025.12 directory or only draft?

rgommers:
> Should we also have this change in 2025.12 directory or only draft?

Only in draft seems fine to me. Normally we only backport bug fixes.

kgryte (approved):
LGTM

PR #1004, opened 2026-04-16
docs: add migration tables for NumPy as PyTorch

mtsokol:


spec/draft/migration_guide.md
```diff
@@ -234,3 +234,84 @@ offers a set of useful utility functions, such as:
...
+
+## Migration patterns for selected libraries
+
+Below, you can find a non-exhaustive list of API calls that are present in NumPy
+and PyTorch but are not supported by the Array API Standard. For each of them, we
+provide the recommended alternative from the standard, along with some notes on
+how to use it.
+
+### NumPy
+
```
rgommers, line 246: It may be good to start here that the Array API column uses `xp.` but all those calls works with `np.` as well - and usually it's more idiomatic NumPy anyway.

spec/draft/migration_guide.md
```diff
@@ -234,3 +234,84 @@ offers a set of useful utility functions, such as:
...
+| `np.trace(x)` | `xp.linalg.trace(x)` |  |
+| `np.diagonal(x)` | `xp.linalg.diagonal(x)` |  |
+| `np.cross(a, b)` | `xp.linalg.cross(a, b)` |  |
+| `np.outer(a, b)` | `xp.linalg.outer(a, b)` |  |
+| `np.matmul(a, b)` | `xp.linalg.matmul(a, b)` or `xp.matmul(a, b)` |  |
+| `np.ravel` | `xp.reshape(x, (-1,))` |  |
+| `x.flatten` | `xp.reshape(x, (-1,))` |  |
+
+### PyTorch
+
```
rgommers, line 296: I think this needs a similar but slightly different note: most of the second column is available in PyTorch, but there are a few entries that rely on `array-api-compat` usage (which is the recommended way to add PyTorch support to a library that wants to use the array API).

rgommers (commented):
Thanks @mtsokol! Overall this looks great and will be quite useful. A couple of comments on adding some clarifying text.

spec/draft/migration_guide.md
mtsokol: Right, for PyTorch we need `array-api-compat` in way more cases so I think it's better to require compat anyway for clarity.

spec/draft/migration_guide.md
mtsokol: That's right, I added a clarification.

spec/draft/migration_guide.md
mtsokol: Done.

tylerjereddy:
ah yeah, really useful, just read through it

kgryte (approved):
LGTM

PR #1005, opened 2026-05-01
refactor!: allow subsets of dtypes and allow device-dependent dtype support

ev-br:
Per discussion in gh-998 and the [community meeting on Apr 30th](https://hackmd.io/zn5bvdZTQIeJmb3RW1B-8g?view#Meeting-minutes-30-April-2026), 

- soften the requirement that all dtypes are implemented from _must_ to _should_;
- require that at least one integer dtype and one floating-point dtype  is actually supported (this allows us to not bother with edge cases of e.g. no floats at all);
- recommend that unsupported combinations of `dtype=..., device=...` raise an exception + specify that creation functions _should_ raise. Note that this still leaves `astype` and `to_device` only implicitly specified and only covered by the recommendation to raise.
- specify that cross-device transfers in `__getitem__` and `__setitem__` are implementation-defined (and presumably covered by the recommendation to raise).

### DIFF
--- a/spec/draft/API_specification/data_types.rst
+++ b/spec/draft/API_specification/data_types.rst
@@ -39,8 +39,7 @@ objects in its main namespace under the specified names:
 | complex128   | Double-precision (128-bit) complex floating-point number whose real and imaginary components must be IEEE 754 double-precision (64-bit) binary floating-point numbers (see IEEE 754-2019). |
 +--------------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
 
-If a library only supports a subset of data types, it must support at least one data type
-from each :ref:`category <data-type-categories>`.
+If a library only supports a subset of data types, it must at a minimum support ``bool`` and at least one integer dtype and one real floating-point dtype.
 
 Data type objects must have the following methods (no attributes are required):
 
--- a/spec/draft/API_specification/data_types.rst
+++ b/spec/draft/API_specification/data_types.rst
@@ -5,7 +5,7 @@ Data Types
 
     Array API specification for supported data types.
 
-A conforming implementation of the array API standard must provide and support
+A conforming implementation of the array API standard should provide and support
 the following data types ("dtypes") in its array object, and as data type
 objects in its main namespace under the specified names:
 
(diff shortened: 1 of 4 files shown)


spec/draft/API_specification/data_types.rst
```diff
@@ -39,6 +39,9 @@ objects in its main namespace under the specified names:
 | complex128   | Double-precision (128-bit) complex floating-point number whose real and imaginary components must be IEEE 754 double-precision (64-bit) binary floating-point numbers (see IEEE 754-2019). |
 +--------------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
 
+If a library only supports a subset of data types, it must support at least one data type
+from each :ref:`category <data-type-categories>`.
```
rgommers: I don't think `complex` should be added here, rather `bool` and one integer and one real floating-point dtype must be supported.

rgommers (commented):
Thanks @ev-br! One comment on a `complex` dtype not needing to be required, other than that this LGTM. 

I considered whether to require specific types that are currently universally available (e.g., `float32`, `int32`), but that doesn't really help in any way while it does potentially create another issue if support shifts over time. So "one integer and one float dtype" seems just fine. In practice, libraries are going to just use them anyway and they will exist; actual code will rely on how to get things working with real-world libraries, not probing the must/should distinctions in the spec (that's mostly relevant for the test suite and authors of new array libraries).

spec/draft/API_specification/data_types.rst
ogrisel: That would be enough for scikit-learn at least.

spec/draft/API_specification/data_types.rst
```diff
@@ -39,6 +39,9 @@ objects in its main namespace under the specified names:
 | complex128   | Double-precision (128-bit) complex floating-point number whose real and imaginary components must be IEEE 754 double-precision (64-bit) binary floating-point numbers (see IEEE 754-2019). |
 +--------------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
 
+If a library only supports a subset of data types, it must support at least one data type
```
ev-br:
```suggestion
If a library only supports a subset of data types, it must at a minimum support ``bool`` and at least one integer dtype and one real floating-point dtype.
```

spec/draft/API_specification/data_types.rst
ev-br:
```suggestion
```

ev-br: Updated to the suggestion.

ogrisel (approved):
LGTM.

kgryte (approved):
LGTM

kgryte:
Looks like we have consensus. I will go ahead and merge and we can still tinker with guidance in the lead up to cutting the v2026 revision later this year.

PR #1011, opened 2026-08-07
docs: Update the list of DLPack devices

betatim:
Taken directly from the [dlpack.h header](https://github.com/dmlc/dlpack/blob/77aafa4d3b0f80feffce9ad4c718dd26751ee0e4/include/dlpack/dlpack.h#L75).


rgommers (approved):
These LGTM, thanks @betatim! 

I'll suggest one addition: could you add a code comment that we left out enum 12 (ExtDev) on purpose because its semantics aren't fixed?

betatim:
Done.

